PR #464, opened 2013-09-15
Before hook check moved out of routes matching in dispatcher

cym0n:
Run the test in the branch: t/hook_changing_path.t 
It will go well.
Now take it in a different directory and run it again (obviously still with Dancer2 in your @INC). It will fail.

It's a self contained test, why should it fail?

Problem is that the dispatcher runs before hooks during the route matching and the before hook we introduced change routes, modifying something we didn't configure (/prefix/configured) in something we did (/configured).

When the script is in the t directory this is not a problem because the Dancer2::Handler::File appends a megasplat route to the app's routes and the before hook can be triggered when testing /prefix/configured against it. 
In a directory different from t, Dancer2 environment can't determinate a public directory (under t we have a public...) so the Dancer2::Handler::File can't register itself to the app and the megasplat can't come saving us.

I tried to make the before hook triggered before the route matching, but I'm not sure this is the right thing to do. It makes sense for you?

This is still very dirty code. I left all the old halted context management inside the loop, just copying it also beofre it. But I want your feedback about the question before refactoring.

Problem I exposed has bad side effects in real world. I have a plugin that fails dzil test for that. A good workaround is creating a public directory under the t, but it's a bit dirty...

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -240,6 +240,7 @@ sub _init_for_context {
 
 sub supported_hooks {
     qw/
+      core.app.before_route_match
       core.app.before_request
       core.app.after_request
       core.app.route_exception
--- a/lib/Dancer2/Core/Dispatcher.pm
+++ b/lib/Dancer2/Core/Dispatcher.pm
@@ -51,7 +51,7 @@ sub dispatch {
 
         $app->log( core => "looking for $http_method $path_info" );
 
-        $app->execute_hook( 'core.app.before_request', $context );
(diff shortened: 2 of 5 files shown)


xsawyerx:
I'm closing this issue for the following reasons:

1. The code is quite rough.
2. It would likely need to be updated to the latest version. (It's been a while.)
3. I still have not seen a use-case that justifies having a hook for after match and using that instead of something like `Plack::Middleware::Rewrite` and `forward`.
4. I'd like to clean the pull request backlog.

You're more than welcomed to open a PR with a new rebased version and we could discuss it again. I hope that's okay.

Thanks for understanding.

PR #571, opened 2014-04-24
put HTTP_REQUEST_BASE support back from Dancer 1

fayland:
I think it is still needed in Dancer2 when tested in our server.

the code is copied from Dancer::Request.

Thanks


xsawyerx:
Since `Plack::Middleware::ReverseProxyPath` is available, I think we should close this one. We should follow up on #590 for a comfortable way of integrating it.

Issue #589, opened 2014-05-20
Remove Core::Context in favor of smarter Core::Request


xsawyerx:
One thing that bugs me is `Core::Context`. It is created in the app from the request, contains a reference back to the app, and the app contains a reference to itself. It holds both data and actions, but neither on itself.

Until now I wasn't even sure what the hell it is exactly meant for. Now I understand it better, and it seems like we could actually remove it (and some duplicate code that exists there) in favor of making the `Core::Request` object more aware of able.

I'd like to get some opinions on this.

PR #590, opened 2014-05-22
Add behind proxy middleware

veryrusty:
Collects the request header munging that was being done when building the Request object into a Plack middleware (wraps existing Plack middlewares for request munging behind proxies).

Simplifies the request object (before tackling #589) and adds support for reverse proxies from non-root paths (#571). Devs can (conditionally) apply the middleware (via Plack::Builder) to provide more flexibility than was previously possible (eg BehindProxy -> MultiLang middleware-> Dancer app ).

I'm not 100% convinced that `behind_proxy` should be a global config item though - yell if you have opinions :D

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -64,6 +64,9 @@ MIME::Types = 0
 MIME::Base64 = 3.13
 ; added the "wrap" method in Plack::Builder
 Plack = 1.0016
+; BehindProxy middleware
+Plack::Middleware::ReverseProxy = 0
+Plack::Middleware::ReverseProxyPath = 0
 ; JSON is just a serialiser, so in theory should be optional.
 ; But it is used in the DSL (send_error) and in tests.
 ; And also so much used in apps...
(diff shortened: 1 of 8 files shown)


racke:
This is a very old pull request with a complicated discussion. So if you have a compelling reason to keep it open, please speak up. Otherwise I will consider to close it soon.

veryrusty:
@racke yes its old, but please leave the PR as is.

racke:
:shrug:

xsawyerx:
Reviewed this with @cromedome. Here's a summary of our thoughts:

1. If you are setting `behind_proxy` to true, we can add another middleware that checks whether you are using one of the non-defacto headers and override the defacto header with their values. The middleware would warn each time it detects such a header so users stop using unsupported headers, and it will also warn if there are multiple headers with different values - giving priority to the defacto header.
2. We never supported `Request-Base` in D2 so we can just use the Plack middlewares' header and document we don't support `Request-Base` in the Migration document.
3. Considering it's a security risk to use the first address in a series of proxies in a header, we should switch to the last one (essentially letting the Plack middleware do what it's supposed to). We can't think of a way to make this transition easy or seamless, so we'll just document it clearly in the release.

PR #645, opened 2014-07-30
Added serialize/deserialize DSL keywords

DavsX:
These keywords would use the serializer engine, which is set on the app. That way, the Dispatcher can identify possible serialization/deserialization errors.

Current to_json, to_yaml, from_json etc methods all create a new object, which will then be used for the serialization job. But the error checking is done only on the $app->engine('serializer')

These keywords would also use the serializer, which is configured in the app (json, yaml, dumper whatever).

### DIFF
--- a/lib/Dancer2/Core/DSL.pm
+++ b/lib/Dancer2/Core/DSL.pm
@@ -30,6 +30,7 @@ sub dsl_keywords {
         dancer_major_version => { is_global => 1 },
         debug                => { is_global => 1 },
         del                  => { is_global => 1 },
+        deserialize          => { is_global => 1 },
         dirname              => { is_global => 1 },
         dsl                  => { is_global => 1 },
         engine               => { is_global => 1 },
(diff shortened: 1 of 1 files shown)


veryrusty:
I'm not keen on these keywords added to the core either. 

Slightly off-topic: With the recent change regarding serializing everything if a serializer is defined in the app, even the helpers (`from_json`, `to_json`, etc) are not really that "helpful" - eg. they handle errors poorly, noted at being slow due to object instantiation and they don't use the config from the apps serializer. Is it worth contemplating moving those helpers (and possibly the keywords in this Pr) into a plugin?

DavsX:
If we would get 'rid' of from/to_json/yaml etc keywords, that could possibly help us to finally solve https://github.com/PerlDancer/Dancer2/issues/686. That way those keywords could have their own encoding policy separate from the way the encoding is handled when a route serializer is set.

veryrusty:
@DavsX that was my thinking .. let me write something up as a separate issue.

DavsX:
Good idea! The contents of this PR are outdated anyway, so I'm closing this and the discussion should go into an issue, so others would see it as well.

Issue #644, opened 2014-07-30
to_json masks exceptions thrown by JSON


LionsPhil:
`JSON::to_json` will die if you pass it objects without its `allow_blessed` options. Dancer's `to_json` keyword wrapping this masks these exceptions. It is not clear if this is intentional or a documentation omission (drilling down through the code I can't see where the exception is actually being caught). It seems more correct to me that it should not mask them, since the effect of returning undef in the most common use case:
```perl
return to_json { some => $structure };
```
...is that Dancer will get a `return undef` and serve a 200 OK response, when your application has encountered a bug and should (probably) be serving its HTTP 500 page.

You can reproduce this trivially by taking the default `dancer -a` app and replacing the route handler with:
```perl
my $object = bless({}, 'oops');
my $json = to_json { oops => $object };
warn "JSON response isn't defined" unless defined $json;
return $json;
```
Which will emit a warning and server a zero-byte document with status 200. Compare to replacing the `my $json` line with using JSON directly:
```perl
use JSON qw();
my $json = JSON::to_json { oops => $object };
```
Which will serve a 500 page and log that the route threw an exception.

xsawyerx:
A 500 error indicates there is a problem with the application code (or with the server) and this seems like it's either developer error (there is no HTTP status for such a thing, sort of) or a bad request, which is a 400.

However, it would not be clear if the problem occurred because of the former or latter or sending a 400 automatically tells the user they screwed up, when in fact it could have been something the developer was doing.

On a `deserialize`, it is the user's fault, most likely. However, blowing up and returning a 400 does not provide the developer with an opportunity to do something about this.

On a `serialize`, it's the developer returning something bad and thus should not return a 400.

I vote to close the issue.

LionsPhil:
I'm not sure I follow your comment, because the problem is you'll actually get a "successful" 200 status if you're doing `return to_json($something_unserializable)`, which is certainly wrong, and no warning or error is logged. (I haven't looked at if there's been any change in this area recently.)

If you intend `to_json` to mask exceptions and not be used like this, I think the bugfix is just documenting this (since it's an unexpected difference from `JSON::to_json`).

xsawyerx:
I was looking at it from the serializer perspective. The serializer shouldn't crash on this.

LionsPhil:
[...]

PR #647, opened 2014-07-31
Allow to/from_json to throw

LionsPhil:
```
Do not implement to/from_json in terms of the (de)serialize()
methods of a temporary Serializer instance, since the error
generated is lost and they will just return undef. Let them throw
instead so that the route code can handle the exception, or propagate
to general "route died" handling. Configuration options are still
merged in using a temporary instance.

Add test dependency on Test::Trap for testing exceptions.

Add unit tests for unserializable data for the JSON serializer.
```
Some considerations for review:
 * Addresses issue #644 for the case where you are using to/from_json without it being set as your serializer (e.g. because you want to respect the configuration, but do not want to blanket handle complex returns from routes).
 * The refactoring to `_merge_options` means that configuration can now affect *de*serialization as well. I'm not sure if this is desirable.
 * I am not clear on what the *intended* behaviour is if you set up a serializer then return unserializable data from your route, so the test for this case is quite loose (and has a TODO). At the moment Dancer seems to be choking on an undef internally in this case, but fixing that seems like material for a separate commit.
 * Should possibly do the same for YAML etc. if this is deemed to be an appropriate change.
 * (First time I've done the pull request workflow, so I hope I've got this right!)

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -110,7 +110,6 @@ HTTP::Body = 0
 Test::TCP = 0
 Test::MockTime = 0
 Test::Script = 0
-Test::Trap = 0
 LWP::Protocol::PSGI = 0.06
 
 ; ; for maintainers, see with mst how to avoid these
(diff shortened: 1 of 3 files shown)


xsawyerx:
Implementation in #972 merged. Thank you for the proposed solution! :+1:

PR #693, opened 2014-09-19
Feature/handler file config option

DavsX:
This PR does a couple of thing (I wanted to separate those into their own PR, but they kinda depend on each other)

* The 'public' config option is added to D2::Core::App's default_config inside _build_default_config. It's value is $ENV{DANCER_PUBLIC} || path($app->location, 'public'). When the public dir is specified in the config, this value will be overwritten in the resulting merged $app->config.
The duplicated logic of building the public dir location is removed from D2::Handler::File and D2::Core::Error; they both use $app->config->{public} now.

* public_dir is removed from the File route handlers default config inside the same _build_default_config method. There is already a Dancer2-global 'public' config option, so why to duplicate, we can simply use that.

* the route_handlers config is changed. Instead of an arrayref, it is now a hashref, like this:
```yaml
route_handlers:
    File: 1
    AutoPage: 0
```
This way individual route handlers can be explicitly enabled/disabled, you no longer need to set 'route_handlers: []' to disable the File handler, you can now set it's value explicitly to 0.

* For now, I added deprecation warning for the old route_handlers config use. If it's value is an old-style arrayref, it is parsed into the new format, so it is working, but it now issues a warning, that it is deprecated.

* I deprecated the auto_page config option as well. There were two config options to enable/disable this feature; AutoPage => 1 inside route_handlers and auto_page itself. The old behavior was like this: the AutoPage handler was created and added to the app's route_handlers array, because the AutoPage config value inside _build_default_config was 0. But inside Handler::AutoPage->register the auto_page option was used to determine if we need to add any routes. 
Now there is only one config option for it; if it is 0, the handler is not even created.

* I updated Dancer2::Config and Dancer2::Cookbook according to these changes.

### DIFF
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -61,10 +61,12 @@ If you prefer YAML, a sample YAML based config file might look like this:
 
     appname: "Hello"
     charset: "UTF-8"
-    auto_page: 1
 
     session: "YAML"
     serializer: "JSON"
+
+    route_handlers:
+        AutoPage: 1
 
     plugins:
       DBIC:
@@ -77,9 +79,11 @@ If JSON is more your thing, your file might look more like this:
     {
         "appname": "Hello",
         "charset": "UTF-8",
-        "auto_page": "1",
         "session": "YAML",
         "serializer": "JSON",
+        "route_handlers": {
+            "AutoPage": 1
+        },
         "plugins": {
             "DBIC": {
                 "default": {
@@ -94,7 +98,6 @@ If you like Apache configuration files, try something similar to:
 
         appname = Hello
         charset = UTF-8
-        auto_page = 1
         session = YAML
         serializer = JSON
         <plugins>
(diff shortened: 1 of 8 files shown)


xsawyerx:
@DavsX, looking at this now, I'm missing one thing: Why. :)

What I'm seeing is basically "I think the route handlers should be a configuration option with keys inside for the specific route handlers someone wants to enable." - So far so good. That's worth discussing and making a decision about. We can then remove the top level handler configuration keys.

What I'm wondering is, did I misunderstand this issue? Did you have some other reason or something else in mind? Are you addressing anything else?

rleir:
Is this all intended to DRY the code a bit? I am in favour of that. Or does it actually fix something? Unfortunately this issue is stale now, and maybe it should be closed. When you get a moment!

xsawyerx:
@DavsX, @rleir, looking back now, I think the problem this PR has is that the `File` handler is only used right now for `send_file`. It used to be for static content, and was last. That was removed and the static content is now handled by the `Static` middleware, and it is done first (as D1 does it - which we deemed is more correct).

As it appears, the only "handler" active is the `AutoPage` handler, and I'm not sure this change is necessary currently. Perhaps the config part.

Should we close this?

xsawyerx:
No comment. I vote for closing.

cromedome:
@xsawyerx: I second that.

PR #747, opened 2014-11-03
Pr/test with return failures

shumphrey:
If a before_template_render hook dies, and a Dancer2::Core::Error uses a template, then this causes the dispatch block to die.
This means app->cleanup and app->clear_with_return never get called.
All subsequent requests however unrelated now have the wrong session and wrong with_return.


xsawyerx:
The tests seem to pass now with all the work @veryrusty has done on this. I'm closing this PR. I added another test that this PR carries which we don't have yet.

Thanks! :+1:

Issue #356, opened 2013-08-04
Add require_environment setting


yanick:
Port perldancer/dancer#946 to Dancer2

xsawyerx:
So, if I could delete all my previous comments on this issue and write this simple thing:
I feel like adding this additional configuration setting feels awkward.

Instead, what if we had "include" available as a configuration option, which would require us to include files? We could have it include any number using a glob (any that match) or filenames. If a file is mentioned by name (file-name, get it?), and it doesn't exist, it will crash. We make sure we never read files more than once, so if we already read the environment file (because it existed and we automatically do that), and you mentioned it as an include, it won't be read again. But, if it didn't exist (silently ignoring it), it would now try to find it and fail loudly.

This would be a generic solution which could be used not only to solve this problem, but to add additional include files (which we currently don't support).

cromedome:
I'm very much in favor of an include directive. There are times I wish I could tuck some information away in another file separate from the ones Dancer uses, and I'd like to use them without creating my own YAML/JSON reader every time I need to access them.

How would this interact with the notion of `strict_config`? IIRC, `strict_config` would create an object from the config file. Would we leave anything included out of this object? Find a way to add functionality to the config object? Dump the existing config object and rebuild it with the new files? I can see use cases for all three, but my inclination would be don't alter the config object once it's instantiated.

:+1: on the include idea though!

GeekRuthie:
I'm looking at `Dancer2::Core::Role::ConfigReader`, and it looks like `sub load_config_file` is where an include-directive handler would need to go--I'm willing to take a swing at this, if there's still a desire for that.

One question, though--there's an ancient note in this sub `# TODO handle mergeable entries`, and I wanna make sure I understand **that** part of the problem--it seems like this would be a good time to clear up that tech debt, as well.  By "mergeable entries," are we talking about correctly merging keys in the config, so that something like this will work:

In `config.yml`:
`plugins:
   Foo:
      setting_1: "fizz buzz"`

..then in an environment file:
`plugins:
   Foo:
      setting_2: "foo bar baz"`

As things are currently, I *think* that the environment would take precedence, and setting_1 would not be set.  By "mergeable", do we mean that in this scenario, both setting_1 and setting_2 should be set?

yanick:
>  By "mergeable", do we mean that in this scenario, both setting_1 and setting_2 should be set?

That'd be my understanding, yup!

[...]

PR #780, opened 2014-11-25
Implement require_environment config option (docs missing atm)

DavsX:
Implement a require_environment config option, which if present checks
if the config file for the currently set environment was found. If not,
it raises an error.


rleir:
Is this covered by #356? If so, can you close this when you get a moment? Thanks

cromedome:
Closing in favor of #356. Thanks all.

cromedome:
That being said, I won't forget this is sitting out here - may still be tremendously helpful as we sort out #356. Thanks for the PR @DavsX.

Issue #754, opened 2014-11-06
Calling plugins inside plugins crashing


racke:
This works in Dancer1 and is definitely a needed feature - we don't want code duplication and we want KISS.

Demonstrated by test in https://github.com/racke/Dancer2-Plugin-Auth-Extensible-Provider-DBIC.

Output:

```
[MyApp:25330] debug @2014-11-06 16:16:15> Attempting to authenticate max against realm users in /home/racke/perl5/perlbrew/perls/perl-5.20.0/lib/site_perl/5.20.0/Dancer2/Core/Route.pm l. 142
[MyApp:25330] error @2014-11-06 16:16:15> Route exception: Cannot load provider Dancer2::Plugin::Auth::Extensible::Provider::DBIC: Can't locate object method "dsl" via package "Dancer2::Plugin::Auth::Extensible::Provider::DBIC" at /home/racke/perl5/perlbrew/perls/perl-5.20.0/lib/site_perl/5.20.0/Dancer2/Plugin.pm line 88.
BEGIN failed--compilation aborted at /home/racke/interchange/Dancer2-Plugin-Auth-Extensible-Provider-DBIC/lib/Dancer2/Plugin/Auth/Extensible/Provider/DBIC.pm line 6.
Compilation failed in require at /home/racke/perl5/perlbrew/perls/perl-5.20.0/lib/site_perl/5.20.1/Dancer2/Plugin/Auth/Extensible.pm line 547. in /home/racke/perl5/perlbrew/perls/perl-5.20.0/lib/site_perl/5.20.0/Dancer2/Core/Dispatcher.pm l. 68
```

hvoers:
This must be at the top off all todo lists.

racke:
"must" is a bit strong ... but a big boulder in terms of moving Dancer2 to be production ready IMHO.

xsawyerx:
This was fixed by a patch long ago. @yanick is working on a new implementation which will cover it better, but the problem itself does not exist anymore.

murwiz:
I just encountered this; are there particular versions I should be looking for to make sure I have the patch?

xsawyerx:
@murwiz, version 0.160000.

PR #792, opened 2014-11-29
RFC: Allow prototypes for keywords:

xsawyerx:
This is just a quick thing I put together when discussing some
syntax for prefix setups with Aristotle. Just an idea and an
implementation we can discuss.

I don't know how important it is to allow prototypes, but if it
doesn't take anything from us, why not.

This breaks the plugin_syntax.t test - I didn't check why.

### DIFF
--- a/lib/Dancer2/Core/Role/DSL.pm
+++ b/lib/Dancer2/Core/Role/DSL.pm
@@ -50,7 +50,7 @@ sub export_symbols_to {
     my $exports = $self->_construct_export_map($args);
 
     foreach my $export ( keys %{$exports} ) {
-        no strict 'refs';
+        no strict 'refs'; ## no critic (TestingAndDebugging::ProhibitNoStrict)
         my $existing = *{"${caller}::${export}"}{CODE};
 
         next if defined $existing;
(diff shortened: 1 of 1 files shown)


xsawyerx:
No one had complaints so I'm merging it. I would like the streaming support to use it for shinier syntax.

xsawyerx:
Bah, I forgot it broke stuff.

xsawyerx:
Rebased and pushed force.

veryrusty:
As this evals a string to get the coderef, it should be possible to include the idea from #819, to reduce one level of indirection in the compiled subs; something like
```
    my $compiled_code = eval qq{
        sub($prototype) {
          if ( ! \$opts->{'is_global'} ) {
            \$self->app->has_request or
              croak "Function '\$keyword' must be called from a route handler";
          }
          \$self->$keyword(\@_);
        };
    };
```

veryrusty:
The test failure is due to the import into the plugin's namespace of the DSL keywords, the ones with the deprecation notice to use `$dsl->keyword` instead.

`t/lib/PluginWithImport.pm` currently has
```
use strict;
use warnings;

use Dancer2;
use Dancer2::Plugin;
...
```
which imports the DSL symbols once with the prototypes (`use Dancer2`), and then attempts to import the wrapped versions without prototypes.

It's simple to remove the `use Dancer2;` line to make the test pass. **BUT** I've done a quick scan of plugins on cpan/github and this will cause downstream breakage as there are more than a handful of plugins that do this. eg
  * on cpan: BrowserDetect, Sixpack, Email
  * github: Interchange6 (semi random example to ping @racke for feedback )

I'm not sure what the best way to tackle this. :(

xsawyerx:
I like the improved `eval`. :+1:

xsawyerx:
Although, we might be able to do the condition outside, so it's only run when compiling.

hvoers:
On Sat, 7 Feb 2015, Russell Jenkins wrote:

> The test failure is due to the import into the plugin's namespace of the DSL keywords, the ones with the deprecation notice to use $dsl->keyword instead.
> 
> t/lib/PluginWithImport.pm currently has
> 
> use strict;
> use warnings;
> 
> use Dancer2;
> use Dancer2::Plugin;
> ...
> 
> which imports the DSL symbols once with the prototypes (use Dancer2), and then attempts to import the wrapped versions without prototypes.
> 
> It's simple to remove the use Dancer2; line to make the test pass. BUT I've done a quick scan of plugins on cpan/github and this will cause downstream breakage as there are
> more than a handful of plugins that do this. eg
>
>  *  on cpan: BrowserDetect, Sixpack, Email

D*::Plugin::Email is one of the plugins that I can use with D1 and not with D2.

>  *  github: Interchange6 (semi random example to ping @racke for feedback )

See also: https://github.com/PerlDancer/Dancer2/issues/754

> I'm not sure what the best way to tackle this. :(

What about:

http://lists.preshweb.co.uk/pipermail/dancer-users/2015-February/004358.html

-- 
Henk

veryrusty:
I have a simple solution.  :smile:

I'm putting together a new Pr that also includes allowing plugins to add prototypes to their registered keywords. Just some test cases to go :)

veryrusty:
#848 is an alternative to this Pr. with passing tests :)

xsawyerx:
#848 was merged, we can close this.

Issue #793, opened 2014-11-29
Make prefix a full spec like a path


xsawyerx:
Currently the `prefix` can only be a simple string, it cannot be a regexp in and of itself. We should allow it to have the same spec as a regular path.

This is mentioned (and, in a way, is part of) #558.

xsawyerx:
Resolved in #845.

Issue #558, opened 2014-04-04
Prefix inconsistency


baynes:
If I have the following routes:
prefix undef;
get '/a' ... it matches '/a' and does not match '/a/'.
get '/b/' ... it matches '/b/' and does not match '/b'.
which is what I would expect.

But if I write the following:
prefix '/b'; get '/' ... it matches '/b' and '/b/'.
But I would expect it to behave as get '/b/' ... above.

It can give some nasty surprises if one has relative links in the generated page. Depending on if the page is loaded as '/b' or '/b/' the links are relative to different places.

It is also desirable to be able to write:
prefix '/b'; get '' ... to be able to match '/b' and not '/b/'.

xsawyerx:
This has been resolved in the latest version thanks to #845.

PR #817, opened 2014-12-23
Consistent prefixes: variables and leading forward slashes:

xsawyerx:
I couldn't break this down into multiple commits so instead a
single commit which fixes two issues. They are both tricky to
grasp and one might introduce breakage (while reducing breakage,
I guess), so they're important to understand.

The first relates to allowing variables in prefixes with a single
path request. If your route is simply '/', variables in the prefix
are simply ignored:

    prefix '/:board' => sub {
        get '/' => sub {...}
    };

Now the actual route registered is '/:board/', literally. That
means that it won't accept variables but only '/:board/'.

The second issue is accepting anything that does not start with
a leading forward slash. This would be an error:

    get '' => sub {...}; # breaks

If you have a prefix and you want to still grab the basic path
without the ending slash it would fail:

    prefix '/:board' => sub {
        get '' => sub {...}; # breaks
    };

This is now fixed too.

One thing to research and perhaps write a test for is the behavior
using path() or path_info(). path() is defined by
path_info() || '/', which might be a problem.

Another issue to recognize is that if someone had the following
code and we don't cover for it, they will receive a 404:

    prefix '/:board' => sub {
        get '/' => sub {
            # expecting to grab both /foo/ and /foo
        };
    };

This, if successful, closes GH #558 and possibly GH #793.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -1,9 +1,7 @@
 package Dancer2::Core::Request;
-
 # ABSTRACT: Interface for accessing incoming requests
 
 use Moo;
-
 use Carp;
 use Encode;
 use HTTP::Body;
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -138,17 +138,10 @@ sub BUILDARGS {
     my $prefix = $args{prefix};
     my $regexp = $args{regexp};
 
-    # regexp must have a leading /
-    if ( ref($regexp) ne 'Regexp' ) {
-        index( $regexp, '/', 0 ) == 0
-            or die "regexp must begin with /\n";
-    }
-
     # init prefix
     if ( $prefix ) {
         $args{regexp} =
             ref($regexp) eq 'Regexp' ? qr{^\Q${prefix}\E${regexp}$} :
-            $regexp eq '/'           ? qr{^\Q${prefix}\E/?$} :
             $prefix . $regexp;
     }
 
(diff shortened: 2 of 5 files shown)


veryrusty:
Thanks @yanick for the cool head - pinning down the spec is what's this is about.

The 3rd option gives the most flexibility, and after some reflection is my preferred option too.
I also like the idea of printing the routes - we do save the (concatenated) strings before turning them into a regex in Core::Route, so this may be straight forward to add.

 I _think_ the only tricky/technical part is dealing with the "default" prefix `/`. For example
```
use Dancer2;
get 'x' => {...};
get '/x' => {...};
```
does this throw an error for the first route (`'x'`), or do these
  * match `/x` and `//x` respectively (not back compatible),
  * both match `/x`,      (implicit '/' as the prefix)
  * match `x` and `/x` if the SCRIPT_NAME header is set, otherwise both match '/x' ?

Thoughts ?

ambs:
Just a note: although we might support strange constructs, like that prefix for sawyer, and route for the x, and although we should test them, in order to know things are accordingly our "internal" decision, that doesn't mean we need to document all the details, and just document "what you should do".

If some user finds out that he can tweak a little more, he will, hopefully, understand the logic.

Well, my cents, but still not sure this is the best approach.

veryrusty:
To get this discussion kicked off again, I've created Pr #845 with some extra tests and a possible approach to handling the case when the route prefix is undef but the concatenated route regexp does not begin with a `/`.

veryrusty:
This was merged as part of #845. @xsawyerx++ 

Thanks for the discussion everyone!

xsawyerx:
++ everyone!
@veryrusty++ for #845!

PR #819, opened 2014-12-28
Simplify D2::Core::Role::DSL::export_symbols_to

JRaspass:
When importing the DSL keywords into the caller they are currently wrapped in
subs that close over $self like so:

*caller::foo = sub { $self->foo(@_) }

This allows simple calling conventions for the caller. If the keyword has
restrictions on where is can be called from (is_global=0) then this sub is
additionally wrapped like so:

my $code = sub { $self->foo(@_) };

*caller::foo = sub {
    croak 'Not from here buddy' unless $self->app->has_request;

    $code->(@_);
};

This is wasteful as it creates another layer of indirection. This commit changes
how those DSL keywords are imported like so:

*caller::foo = sub {
    croak 'Not from here buddy' unless $self->app->has_request;

    $self->foo(@_);
};

One point of note is that export_symbols_to no longer returns a list of what it
has imported, the two and only calls to this don't seem to make use of this so I
can't see it being a huge issue but it could be made to work if needed.

### DIFF
--- a/lib/Dancer2/Core/Role/DSL.pm
+++ b/lib/Dancer2/Core/Role/DSL.pm
@@ -47,49 +47,29 @@ sub dsl { $_[0] }
 # exports new symbol to caller
 sub export_symbols_to {
     my ( $self, $caller, $args ) = @_;
-    my $exports = $self->_construct_export_map($args);
 
-    foreach my $export ( keys %{$exports} ) {
+    $caller .= '::';
+
+    while ( my ( $keyword, $opts ) = each %{ $self->keywords } ) {
+        # Skip if the keyword was excluded from importation.
+        next if $args->{"!$keyword"};
+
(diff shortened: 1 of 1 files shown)


veryrusty:
@JRaspass Thanks for taking the time to refactor your previous commits.

Nice to see that one level of redireciton removed. I'm also happy with `export_symbols_to` no longer returning the list of what was imported (and its easy enough to reintroduce if ever needed). I've giving it a :+1:.

ping @shadowcat-mst: I also do not follow your previous comment in this issue, could you please elaborate?

veryrusty:
The keyword prototypes work has been merged which included similar logic (based off this Pr) to remove the extra layer of indirection. Closing Pr.

@JRaspass thanks again for making the time and effort for putting this together.

PR #824, opened 2015-01-02
Put MIME::Types as optional

ambs:
The mime PR!
````````````````
 _,_ 
(___)
 + + 
( - )
  o  
``````````````````
As discussed in IRC with @xsawyerx, this PR makes MIME::Types optional, and uses Plack::MIME as default.
At the moment I did not test this in a machine without MIME::Types. Not sure if I should add a test file with `Test::Without::Module`.

Cheers

### CODE REVIEW

File: lib/Dancer2/Core/MIME.pm [hunk 1]
```diff
@@ -70,9 +65,10 @@ sub name_or_type {
 
 sub for_name {
     my ( $self, $name ) = @_;
+
     return
          $self->custom_types->{ lc $name }
-      || $self->mime_type->mimeTypeOf( lc $name )
+      || Plack::MIME->mime_type( lc "foo.$name" ) # simulate a filename
```
xsawyerx: AFAICT, you can just do ".$name" and it will work out too.

ambs: yes! I can change that if you wish :)

ambs:
Just a side note: the motivation for this was that plack middleware was serving a .xz file as text-plain. When discussing how to fix that with @xsawyerx we found it would be nice to also make MIME::Types optional, given we have a similar (although simpler) mechanism in Plack::MIME.

veryrusty:
I like where this is going! @ambs++

One question though; Is using the (functional) `MIME::Types::by_suffix` in the fallback going to "just work" in forked environments? 

The MIME::Types docs are quite specific about ensuring the type table is loaded prior to forking. My reading of the implementation is that by using `by_suffix`, MIME::Types won't load the type table until the fallback is called the first time. The comments in Dancer2::Core::MIME indicate this approach has caused trouble in the past (its loading the type table after fork). If the comments are stale then we should clean them up (aka delete them), otherwise we may need a closure around a MIME::Types object, something like:
```
BEGIN {
    if (try_load_class 'MIME::Types') {
        my $mt = MIME::Types->new;
        Plack::MIME->set_fallback(sub { 
            my $mime = $mt->mimeTypeOf($_[0]);
            defined $mime ? $mime->type : '' 
        });
    }
}
```

ambs:
We had just a line like:

``````
MIME::Types->new(...)
`````
to load the data table.
So, probably something like:
````````
BEGIN {
    if (try_load_class 'MIME::Types') {
        MIME::Types->new('only_complete' => 1);
        Plack::MIME->set_fallback( (MIME::Types::by_suffix($_[0]))[0] );
    }
}
`````````
might work as well. But won't mind using the closure. Just not sure if instead of returning the empty string we should probably return undef? Need to look into `Plack::MIME` to check how the fallback works.

veryrusty:
@ambs++ for the quick reworking!  :+1: for commit b70f9f2

Issue #825, opened 2015-01-02
Add a README.md


ambs:
Add a readme file to the root folder of the repository with minor information, like the link to our website.

PR #826, opened 2015-01-03
Add README

nunorc:
Add a minimalist README.md file, described in PerlDancer/Dancer2#825


xsawyerx:
Since we're using Dist::Zilla, we would just create it using the appropriate dzil plugin.

While I appreciate the work, we would then need to keep updating it. If we use a dzil plugin, it gets created automatically on a release.

Would you like to provide a pull request that adds the right plugin?

nunorc:
Workin on a new pul request, thank you.

xsawyerx:
@nunorc Thank *you*. :)

Issue #769, opened 2014-11-24
Rename show_errors as show_stacktrace?


xsawyerx:
`show_errors` actually means to show stacktrace. It's confusing. Wouldn't be better to rename it to `show_stacktrace`?

We can still support `show_errors`, but at least now there's a more descriptive configuration.

ambs:
Just for reference, this includes:

  * changing tests (if any)
  * change the default config file
  * update configuration
  * update the code itself.

And :+1: from me.

msouth:
One good thing about show_stacktrace is that most developers are going to immediately know that this variable is something about output meant for consumption by developers.

nunorc:
Pull request #829 only does the renaming part, I'm guessing that another part would be to support (at least for some time) both options, not sure what is the deprecation cycle on such matters.

xsawyerx:
@nunorc that was my exact comment just now on the PR. :)

xsawyerx:
Merged #1662.

Issue #743, opened 2014-10-31
Define a proper deprecation policy


ambs:
Hello

There was some "discussion" between me and @xsawyerx about deprecation of `param`. Not that we disagree on deprecating it, but rather, we disagree on what is deprecation.

So, this thread is to ask for a proper discussion on what is deprecation for Dancer2.

In my point of view, the definition of deprecation we had in the old days for Dancer was working well:

   1 First, the deprecated functionality/feature was marked as **soft deprecated**. It still worked, but issued a warning to the log files.
   2 Then, after a period of time, or after a fixed number of releases, the functionality/feature was marked as **hard deprecated**. It no longer worked, but an error was thrown on its usage.
   3 Again, after a period of time, or after a fixed number of releases, the functionality/feature disappeared completely.

This allowed us to clean code that we think is bad, buggy or that makes confusion to unaware developers. Also, it would allow us to, for example when deprecating a DSL keyword, to have it free again after some time, in order to fix it properly, for example.

In that "discussion" I had with @xsawyerx , his idea was to leave the `param` DSL keyword hanging in the code, but issuing a warning. As I've told him, for apps that are running, developers will not notice a feature is deprecated unless it gets to the **hard deprecated** stage, where it fails. In other words, it is very not probable they will look into the app logs unless it is failing. If we keep the code warning for ever, what we are doing is making our log files bigger (a lot!).

Therefore, I ask that we think about deprecation, and define a proper deprecation policy that should be followed by any feature, keyword or submodule that we want to change or remove.

VinceW:
Closely related to deprecation is in my opinion backward-compatibility.

The perl.org documentation is pretty clear about this at: http://perldoc.perl.org/perlpolicy.html#BACKWARD-COMPATIBILITY-AND-DEPRECATION

I think a Dancer2 policy should adhere to this policy because companies, project managers, change managers, developers, etc who are familiar with Perl (but not with Dancer2) are probably expecting the same compatibility/deprecation policy as Perl in general has.

In respond to the scenario outlined by @ambs:
> As I've told him, for apps that are running, developers will not notice a feature is deprecated unless it gets to the hard deprecated stage, where it fails. In other words, it is very not probable they will look into the app logs unless it is failing. If we keep the code warning for ever, what we are doing is making our log files bigger (a lot!).

[...]

PR #829, opened 2015-01-07
Rename show_errors to show_stacktrace

nunorc:
This plainly renames show_errors to show_stacktrace, regarding issue PerlDancer/Dancer2#769

### DIFF
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -308,7 +308,7 @@ to use them.
 Sessions are configured with a corresponding L</Session engine> section, as
 shown below.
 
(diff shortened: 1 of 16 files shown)


xsawyerx:
I will be happy to merge this once I merge @SnigdhaD's work on the docs, just to make sure we don't have to rebase over 3,000 lines of changes... again. :)

xsawyerx:
But also, and that's important to know, we can't simply switch. We need to deprecate it. That means produce a warning and still support it. Maybe we want to revisit #743 and sort that out soon?

xsawyerx:
Merged #1662.

PR #834, opened 2015-01-14
Improve scoping of app cleanup after dispatch (#762)

### LINKED ISSUE #762: 404s cause internal server error if you use a 404 template
This is related to the various scoping issues.
If you use a views/404.tt, and you try and go to a route that doesn't exist, you will get an internal server error(500), not a 404.

This is because the templating system requires the request and session, but by the time sub response_not_found is called, all the engines have had their request and session objects cleared by cleanup()

veryrusty:
If the request dispatch fails to match any route, render any error template before cleaning up the app context, then immediately return the rendered (error) response.

This is as close to using 'local' vars for scoping as we can get, without actually using localized vars.

Resolves #762.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1198,9 +1198,8 @@ DISPATCH:
         last;
     }
 
-    $self->cleanup;
 
-    # make sure Core::Dispatcher recognizes this failure
+    # No response! ensure Core::Dispatcher recognizes this failure
     # so it can try the next Core::App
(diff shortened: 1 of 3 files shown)


shumphrey:
looks good :thumbsup:

xsawyerx:
Excellent! :+1: 

I have a working concept of how to completely localize the request and response objects so they no longer exist in App, and allow for asynchronous code. Working PoC existing on a piece of paper. :)

veryrusty:
Woot. Two thumbs up. Merged :dancers:

PR #836, opened 2015-01-23
Delayed responses (asynchronous + streaming)

xsawyerx:
It's finally in. I went nuts. It worked.

I want to test it some more and write a few example usages for it. I also want to merge the keywords with prototypes branch because it will smooth out the syntax.

What do you think? :)

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1235,41 +1235,39 @@ sub _dispatch_route {
     $self->execute_hook( 'core.app.before_request', $self );
     my $response = $self->response;
 
-    my $content;
     if ( $response->is_halted ) {
-        # if halted, it comes from the 'before' hook. Take its content
-        $content = $response->content;
+        return $self->_add_content_to_response(
+            $response, $response->content,
+        );
     }
-    else {
-        $content = eval { $route->execute($self) };
 
+    $response = eval {
+        $route->execute($self)
+    } or do {
         my $error = $@;
-        if ($error) {
-            $self->log( error => "Route exception: $error" );
-            $self->execute_hook( 'core.app.route_exception', $self, $error );
-            return $self->response_internal_error($error);
-        }
+        $self->log( error => "Route exception: $error" );
+        $self->execute_hook( 'core.app.route_exception', $self, $error );
+        return $self->response_internal_error($error);
+    };
+
+    return $response;
+}
+
+sub _add_content_to_response {
+    my ( $self, $response, $content ) = @_;
+
+    defined $content or return $response;
+
+    # The response object has no back references to the content or app
+    # Update the default_content_type of the response if any value set in
+    # config so it can be applied when the response is encoded/returned.
+    if ( exists $self->config->{content_type}
+      && $self->config->{content_type} ) {
+        $response->default_content_type($self->config->{content_type});
     }
 
-    $response->has_content
-        and $content = $response->content;
-
-    if ( ref $content eq 'Dancer2::Core::Response' ) {
-        $response = $self->set_response($content);
-    }
-    elsif ( defined $content ) {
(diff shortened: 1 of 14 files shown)


veryrusty:
@xsawyerx++ I really like where this is going.

I note that the new `Core::Response::Delayed` class does not consume `Dancer2::Core::Role::Headers` like `Core::Response` does. If someone is using sessions with delayed responses, this is going to cause issues with the session header management in the `after_request` hook. As long as the response attribute "handles" the required methods, ( `headers`, `header` and `push_header` ), this should all "just work".

xsawyerx:
While trying to work on an example with @mickeyn I found a problem with this implementation.

I'm going to try and sort it out before I handle the issues you mentioned.

Let me dive back in.

xsawyerx:
I solved it.

I need to sort out @SnigdhaD's branch first and then I'll get back to this.

Basically I'm introducing another keyword, `delayed_sub` for now, which is used instead of a sub, and that gets localized again. Actually works. I wrote a Dancer2 + AnyEvent example. I'll write a few more.

xsawyerx:
@veryrusty @shumphrey I just added the last item (2 commits) necessary to make this complete.

I want to add some more tests and documentation, but the code is there.

Please to review. :)

xsawyerx:
The following is an example with `AnyEvent` and various timers.

    use Dancer2;
    use AnyEvent;

    my @timers = ();
    my $times  = 5;
    get '/status' => sub {
        delayed sub {
            print "Starting...\n";
            flush;
            content "foo\n";

            push @timers, AE::timer 2, 1, delayed sub {
                @timers == 1
                    and push @timers, AE::timer 0, 0.5, delayed sub {
                        content "baz\n";
                    };

                content "bar\n";
                if ( $times-- == 0 ) {
                    done;
                    undef @timers;
                    push @timers, AE::timer 2, 0, sub {
                        print "Done!\n";
                        undef @timers;
                    };
                }
            };
        };
    };

    __PACKAGE__->to_app;

You can run this code on the server with:

    plackup -Ilib eg.pl

(Assuming you're in the repository directory and have the code saved in a file `eg.pl`.)

You can run a check on it with:

    curl localhost:5000/status

What this code does is:
* Print "Starting" to the terminal (server).
* Flush headers and send "foo" to the client.
* Start a timer that will run in 2 seconds.
* That timer will send "bar" to the client every second.
* The first time this timer will run, it will create another timer which will run right away.
* The second timer will run instantly and will send "baz" to the client every half a second.
* Once the main timer runs enough times, it clears off all the timers, creates a new one, and closes the connection.
* The last timer (after all timers are done and the connection is closed) runs after two seconds, once, and prints "Done!" to the terminal (server).

What this showcases is:
* You can flush
* You can send stream to the user
* You can create asynchronous actions with `delayed` keyword
* You can still create asynchronous actions with regular `sub`, but they won't have keywords

I think that's good for now. :)

veryrusty:
This looks good. I wanted to ensure that a delayed (but not async) responses have after hooks setting headers correctly (eg. for session flushing). Easiest way was to include another test case (e3eba7b).

@xsawyerx I applaud your awesome craziness :D  Its a :+1: from me!

Possible things to look at later:
  * When the dispatch calls the [after_request hook](https://github.com/PerlDancer/Dancer2/blob/feature/delayed-responses/lib/Dancer2/Core/App.pm#L1193), the request object is passed as a param. That may be a delayed response object. We may want to stop passing that param, or pass `$Dancer2::Core::Route::RESPONSE` as the methods a delayed response object has are very different (no headers for example).

  * I haven't though this through, but I wonder if its worth considering having the `flush` keyword (which writes out response headers), to do things like session flushing the current after_request hook does. As I said, I haven't thought it through and maybe async and sessions don't need it..??

xsawyerx:
Merged!

Issue #840, opened 2015-02-01
No after hook executed after redirect means no session (e.g. dancr.pl example fails)


bigpresh:
Krys006 on #dancer was having issues with the tutorial dancr.pl from Dancer2::Tutorial - he could submit the login form, and saw the splash login message, but the login link remained (instead of logout), and there was no form to add an entry.

A bit of investigation showed that no `Set-Cookie` header was being sent, so there was no session in the next request (although - I'm puzzled how the splash message worked, now I come to think of it, but that's another matter).

The dancr.pl code does:

```perl
            session->write('logged_in' => true);
            set_flash('You are logged in.');
            return redirect '/';
```

Changing it to return some content, rather than a redirect, solved the issue - a Set-Cookie header was sent, establishing a session, and everything worked as expected after that.

A bit more digging showed that `after` hooks don't appear to run after requests which perform a redirect.  I can't see why that should be the case - and that sounds like the bug to fix?

(In dancr.pl I added:

```perl
hook after => sub {
    warn "After hook executing for request to " . request->path;
};
```

... I would see it execute for "/", then for "/login" (requesting the login form), but after submitting the login form, I'd see it execute only for "/" (when the browser re-fetches the main page), but *not* for "/login" when the login form was actually submitted.)

Possible that this is related to #811 (as a cause)?

bigpresh:
I note that the implementation of `redirect` claims:

```perl
    # Short circuit any remaining before hook / route code
    # ('pass' and after hooks are still processed)
    $self->has_with_return
        and $self->with_return->($self->response);
```

Seems that comment may be a lie :)  I'm not familiar enough with D2's low-level guts and Return::MultiLevel to figure it out at this moment, so I'd very much appreciate input from anyone who can help?

veryrusty:
@bigpresh The after hook is being called - there are explicit tests for this in the test suite.

The reported symptoms are due to the session implementation being "lazy". We try not to create a session object unless we "need" it; as such, the session object is only set in the engine objects if the `session` keyword is used.

The route for `/` in the tutorial does not call `session`, hence no session object is instantiated or set in the template engine before the template is rendered. (This is may well be what was being seen in #811.)

There has to be a more friendly way than expecting users to call `session` before rendering a template that involves rendering session data, but at the same time keeping the session implementation as lazy as possible.

bigpresh:
Well, my test in my test app suggested the after hook did *not* get called for the request that was redirected.

[...]

PR #842, opened 2015-02-03
Ensure session data is available to template engines

veryrusty:
Sessions are lazy. Some effort was put in to ensure we do not instantiate a session object if it was unnecessary. 

Template engines *should* add any session data to the hash of tokens before rendering. However, if the session object hasn't been created, there is no session data for template engines to add.  The (dancr) app in our tutorial fails because of this. See #840.

This Pr fixes the existing tests to trigger this situation, and ensures a session object is created (if needed) before processing a template.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -7,6 +7,7 @@ use Scalar::Util       'blessed';
 use Module::Runtime    'is_module_name';
 use Return::MultiLevel ();
 use Safe::Isa;
+use Scalar::Lazy       ();
 use File::Spec;
 
 use Plack::Middleware::Conditional;
@@ -667,6 +668,11 @@ sub template {
 
     my $template = $self->template_engine;
     $template->set_settings( $self->config );
+
+    # A session may exist but the route code may not have instantiated
+    # the session object (sessions are lazy). If this is the case, do
+    # that now, so the templates have the session data for rendering.
+    $self->has_session && ! $template->has_session && $self->setup_session;
 
     # return content
     return $template->process( @_ );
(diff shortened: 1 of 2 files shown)


lib/Dancer2/Core/App.pm
```diff
@@ -7,6 +7,7 @@ use Scalar::Util       'blessed';
 use Module::Runtime    'is_module_name';
 use Return::MultiLevel ();
 use Safe::Isa;
+use Scalar::Lazy       ();
```
xsawyerx: I don't see `Scalar::Lazy` used anywhere.

lib/Dancer2/Core/App.pm
```diff
@@ -668,6 +669,11 @@ sub template {
     my $template = $self->template_engine;
     $template->set_settings( $self->config );
 
+    # A session may exist but the route code may not have instantiated
+    # the session object (sessions are lazy). If this is the case, do
+    # that now, so the templates have the session data for rendering.
+    $self->has_session && ! $template->has_session && $self->setup_session;
```
xsawyerx: Could we use a logical `and` here at the end, so separate the *check* from the *action*?

    $self->has_session && ! $template->has_session
        and $self->setup_session;

lib/Dancer2/Core/App.pm
veryrusty: Oops - that's a leftover for another experiment that I missed removing :( Thanks for spotting it!

veryrusty:
@xsawyerx Removed the errant unused dependency and separated the logic in commit 2eb80e3.

That commit should get squashed with fb32840 before merging. Happy to do that if there are some thumbs up :)

xsawyerx:
Merged, thanks! :+1:

PR #843, opened 2015-02-04
Add missing attributes for test class

haarg:
The test relied on Moo classes without any attributes storing everything
given to their constructor.  This will be changed with a future version
of Moo.  Instead, create the appropriate attributes for the test class.


veryrusty:
Thanks @haarg for the PR and heads up for the upcoming Moo 2.0 changes !

PR #845, opened 2015-02-04
Feature/prefix consistency modified

veryrusty:
This is based on @xsawyerx awesome work from #817 with some extra tests.

In the case when the prefix is `undef` and the concatenated route regexp does not begin with a `/`, implicitly adds a leading '/' to make it a matchable route. i.e. both the following will now match the path `/`.
```
  prefix '/' => sub { get '' => sub {...} };
  prefix '/' => sub { get '/' => sub {...} };
```
This seems a reasonable balance between doing what was expected, but some may find it inconsistent.
In which case I'd be happy to die in the case when the prefix is undef and the (concatenated) route regexp did not begin with a `/`.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -1,9 +1,7 @@
 package Dancer2::Core::Request;
-
 # ABSTRACT: Interface for accessing incoming requests
 
 use Moo;
-
 use Carp;
 use Encode;
 use HTTP::Body;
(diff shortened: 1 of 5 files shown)


veryrusty:
<bump> It wasn't my intention to bike-shed/stall #817, but its been almost 2 months..

xsawyerx:
Yeah, I need to start reviewing everything. I got some vacation time this weekend and next, so I'll sit and work on some issues.

@veryrusty keep nudging. :)

xsawyerx:
Merged!

Issue #846, opened 2015-02-05
Add automated tests for changes in Issue 565


drmuey:
Will add details in a bit, just need an issue # for now.

drmuey:
issue 565

pull request 847

xsawyerx:
@drmuey care to update with the test? :)

drmuey:
you bet, I am in the middle of crazy town ATM so it might be a bit :)

xsawyerx:
Take your time. :)

PR #847, opened 2015-02-05
Issue 565: Fix HTTP Status template logic and documentation

drmuey:
* added: do, per doc, error_template after error specific static file
* fixed: error specific template file would result in a
* fixed: Config POD had incorrect template variables.
* added: more template info to Config POD including the previously
  undocumented $views/$error_code.tt
* removed unreferenced internal method _render_html
  (i.e. it is confusing and probably why this was overlooked since
   it calls a non-existent method error_template())
* other POD/comments that needs to note $views/$error_code.tt

Current t/ all pass, manual testing passed, new t/ tests via GH #846.

### DIFF
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -312,7 +312,7 @@ shown below.
 
 If set to true, Dancer2 will render a detailed debug screen whenever an
 error is caught. If set to false, Dancer2 will render the default error
-page, using $public/$error_code.html if it exists or the template specified
+page, using $public/$error_code.html if it exists, $views/$error_code.tt or the template specified
 by the C<error_template> setting.
 
 The error screen attempts to sanitise sensitive looking information
@@ -323,7 +323,7 @@ divulging details.
 =head3 error_template (template path)
 
 This setting lets you specify a template to be used in case of runtime
-error. At the present moment the template can use three variables:
+error. At the present moment the template (as well as $views/$error_code.tt templates) can use four variables:
 
 =over 4
 
@@ -331,15 +331,29 @@ error. At the present moment the template can use three variables:
 
 The error title.
 
-=item B<message>
+=item B<content>
 
-The error message.
+The error specific content (if any).
 
-=item B<code>
+=item B<status>
 
-The code throwing that error.
+The HTTP status code throwing that error.
+
+=item B<exception>
+
+The stringified exception (e.g. $@) if any.
 
 =back
+
+Keep in mind that 'content' and 'exception' can vary depending on the problem.
+
+For example:
+
+A 404 has an empty 'exception' and 'content' contains the URI that was not found. Unless you do the 404 yourself via  C<send_error("You chose … poorly!", 404);>, then 'content' is 'You chose … poorly!'.
+
+A 500 because of, say, dividing 0 by 0 will have an empty 'content' and 'exception like 'Illegal division by zero at …'.
+
+A 401 from C<send_error("You can not know the secret until you sign in grasshopper!", 401);> will have an empty 'exception' and 'content' will contain 'You can not know the secret until you sign in grasshopper!'.
 
 =head2 Logger engine
 
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -312,7 +312,7 @@ shown below.
 
 If set to true, Dancer2 will render a detailed debug screen whenever an
 error is caught. If set to false, Dancer2 will render the default error
-page, using $public/$error_code.html if it exists, $views/$error_code.tt or the template specified
+page, using C<$public/$error_code.html> if it exists, C<$views/$error_code.tt> or the template specified
 by the C<error_template> setting.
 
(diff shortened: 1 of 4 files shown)


drmuey:
For Issue 565

lib/Dancer2/Config.pod
```diff
@@ -312,7 +312,7 @@ shown below.
 
 If set to true, Dancer2 will render a detailed debug screen whenever an
 error is caught. If set to false, Dancer2 will render the default error
-page, using $public/$error_code.html if it exists or the template specified
+page, using $public/$error_code.html if it exists, $views/$error_code.tt or the template specified
```
xsawyerx: This should be in a POD C<> or F<> format.

lib/Dancer2/Config.pod
```diff
@@ -323,7 +323,7 @@ divulging details.
 =head3 error_template (template path)
 
 This setting lets you specify a template to be used in case of runtime
-error. At the present moment the template can use three variables:
+error. At the present moment the template (as well as $views/$error_code.tt templates) can use four variables:
```
xsawyerx: This too.

lib/Dancer2/Config.pod
```diff
@@ -331,16 +331,30 @@ error. At the present moment the template can use three variables:
...
+
+The stringified exception (e.g. $@) if any.
 
 =back
 
+Keep in mind that 'content' and 'exception' can vary depending on the problem.
+
+For example:
+
+A 404 has an empty 'exception' and 'content' contains the URI that was not found. Unless you do the 404 yourself via  C<send_error("You chose … poorly!", 404);>, then 'content' is 'You chose … poorly!'.
```
xsawyerx: Can we remove the Unicode ellipsis here for three dots instead?

lib/Dancer2/Config.pod
```diff
@@ -331,16 +331,30 @@ error. At the present moment the template can use three variables:
...
+
+A 500 because of, say, dividing 0 by 0 will have an empty 'content' and 'exception like 'Illegal division by zero at …'.
```
xsawyerx: Same.

xsawyerx:
Looks good to me. Thank you.

Would it be possible to get a test added to this? I would prefer having two actually: one for the class and one for the DSL.

lib/Dancer2/Config.pod
drmuey: good idea, will do, thanks

lib/Dancer2/Config.pod
drmuey: Sure, force of habit, I'm a unicode nut ;)

drmuey:
@xsawyerx Thanks for the code review, I just pushed the POD tweaks you requested.

I'll try to add the tests too but it may take a while as I am in the midst of a crazy few weeks.

Normally I'd be all about including the tests with the change but due to the time factor and since it was totally broken before and the testing noted in the initial commit being good, I opted to create #846 in order for me or someone else to follow up so that the fix could go in sooner but am happy to do it here if the delay is acceptable.

Let me know, thanks!

lib/Dancer2/Core/Error.pm
```diff
@@ -255,8 +254,13 @@ has content => (
                     }
                 );
             };
+
             # return rendered content unless there was an error.
-            defined $content && return $content;
+            if (defined $content) {
+                # Why is this necessary here and not with other templates e.g. like route templates?
+                utf8::encode($content) if utf8::is_utf8($content); # less confusing via String::UnicodeUTF8: $content = get_utf8($content) if is_unicode($content)
```
DavsX: Is this something to be conserned about? http://stackoverflow.com/a/14584079

drmuey: The stackoverflow thing is a bit alarmist. The use of utf8:: is ok as it is checking the state of the variable. 

If we don't do it the data is garbled, just seemed odd that it need done since it isn't with other template renderings in dancer2.

If the template thing is changed this code will still work right due to the check.

HTH

veryrusty: You can configure an app to use an encoding other than UTF-8. The "Why is it necessary" comment and having to do that twice points to this being the wrong place to be dealing with encodings.

For the content returned from a route (including templates), that content is added to the response object then the response is asked to encode it via `$response->content($content); $response->encode_content;`.

`Dancer2::Core::Error->throw` populates the response. Do any encoding there.

drmuey: It was needed in an app that was set to UTF8. We could only do it if the config is set to UTF8 to make it that much safer I suppose. I'll add that ASAP.

I also agree that it probably points to another bug that needs fixed that was never seen since the error feature has been broken. If that gets fixed this part will still be good to go.

veryrusty:
@drmuey @DavsX I've rebased and squished these commits and added another to deal with Error serialization and content encoding in Pr #849.

xsawyerx:
#849 merged. Thanks!

PR #848, opened 2015-02-09
Feature/keyword prototypes

veryrusty:
This is an alternative to #792. The main differences are:
  * keyword prototypes are "optional", well, the default is undef. There are some that consider *prototypes to be evil*.. I prefer this default.
  * the prototype is applied to the keywords coderef via `Scalar::Util::set_prototype`. This avoids eval of the code strings, but does require perl 5.8.1+
  * keywords registered from plugins may also have a prototype applied.

"Old style" plugins that `use Dancer2; use Dancer2::Plugin;` don't break. When the subs with the deprecation warning redefines the imports from Dancer2, we copy the existing prototype for those coderefs.

There are some very basic tests. But still no documentation. I ran out of tuits. :(

### DIFF
--- a/lib/Dancer2/Core/Role/DSL.pm
+++ b/lib/Dancer2/Core/Role/DSL.pm
@@ -76,6 +76,12 @@ sub _compile_keyword {
             $self->$keyword(@_)
         };
 
+    return $self->_apply_prototype($code, $opts);
+}
+
+sub _apply_prototype {
+    my ($self, $code, $opts) = @_;
+
     # set prototype if one is defined for the keyword. undef => no prototype
     my $prototype;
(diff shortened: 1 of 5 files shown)


xsawyerx:
Sweet! :+1: 

We can apply the `eval`ing I've added in the #792.

xsawyerx:
Anyway, I want to merge it. :)

veryrusty:
I don't follow the `eval`ing comment; one of the aims of this Pr was to not need `eval`s of stringified code. This does contain the same logic on `is_global` to remove the level of indirection too.

@xsawyerx Merge away!

xsawyerx:
Merged!

PR #849, opened 2015-02-15
Error template logic and encoding cleanup

veryrusty:
This is @drmuey commits from #847 (rebased onto current master and squished into a single commit) to fix the Fix HTTP Status template logic. There are still tests "outstanding" as per #846.

Those changes requires the Error `content` attribute to have byte (octect) strings to be for the http response. That implies the Error object needs to know about such things, hinting that we didn't have the abstraction/delegations correct; Error objects should only be concerned with character strings.

The response class handles serialization & encoding already, and `throw`ing an error populates a response object. The extra commit ensures that the Error's response object is generated with a serializer if the Error object has one, and calls `response->encode_content` when the error is thrown. Delegation FTW. 

Thanks also to @DavsX for picking this up in previous comments :)

### DIFF
--- a/Changes
+++ b/Changes
@@ -7,6 +7,7 @@
     * GH #843: Add missing attributes to Moo class used in tests. (Graham Knop)
     * GH #839: Correct "no_server_tokens" definition in production.yml.
                (Nikita K)
+    * GH #565: Fix HTTP Status template logic and documentation (Daniel Muey)
 
     [ ENHANCEMENT ]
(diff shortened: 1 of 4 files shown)


drmuey:
Thanks @veryrusty and @DavsX, this is a much better encoding bug fix than my temporary hack! As I mentioned previously I'd be unable to work on it as I am in  a crazy busy spot ATM so I was happy to pop in and find this done already :)

xsawyerx:
Merged, thanks! :+1:

Issue #852, opened 2015-02-21
Uninitialized variable in Core::Request


pants:
This appears from time to time in my logs:
> Route exception: Use of uninitialized value $value in string ne at /usr/local/lib/perl5/site_perl/Dancer2/Core/Request.pm line 664. in /usr/local/lib/perl5/site_perl/Return/MultiLevel.pm l. 97

This is part of `_build_cookies`. Malformed cookies (not sure where they come from, but regardless...) without an `=` will cause Dancer2 to die. It may be worth changing `Core/Request.pm` line 664 from:

    if ( $value ne '' )

to

    if ( defined $value and $value ne '' )

veryrusty:
@pants++ for reporting the issue! Could you put together a test case and your suggested fix in a Pr ?

pants:
Never done this process before. Please let me know if I did the fork/pull request/etc. correctly!

veryrusty:
@pants your fork/pr process was awesome! :dancers:

xsawyerx:
Solution merged, so I'm closing this.

Thanks. :)

PR #853, opened 2015-02-21
Fix an uninitialized variable on malformed cookies (without a value).

pants:
Add such a cookie to the request test. It should simply be parsed same as
the valued cookies.

Fixes issue #852


veryrusty:
@pants Looks good to me. Thanks!   :+1: 

( ping @xsawyerx - I don't see why this shouldn't get merged, the test case is worth keeping with even with your upcoming rewrite of the internals of Core::Request. )

xsawyerx:
Merged, thanks! :+1:

PR #854, opened 2015-02-24
Fix the after_template_render hook example.

assistcontrol:
The example given for the `after_template_render` hook doesn't actually alter the content.

`$ref_content` is local to the sub, so anything done to it won't show up to the caller. Need to dereference it to assign to it; the `after_layout_render` example has it right.


xsawyerx:
Merged, thanks! :+1:

PR #859, opened 2015-03-06
Add "accept" header to supported routing match options

perlpunk:
It would be nice if one could route depending on the accept header (for example if it contains an API version).

### DIFF
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -46,7 +46,7 @@ sub _check_options {
 
     my @supported_options = (
         qw/content_type agent user_agent content_length
-          path_info/
+          path_info accept/
     );
     for my $opt ( keys %{$options} ) {
         croak "Not a valid option for route matching: `$opt'"
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -272,7 +272,7 @@ keyword will return a reference to a copy of C<%+>.
 =head3 Conditional Matching
 
 Routes may include some matching conditions (on content_type, agent,
-user_agent, content_length and path_info):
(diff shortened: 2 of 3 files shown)


xsawyerx:
I personally don't like the matching options (a rarely-used feature, as far as I know), but this does make sense to add.

The only problem I have is that we haven't really discussed these options. They all seem to be HTTP headers, except `path_info`, which isn't. Also, other HTTP headers are not available, so I'm not sure what's really going on there.

@sukria, Could you comment on what the original intention was, and which direction you think it should go? Should all HTTP headers be supported? Is this a logical list rather than HTTP header-based?

@perlpunk, Is this disrupting you from still writing a path that checks the `accept` field and moves on if it isn't available?

    get '/' => sub {
        request->accept eq '...' or pass;
        ...
    };

perlpunk:
@xsawyerx, I'm trying to write methods depending on the API version in the accept header.
Of course I can do it manually by adding my own layer which routes to the correct methods. I just wanted to avoid this ;-)
mojolicious for example supports routing in a very flexible way by defining a custom subroutine.
Dancer only supports a list of fields of the request, and I'm not sure why the list of fields should be artificially limited...

xsawyerx:
So other than adding more fields, what else can we do to make it flexible?

perlpunk:
I would suggest something like this:
```
    my $route_w_accept = Dancer2::Core::Route->new(
        method  => 'get',
        regexp  => '/',
        code    => sub {'options'},
        options => {
            'callback' => sub {
                my ($req) = @_;
                my $accept = $req->header('Accept');
                return 1 if $accept =~ m{foo/bar-v3};
                return 0;
            },
        },
    );
```
Works fine for me.

PR #860, opened 2015-03-09
Eliminate a memory leak with hooks

xsawyerx:
When we compile hooks (`compile_hooks` method), we create a closure
that holds on to `$self`, causing a memory leak.

This wasn't detected in our memory leak test because we didn't
call `to_app` method on it (which finalizes the app and compiles
the hooks).

After calling `to_app`, the error shows itself. We then fix it
by closing over a weakened copy.

We also moved `$self->request` to `$Dancer2::Core::Route::REQUEST`,
which represents the current localized request.

(pinging @veryrusty to approve)

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,6 +2,8 @@
 
     [ BUG FIXES ]
     * Fixed another memory leak with compiled hooks. (Sawyer X)
+    * Fixed a memory leak with conditionally applied static middleware
+      (Russell Jenkins)
 
 0.159002  2015-03-03 19:21:21+01:00 Europe/Amsterdam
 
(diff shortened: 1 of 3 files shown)


veryrusty:
Now that we call `to_app` in the memory cycle tests, I would like to see the generated coderef used instead of `Runner->apps->[0]`. Doing so reported another memory cycle in the way the we were generating the static middleware and applying it conditionally to the apps psgi coderef. Hence the extra commit :)

@xsawyerx Looks good to me. :+1:  (and please review my extra commit :)

xsawyerx:
Excellent! @veryrusty++

xsawyerx:
Merged! :+1:

PR #864, opened 2015-03-22
calling plugin_setting from plugins in plugins

hvoers:
Tested with Dancer2::Plugin::DBIC

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -1,6 +1,6 @@
 package Dancer2::Plugin;
 # ABSTRACT: Extending Dancer2's DSL with plugins
-
+$Dancer2::Plugin::VERSION = '0.159002';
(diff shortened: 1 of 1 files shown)


racke:
Please reduce your patch to the changes in _plugin_setting.

hvoers:
I am following GitGuide.md

I am not patching '_plugin_setting'.

This is a patch of Dancer2::Plugin.pm

racke:
Yes, but remove the POD changes from your patch.

hvoers:
OK. I think you mean not my POD but the other pod that by magic appeared in the path?
Like =method ---> =head ?

racke:
Yes, DistZilla replaces =method and adds version when you build the package.

hvoers:
Ok. Discard this.
I will do 
GitGuide.md
from scratch ... :(

PR #868, opened 2015-03-26
Dancer2::Core::Error - fix broken accessor

cdmalon:
Dancer2::Core::Error::throw() attempts to set the response using set_response(), but response is 'rw', not 'rwp'.

### DIFF
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -295,7 +295,7 @@ sub _build_content {
 
 sub throw {
     my $self = shift;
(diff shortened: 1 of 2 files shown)


veryrusty:
Thanks for the Pr @cdmalon. Any chance of a test case for this code path before this is merged ?

veryrusty:
Sorry for taking so long to reply @cdmalon - I missed the notification of the extra commit with the test :(

That all looks good. Merged as 2fa76117.  Thanks again!  :dancers:

xsawyerx:
:+1: Thank you!

PR #869, opened 2015-03-29
send_file implementation with filehandle and streaming support

veryrusty:
Use delayed response to provide streaming and `send_file($filehandle)` support.

Almost every PSGI server supports streaming responses;  This PR makes streaming the default, and delegates the handling of writing the content into the actual response to the server. To make this "just work", you need to pass a filehandle as the response body; adding support for filehandles was almost a side effect :)   This should see a memory reduction in apps that use `send_file`, as the app no longer reads all the content of the static file into memory before generating the response.

`send_file` attempts to detect if it was passed a filehandle (code stolen from Plack), has the special case for IO::Scalar objects that D1 had (not sure if its actually needed), and then falls back uses the previous logic for finding the static file to be served. If the content is being streamed, a Response::Delayed object is returned, otherwise (as was previously done), the content is slurped into the current response, and that is returned.

Also cleaned up the last handful of references to the setting `public` and added a note to the migrations docs that D2 uses the setting `public_dir`.

This could probably use more tests. I wanted to get this out there for everyone to look over..
(plus March is almost gone and there is a #CPANPR challenge :P :dancers: )


hvoers:
Wow!
I like this.

xsawyerx:
Incredible!

veryrusty:
<blush> I'm standing on the shoulders of giants!

Note that if you use `send_file` in an app that has a serializer configured, this implementation does NOT serialize the content, whereas the current implementation does serialize (which I think is wrong).

veryrusty:
Added tests for `send_file($filehandle)` and `send_file($scalarref)`. :smile: 
(not planning on amending this further until its been reviewed)

xsawyerx:
Merged!

PR #871, opened 2015-03-31
Feature: use Safe instead of eval

zurborg:
Use safe compartment to prohibit unsafe code in deserializer

### DIFF
--- a/lib/Dancer2/Serializer/Dumper.pm
+++ b/lib/Dancer2/Serializer/Dumper.pm
@@ -5,6 +5,7 @@ package Dancer2::Serializer::Dumper;
 use Moo;
 use Carp 'croak';
 use Data::Dumper;
+use Safe;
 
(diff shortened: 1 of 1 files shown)


veryrusty:
@zurborg Thanks for the Pr!

Looks good to me. We'll need to add `Safe` to the dependency list in `dist.ini`.

hvoers:
Isn't Safe just a Core module?

veryrusty:
@hvoers so it is :)  I missed that Safe is a dual-life module.. I should have had coffee *before* looking at the Pr ;)

veryrusty:
Explicitly giving this a :+1:  Anyone else ?

zurborg:
`dzil listdeps` returns `Safe` - IMHO there is no need to add `Safe` to the dependency list by hand.

xsawyerx:
:+1:

xsawyerx:
Merged, thanks! :+1:

PR #874, opened 2015-03-31
Update Migration.pod fix #816

### LINKED ISSUE #816: Migration: session_domain was renamed to cookie_domain
As far as I'v just found out the session_domain configuration option has been removed
Instead of that there is one called cookie_domain. If that's the case, it would be nice if you could explain in the migration document, how the session management changed.

chenchenzhao000:
add session engine to the migration guide


racke:
I would be nice to have extra information about the session engine in the migration manual. However,
your patch is renaming the file itself and adding a bunch of empty lines.

xsawyerx:
I think the rename is an artifact of Git struggle syndrome. :)
We could ignore that and take the content.

:+1: on merging a clean version of this.

veryrusty:
I've added some extra POD formatting, and merged with commit 48790c3.

@chenchenzhao000 Thanks for the Pr. I took a guess as to your name to include in the `Changes` file; please let us know if it need correcting (and apologies if I got it wrong!).

racke:
￼:clap: to everyone involved.

xsawyerx:
@chenchenzhao000++
@racke++
@veryrusty++

:smiley:

chenchenzhao000:
@veryrusty Thank you for adding extra POD formatting for it,it looks better than before:)

PR #875, opened 2015-04-01
Fix tt: Internal Server Error 500

valerycodes:


xsawyerx:
@valerycodes Did you test it against the code example given in the original issue?

xsawyerx:
Merged! :+1:

Issue #416, opened 2013-08-30
Remove auto-discovery of appdir


xsawyerx:
The auto-discovery feature of Dancer2 is no longer necessary and impedes the ability to have the application directory separate from where the code library actually is.

It's magic when magic isn't necessary.

sakshee3:
@xsawyerx  which auto-discovery feature is left?  since you already removed the appdir (if there is) which is in the current dir.

veryrusty:
@sakshee3 See  [`Dancer2::Core::Role::HasLocation->_build_location`](https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Role/HasLocation.pm#L31)

xsawyerx:
To make a better note here for myself: If there is an `appdir`, why do we still run this code? I don't mind keeping it if there is no `appdir` configured.

PR #876, opened 2015-04-02
Working on Issue #416

sakshee3:
PR to solve #416


sakshee3:
@xsawyerx  I think we actually need it for some of the functionalities to work properly. What I can find from the code, and tests that there is no other way to refer to the appdir directly without finding it. What do you say ?

xsawyerx:
@sakshee3 I think the issue is a bit trickier.

Since some of the functionality requires is, it cannot be removed. At the same time, you cannot create an application that sits in "Location A" but is loaded from a file in "Location B". This is due to the auto-discovery.

This PR cannot solve it, which is why I intend to close it.

To solve this issue we would need to put more research and thought into this.

Issue #791, opened 2014-11-29
Logs refer wrong package name


ambs:
In recent Dancer2 installations, logs when running in devel mode (bin/app.pl), show:

```
[WN:57425] core @2014-11-29 14:29:47> looking for get / in /opt/perl/lib/site_perl/5.20.1/Plack/Middleware/RemoveRedundantBody.pm l. 14
```

But it should report the module .pm file, not the Plack one :-)
Or, if it is not relevant, just remove the module name :-)

ambs:
:(

sakshee3:
Can we use the idea of **public** and **views** directory, to find out who is calling the logger ? and then adapts the magic number ?

sakshee3:
or we can do something like this , using **while** loop, which will solve our problem.
http://docstore.mik.ua/orelly/perl/prog/ch03_010.htm

ambs:
if we go while, then run it only once to compute the magical number :dancers:

sakshee3:
so we can probably add that as a sub so that it stores that value and then we can get rid of computing while again and again

ambs:
Or even, try the 'magic number' that is most common, and if it is wrong (can we find out if it is wrong?) compute it? No idea... just random ideas.

xsawyerx:
I think computing it on compile is a valid idea. We had to adjust the "magic number" more than once in the past.

Perhaps something easier would be to have a magic number, but to write a test that verifies it is the correct number. If we change any of the stack and the number changes, the test fails, and we adjust the number to what the test says (since the test will compute it for us).

ambs:
@xsawyerx , if I understood correctly the problem, depending on how Dancer2 is being used, the number is different, and if so, just the test will not be enough.

sakshee3:
Isn't the following response correct:

when run with 
```
#!/usr/bin/perl 
use Dancer2;
set show_errors => 1;
get '/'         => sub {
    return "Hello World!";
};
start;
```
we should get:
```
>> Dancer2 v0.159003 server 3649 listening on http://0.0.0.0:3000
[main:3649] core @2015-04-04 21:06:16> looking for get / in /usr/local/share/perl/5.18.2/Plack/Middleware/RemoveRedundantBody.pm l. 14
[main:3649] core @2015-04-04 21:06:16> Entering hook core.app.before_request in /usr/local/share/perl/5.18.2/Dancer2/Core/App.pm l. 1251
[...]

PR #880, opened 2015-04-07
Cleanup: logging methods

veryrusty:
The DSL keywords for logging (`debug`, `info`, `warning` and `error`) as well as `Role::Hookable` now call the `log` method on an app object.  

While this doesn't solve #791, it reduces the number of cases we need to deal with for the "magic" number of call stacks to skip to report the caller from 4 to 2.


xsawyerx:
:+1:

xsawyerx:
Merged, thanks! :+1:

PR #881, opened 2015-04-11
Add cookie timeout example to YAML session

abeverley:
I dare say this is in the documentation somewhere, but I couldn't find it, and this is one place where I thought I might find the information.

### DIFF
--- a/lib/Dancer2/Session/YAML.pm
+++ b/lib/Dancer2/Session/YAML.pm
@@ -58,6 +58,7 @@ files in /tmp/dancer-sessions
       session:
         YAML:
           session_dir: "/tmp/dancer-sessions"
+          cookie_duration: 3600    # Default cookie timeout in seconds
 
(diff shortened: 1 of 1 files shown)


xsawyerx:
This should be part of the main documentation for Cookies:
https://metacpan.org/pod/Dancer2::Manual#cookies

Would it be possible to rewrite this in that section?

abeverley:
Okay, I'll look into it a bit more and hopefully come up with something useful. I just remember chasing my tail around various parts of the documentation, and not understanding what linked with what.

It definitely needs something in the session documentation, even if it's just to point out that cookie configuration values can go in the session configuration (?) and a link to the relevant cookie documentation. That was part of the confusion for me.

xsawyerx:
Merged, thanks! :+1:

For the record, the `cookie_duration` configuration is documented in the migration manual and in the `SessionFactory` role.

Issue #879, opened 2015-04-05
dzil build putting wrong $VERSION in Dancer2.pm


shadowcat-mst:
sherlock$ dzil build
[DZ] beginning to build Dancer2
[@Filter/ExtraTests] rewriting release test xt/release/distmeta.t
[@Filter/ExtraTests] rewriting author test xt/author/no-tabs.t
[@Filter/ExtraTests] rewriting release test xt/release/pod-syntax.t
[@Filter/PkgVersion] skipping lib/Dancer2.pm: assigns to $VERSION
[DZ] writing Dancer2 in Dancer2-0.159003
[MarkdownInRoot] overriding README.mkdn in root
[DZ] building archive with Archive::Tar; install Archive::Tar::Wrapper 0.15 or newer for improved speed
[DZ] writing archive to Dancer2-0.159003.tar.gz
[DZ] built in Dancer2-0.159003
sherlock$ grep VERSION Dancer2-0.159003/lib/Dancer2.pm 
$Dancer2::VERSION ||= '0.159000';
=head1 VERSION
sherlock$ 

So PkgVersion appears to be skipping it because it finds the version already present, thereby leaving the defaulted version in Dancer2.pm rather than the correct one. However -

16:09 < corecache> during the build, I get a notice from Filter/PkgVersion that 
                   it skips VERSION assignment for Dancer2.pm, which would 
                   actually result in the correct version string
16:14 <@mst> oh, interesting
16:14 <@mst> are you building the right commit?
16:15 <@mst> hrm, master is at 0.159000
16:15 <@mst> dist.ini sets final version
16:15 <@mst> yeah, that's fucked up
16:16 <@mst> sawyer: I'm unsure what your dzil config is supposed to be doing, 
             but I don't think that was it :)
...
10:49 <@sawyer> mst: the dist.ini provides a $VERSION variable and 
                lib/Dancer2.pm provides a default if it's not supplies
10:49 <@sawyer> *supplied
10:49 <@sawyer> it's for a test and should be removed
10:49 <@sawyer> but dist.ini is actually configured correctly

I can't understand how not replacing the placeholder version is correct, though.

xsawyerx:
The problem is not `dzil`. The problem is that there's a test that checks for a version. Some tests could ascertain the version so sukria hard-coded it for testing purposes.

We can remove, find the test that fails (if it still exists). and fix it.

shadowcat-mst:
[...]

PR #882, opened 2015-04-14
GH #879: Remove the requirement for two version numbers:

xsawyerx:
Dist::Zilla (dzil) was in charge of providing the version number for
the Dancer release, except some test scripts depended on having a
version number too (for example, when testing headers containing the
Server with version).

The issue GH #879 raises is that Dist::Zilla's PkgVersion doesn't
add the version so if dzil's version number is ignored and the one
in the .pm file is used instead.

This commit resolves it by removing the version number everywhere
and any place that requires a version will check for
VERSION || "DUMMY".

### DIFF
--- a/lib/Dancer2.pm
+++ b/lib/Dancer2.pm
@@ -13,11 +13,6 @@ use Dancer2::FileUtils;
 
 our $AUTHORITY = 'SUKRIA';
 
-# set version in dist.ini now
-# but we still need a basic version for
-# the tests
-$Dancer2::VERSION ||= '0.159000';
-
 our $runner;
 
 sub runner   {$runner}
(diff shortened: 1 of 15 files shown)


veryrusty:
I agree with the removal of `$Dancer2::VERSION` as the best fix for the reported issue.

However.. the repeated copying of the `DUMMY` string through the codebase is certainly not "DRY".
( :scream: its ugly ).

I'm not sure if there are side effects; but an alternative is to override the VERSION method, eg:
```
sub VERSION { shift->SUPER::VERSION || '0.000000_000' }
```
then use `Dancer2->VERSION` everywhere we need the version number.

xsawyerx:
@veryrusty I thought about doing that too. I was wondering if that would be too horrendous...

veryrusty:
@xsawyerx I implemented my idea in #883 for a comparison.

xsawyerx:
Closing this in favor of #883.

Issue #794, opened 2014-12-01
Exception in Core::Response when handling response from failed JSON deserialize attempt


dboehmer:
I'm writing an app handling API requests and need to handle invalid JSON/YAML input with a custom error message. Unfortunately Dancer2 behaves strangely when confronted with invalid JSON data.

I built the following example which should have a server side error and send a valid HTTP error back. Instead the code fails at the client side, too, when accessing the response's content.

```
use Modern::Perl;
use Plack::Test;
use HTTP::Request::Common;

{
    package App;

    use Dancer2;

    set logger => 'Console';
    set serializer => 'JSON';

    any '/' => sub {
        return request->data;
    };
}

my $app = App->to_app;

test_psgi $app, sub {
    my $cb = shift;

    say $cb->(POST '/', Content => '{"foo":42}')->decoded_content;
    say $cb->(POST '/', Content => 'invalid')->decoded_content;
}
```

Output from my terminal:

```
{"foo":42}
[App:852] core @2014-12-01 13:00:47> Failed to deserialize the request: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "invalid") at /usr/lib/perl5/site_perl/5.14/JSON.pm line 171. in (eval 166) l. 2
Internal Server Error

Modification of a read-only value attempted at /home/USERNAME/workspace/Dancer2/lib//Dancer2/Core/Response.pm line 72.
```

How to correctly handle invalid data (after all with Serializer::Mutable) on the server side? What's wrong here that decoded_content() fails?

seav:
> You should be able to handle wrong input using the serializer.before hook.

This is currently not possible. Looking at the `Dancer2::Core::Role::Serializer` code, the `engine.serializer.before` hook is only executed on serialization and not on deserialization.

PR #891, opened 2015-04-20
Have set() perform merge when passed a hashref value

jbarrett:
Apropos of some discussion on irc.perl.org#dancer, this change will merge config settings when `set()` or similar is passed a hashref value and the existing value is a hashref.

Use case is, I'm a bad developer and have set config values both in files and in code. I would like to set, say, an 'engines' or 'plugins' value in code without that rolling over existing settings. Instead of having to do, say:

```perl
    set(
        engines  => {
            ( config->{engines} )
            ? %{ config->{engines} }
            : (),
            serializer => {
                JSON => {
                    convert_blessed => 1,
                },
            },
        }
    )
```

You could simply:

```perl
    set(
        engines  => {
            serializer => {
                JSON => {
                    convert_blessed => 1,
                },
            },
        }
    )
```

Though I'm unsure how badly this breaks expected behaviour. Since merge is done for config files, I kind of expected it to be done for `set()` too.

### DIFF
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -157,7 +157,13 @@ sub _set_config_entry {
 
     $value = $self->_normalize_config_entry( $name, $value );
     $value = $self->_compile_config_entry( $name, $value, $self->config );
-    $self->config->{$name} = $value;
+
+    if ( ref $self->config->{$name} eq 'HASH' && ref $value eq 'HASH' ) {
+        $self->config->{$name} = Hash::Merge::Simple->merge( $self->config->{$name}, $value );
+    }
+    else {
+       $self->config->{$name} = $value;
+    }
 }
 
 sub _normalize_config {
--- a/t/config_settings.t
+++ b/t/config_settings.t
@@ -28,4 +28,21 @@ ok( setting('foo') == 43 && setting('bar') == 44,
     'set multiple values successful'
 );
 
+setting( plugins => {
+    Foo => {
+        option => 'value',
+    }
+} );
+set( plugins => {
+    Bar => {
+        setting => 'yes',
+    }
+} );
+
+is( config->{plugins}->{Foo}->{option}, 'value', 'plugins settings merged' );
+is( config->{plugins}->{Bar}->{setting}, 'yes', 'new plugin cfg set correctly' );
+
+set( plugins => 'ohno' );
+is( config->{plugins}, 'ohno', 'plugins value replaced when hashref not passed' );
+
 done_testing;


rleir:
Can we close this issue? (thanks -- Rick)

I was looking at  travis-ci Job #1757.5 log step 5790
"The command "dzil authordeps --missing | cpanm  --no-skip-satisfied || { cat ~/.cpanm/build.log ; false ; }" failed and exited with 1 during ."

That was 6 months ago. Would it pass if run again? Sorry to bother you, hope you have a moment for this, otherwise ignore. 

Your tests seem to run fine for me today, on my box. I get the same failures, whether or not your changes are in the build:
$ dzil test --no-author
...
 
t/dsl/parameters.t ..................................... 1/?     
    #   Failed test 'capture values are empty'
    #   at t/dsl/parameters.t line 303.
    #     Structures begin differing at:
    #          $got = HASH(0x3d4ecd0)
    #     $expected = undef
    # Looks like you failed 1 test of 10.

#   Failed test 'Captured route parameters'
#   at t/dsl/parameters.t line 350.
# Looks like you failed 1 test of 6.
t/dsl/parameters.t ..................................... Dubious, test returned 1 (wstat 256, 0x100)
...

Test Summary Report
-------------------
t/dsl/parameters.t                                   (Wstat: 256 Tests: 6 Failed: 1)
  Failed test:  6
  Non-zero exit status: 1
t/issues/gh-794.t                                    (Wstat: 0 Tests: 2 Failed: 0)
  TODO passed:   2
Files=132, Tests=1556, 20 wallclock secs ( 0.38 usr  0.06 sys + 18.34 cusr  1.59 csys = 20.37 CPU)
Result: FAIL
Failed 1/132 test programs. 1/1556 subtests failed.

jbarrett:
Bad time to rebase master...

xsawyerx:
I've locally rebased it against master and all tests run fine.

I'm a bit uncomfortable merging this for the simple reason that `set` is now explicitly setting a value to a key. If we change that, it actually becomes "add under certain circumstances, set another others" and that makes me uncomfortable. A different keyword that adds to a configuration option might be better, but I'm not sure we should introduce a new keyword just for that.

Always imagine there is someone trying to do the opposite that you're doing and ask yourself if your code accommodates that as well (or whether it should). It's plausible that there would be someone who overrides something in their config and instead it will now merge instead of overriding.

Tricky, this one.

rleir:
Is there a debug tool that traces, for each config var, where it has been set and where it has been overridden? (maybe someone has already done something like this) Would this be complicated task?

jbarrett:
Thanks for the feedback @xsawyerx.

I had considered that someone relies on the existing behaviour, overwriting a complete key, but had trouble envisioning a scenario where this would be useful. That, of course, doesn't mean nobody is relying on that behaviour.

"Add under certain circumstances, set under others" is perhaps more succinctly summarised as "merge when we can". I don't know if this is well defined enough.

Anyway, no biggie, I suspect the number of people doing complex run-time configuration is probably quite low (I'm in this boat as I am integrating with other Plack apps - the config is elsewhere).

Anyway, happy to close.

jbarrett:
Probably not worth expending more energy on this - thanks!

Issue #510, opened 2013-11-24
Hooks for plugins not registered


e11it:
It's possible to register hook only for last included plugin.
I have modified test to demonstrate this issue.

Please take a look: https://github.com/e11it/Dancer2/commit/9ed2da3bd2d8c211e7fa11dad78c5715f32b21ea

Test fails with:
```
t/plugin_syntax.t ............ 1/? Invalid hook name `third_hook' at /Dancer2/Core/DSL.pm line 143.
# Child (hooks in plugins) exited without calling finalize()
#   Failed test 'hooks in plugins'
#   at /System/Library/Perl/5.16/Test/Builder.pm line 252.
# Tests were run but no plan was declared and done_testing() was not seen.
```

drmuey:
this look like the cause of https://github.com/bigpresh/Dancer-Plugin-Database/issues/62 and https://github.com/PerlDancer/Dancer2-Plugin-Auth-Tiny/issues/4

drmuey:
+1 this is stopping a major project I am doing in its tracks so I will paypal $100 to whoever fixes this (I am not familiar enough w/ Dancer2 to offer a patch at this time) ;)

veryrusty:
I've made some progress on this, enough for the test from @e11it to pass. See my comment in #889.

veryrusty:
There is a working implementation for review in #892. :dancers:

veryrusty:
Resolved by #892.

Thanks to @e11it for reporting the issue and thank you to everyones' research, poking and prodding to get the fix/workaround merged.

There is still "brain pain" for anyone working through the plugin code - though we should now have the time to think about the reimplementation :)

drmuey:
Unfortunetely, #510 is not resolved by #892 :( its behaving like it was in my previous post

Dancer2 v.160000 and Dancer2::Plugin::Database v 2.12 still give me: 
   `Hook 'database_connected' does not exist at /Users/dmuey/perl5/perlbrew/perls/perl-5.16.0/lib/site_perl/5.16.0/Dancer2/Plugin/Database.pm line 45.`

commenting out line 45, `$dsl->execute_hook(@_);`, (since I am not using the hooks) allows it to work but that is a limited temporary hack :)

drmuey:
will open new issue since I can't re-open this one

drmuey:
Done it is, #905

PR #892, opened 2015-04-25
workaround: multiple plugins with hooks

veryrusty:
This is a "workaround" (not really a feature or a bugfix) to allow multiple plugins register and add hooks until the core devs settle and implement a rewrite of the plugin system.

Adds a registry of hooks (to Dancer2::Plugin), similar to the register of keywords. Though no checking to see if multiple plugins register the same hook has been implemented.

Registered hooks are added to the DSL object *after* the role composition dance, including adding the appropriate entries to the `hooks` hashref to `has_hook` just works. However, due to that role composition "tango", it was necessary to have consumers of `::Role::Hookable` provide _both_ the `supported_hooks` and `hook_aliases` methods.

Added tests from #510, plus a cut-down test that @ambs reported between Dancer2::Plugin::Deferred and Dancer2::Plugin::Database.

### DIFF
--- a/lib/Dancer2/Core/DSL.pm
+++ b/lib/Dancer2/Core/DSL.pm
@@ -11,6 +11,26 @@ use Dancer2::Core::Response::Delayed;
 
 with 'Dancer2::Core::Role::DSL';
 
+sub _add_postponed_plugin_hooks {
+    my ( $self, $postponed_hooks) = @_;
+
+    $postponed_hooks = $postponed_hooks->{'plugin'};
+    return unless defined $postponed_hooks;
+
+    for my $plugin ( keys %{$postponed_hooks} ) {
+        for my $name ( keys %{$postponed_hooks->{$plugin} } ) {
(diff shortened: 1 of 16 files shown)


lib/Dancer2/Core/DSL.pm
```diff
@@ -11,6 +11,29 @@ use Dancer2::Core::Response::Delayed;
 
 with 'Dancer2::Core::Role::DSL';
 
+sub hook_aliases { +{} }
+sub supported_hooks { () }
+
+sub _add_postponed_plugin_hooks {
+    my ( $self, $postponed_hooks) = @_;
+
+    $postponed_hooks = $postponed_hooks->{'plugin'};
```
ambs: It is guaranteed that that key exists?

ambs:
More than meets the eye at first. Need to give a second look, but my first pass say it looks good.

xsawyerx:
Merged!

ambs:
:+1: 
Now release! GO! :dancer:

Issue #893, opened 2015-04-27
Failing test with YAML.pm


xsawyerx:
This is an interesting failure. I've dug enough into it to find the problem, sort of. I'd like to dig more to find the exact problem even though I can already fix it by working around a problem in a third-party module.

We have a test to assure that we fail to read an invalid file, part of `t/config_reader.t`. We assume the underlying reader we use (`Config::Any`) will crash if it fails.

`Config::Any` delegates the reading to loaders. For YAML it uses `Config::Any::YAML` which internally loads `YAML::XS`, or `YAML::Syck`, or `YAML.pm` (in that order).

It would appear that both `YAML.pm` and `YAML::XS` crash but the one from `YAML` is swallowed. This is at least what I think happens, but I want to context switch and I haven't fully found out what happens.

Now, since on `YAML.pm` the error is swallowed, `Config::Any` doesn't crash and instead returns an empty string as the result. Because of this our code doesn't crash and assumes it returned a structure. When it tries to use the structure, it blows up.

I hope to get back to this soon (today-ish) and find out the root cause.

veryrusty:
The content of `t/config/environments/failure.yml` is actually valid YAML (string literal).

Changing that to something that is invalid YAML (such as inconsistent indentation) is sufficient to trigger the issue. Expect a Pr shortly :)

xsawyerx:
So YAML::XS crashes because?

veryrusty:
Other way around. YAML:XS was correctly parsing failure.yml. Its YAML.pm that's buggy and fails to parse the literal string. Config::Any hides the error and returns undef for the failed parse.

veryrusty:
Ahhh.. there are two issues interacting.
  1. As I've reported, our `failure.yml` environment is actually valid YAML. YAML.pm is buggy and does not parse it.
  2. Config::Any 0.25 was released 2015-04-23. It no longer reports parse errors in config files, which I consider a nasty bug. I've reported that [upstream](https://rt.cpan.org/Ticket/Display.html?id=104079)

veryrusty:
The upstream issue with Config::Any has been resolved with the release of v0.26.

Pr #895 covers anyone with Config::Any 0.25, and ensures we ship broken yaml (for all parsers). @xsawyerx can you have a quick look and do new release ?

veryrusty:
I've merged #895 to resolve this issue. The fix includes:
  * test environment `failure.yml` now contains invalid yaml for all parsers
[...]

PR #895, opened 2015-04-28
Fix for #893

veryrusty:
Ensure `failure.yml` environment config used in the tests is actually invalid YAML. Then catch the undef that Config::Any returns in that case before trying to use it as a hashref.


veryrusty:
Merged as dd137d8

Issue #899, opened 2015-05-04
Dancer2 fails to install 


jasonblewis:
Upgrading Dancer2 to Dancer2-0.160000 using cpanm failed. Not sure if this is a general issue or if its just my machine. 

I'm running on Debian 7.8 with perblrew installed perl v5.20.1. I attempted to upgrade Dancer2 using cpanm Dancer2

first try, cpanm recommended I install JSON::XS
https://gist.github.com/jasonblewis/7f216d8f62152298b911

second try:
https://gist.github.com/jasonblewis/258fa175e6862c1b3f31

veryrusty suggested I upgrade YAML after which Dancer2 successfully installed.

veryrusty was able to reproduce the error by downgrading to YAML 0.85

veryrusty:
I was able to replicate on OS X by downgrading to YAML v0.85.
YAML 0.84 and 0.86 to 0.90 all pass (and so do the recent 1.1x versions).

Note that YAML 0.85 was very short-lived before having a bug-fix release in v0.86.

veryrusty:
Anyone installing Dancer2 is unlikely to have issues updating old YAML versions.. Should we specify YAML v0.86 as the minimum version for YAML ?

veryrusty:
Ping @xsawyerx - Can you see any issues in specifying the minimum version of YAML as v0.86 to avoid the broken version (v0.85) ?

shlomif:
I can reproduce it with YAML-0.85 too on Mageia Linux x86-64 5.

PR #902, opened 2015-05-05
Dep on YAML set to minimal 0.86.

shlomif:
Fix https://github.com/PerlDancer/Dancer2/issues/899 : YAML-0.85 caused
a breakage in the installation.


veryrusty:
Thanks @shlomif !  :dancers: 

Modified so the code doesn't specify the YAML version (we don't do that for other modules) and merged as f9cb53e9.

PR #908, opened 2015-05-15
Change references to Dancer to Dancer2 in the DBIC section.

simbabque:
Since this is D2, especially the command line tool reference to 'dancer -a bookstore' should be 'dancer2 -a bookstore' as new users might not have dancer and would get confused or discouraged. On the other hand, older users would end up with a Dancer (one) app which might be even worse.


veryrusty:
Looks good. :+1: @simbabque++

Merged as 150080be (after reformatting commit message).  Thanks!!

xsawyerx:
:+1:

PR #911, opened 2015-05-19
Base Dancer2::Core::Request on Plack::Request

xsawyerx:
I couldn't see how to do the majority of this in more than one commit.

This commit bases D2::Core::Request on Plack::Request instead of delegating to an internal Plack::Request object.

This allows to utilize the existing implementation of many PSGI perspectives already available in Plack::Request while adding our own variations (such as using "is_behind_proxy") without having to keep another object around and delegating to it.

Here are also a few benchmarks showing about 13.57% increase in speed as a result of this change.

Before:

    $ perl bench-request.pl
    Ran 12 iterations (2 outliers).
    Rounded run time per iteration: 9.741e+00 +/- 2.9e-02 (0.3%)

After:

    $ perl bench-request.pl
    Ran 11 iterations (1 outliers).
    Rounded run time per iteration: 8.419e+00 +/- 3.3e-02 (0.4%)

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -58,7 +58,7 @@ _EVAL
 our $XS_URL_DECODE         = try_load_class('URL::Encode::XS');
 our $XS_PARSE_QUERY_STRING = try_load_class('CGI::Deurl::XS');
 
-our $_count = 0;
+our $_id = 0;
 
 # self->new( env => {}, serializer => $s, is_behind_proxy => 0|1 )
 sub new {
@@ -81,7 +81,7 @@ sub new {
     }
 
     # additionally supported attributes
-    $self->{'id'}              = ++$_count;
+    $self->{'id'}              = ++$_id;
     $self->{'vars'}            = {};
     $self->{'is_behind_proxy'} = !!$opts{'is_behind_proxy'};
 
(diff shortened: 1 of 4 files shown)


xsawyerx:
@veryrusty, @shumphrey, @mickeyn, @yanick ?

xsawyerx:
Rebased. If no one says anything, I'm merging it tomorrow (US Eastern time).

veryrusty:
I've been slowly working through it!

First comment:
* Does the request class really need the `headers_to_array` method? If so, can we avoid the duplication and include it as a role (may involve splitting the existing headers role..)

xsawyerx:
The idea is that `headers_to_array` is in a Moo Role but this PR removes the Moo dependency altogether.

veryrusty:
Use `Role::Tiny` directly.

xsawyerx:
We could. Then again, I think that method might just be redundant and unnecessary and we could remove it altogether. That's why for now I just copied it over.

veryrusty:
I think it is redundant. Converting headers to an array format like that only  makes sense if you are generating a PSGI style response. Not something you'd do from request headers!

xsawyerx:
Done. Rebased and pushed.

xsawyerx:
Merged!

veryrusty:
Yay! @xsawyerx++  Thanks!

PR #915, opened 2015-05-27
Improve error info when no HTTP_X_FORWARDED_HOST

abeverley:
If behind_proxy is set but no HTTP_X_FORWARDED_HOST header is received (such as if the proxy's request was from a client using HTTP/1.0), then Dancer dies with a rather mysterious "Use of uninitialized value in split" error message. This patch improves the error message to help with
debugging.

Incidentally, this problem is confounded when subsequent requests are made from the proxy (even with HTTP/1.1) as the environment variables have already been set. I'm not sure whether this requires a
further patch? If not, this patch at least makes it easier to fix the problem.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -187,7 +187,11 @@ sub host {
     my ($self) = @_;
 
     if ( $self->is_behind_proxy ) {
-        my @hosts = split /\s*,\s*/, $self->env->{HTTP_X_FORWARDED_HOST}, 2;
+        my $env_host = $self->env->{HTTP_X_FORWARDED_HOST}
+            or croak "No HTTP_X_FORWARDED_HOST received from proxy. "
+                ."It's possible that either this request or an earlier "
+                ."request by another client were made using HTTP/1.0.";
+        my @hosts = split /\s*,\s*/, $env_host, 2;
         return $hosts[0];
     } else {
         return $self->env->{'HTTP_HOST'};


xsawyerx:
I'm not sure I understand the second example.

:+1: to the PR. I just want to understand the second example you mentioned before merging.

abeverley:
Sorry do you mean the example in my comments or the actual code?

Correct me if I'm wrong, but what appears to happen is that the first request to a Dancer object initiates the environment variables, which includes HTTP_X_FORWARDED_HOST if applicable. Subsequent requests to the object reuse the existing environment variables (under Starman at least?). The problem is that if HTTP_X_FORWARDED_HOST was missing from the first request (e.g. HTTP/1.0) then even if subsequent requests *do* send HTTP_X_FORWARDED_HOST, it still results in the error, which then needs the server restarting.

This last point is something I was wondering whether needs fixing separately, because at the moment it's possible to break all the plack children by sending requests without HTTP_X_FORWARDED_HOST, which is effectively a denial of service. Of course, it can be fixed by forcing the environment variable in the proxy.

Make sense?

veryrusty:
The `X-Forwarded-Host` header is a "extension" to the HTTP specs (both 1.0 and 1.1). The comment in the "improved" error is misleading.

I'd rather see logic like the following (which is similar to what Plack::Middleware::ReverseProxy does):
```
if ( $self->is_behind_proxy ) {
    if ( $env->{HTTP_X_FORWARDED_HOST} ) {
        my @hosts = split /\s*,\s*/, $self->env->{HTTP_X_FORWARDED_HOST}, 2;
        return $hosts[0];
    }
}
```

[Disclaimer: I prefer to use the reverse proxy middleware(s) rather than the `is_behind_proxy` option. ]

veryrusty:
oops, that needs to be `if ( $self->is_behind_proxy && $env->{HTTP_X_FORWARDED_HOST} ) {...}`

abeverley:
Okay, that's a fair comment and makes sense. I'm still a little concerned of silent failure though, so at a minimum I think there should be a note in the manual (I'm happy to draft one as an update to this patch), and possibly even a warning logged as appropriate.

This still leaves the other problem I described, which is that a request made without that header seems to cause all subsequent requests to fail (even subsequently with that header). I observed this locally, but haven't had a chance to investigate fully.

abeverley:
I've done a bit more testing with this, and believe that this is a denial of service vulnerability.

The "env" attribute is only set on the initial building of the request object, but this is not done for every client request. This means that if the client request that causes the initial build is missing HTTP_X_FORWARDED_HOST, then all subsequent requests also have this environment variable missing, even if they send the header. Consider this:

1. The first client sends a request with the ```Host``` header missing. The proxy forwards the request without HTTP_X_FORWARDED_HOST.
2. The environment is populated without HTTP_X_FORWARDED_HOST set. As such, the aforementioned bug is triggered (```Use of uninitialized value in split```).
3. A completely different client now makes a request *with* a ```Host``` header. The proxy passes this header using HTTP_X_FORWARDED_HOST.
4. Because the env property is already set, this HTTP_X_FORWARDED_HOST value is not set in the request object. Dancer then dies with another ```Use of uninitialized value in split```.
5. All further requests die because of (4) until the server is restarted.

At the time of writing, you can see an example of this here: http://ncc.gads.ctrlo.com/ (I have made a single request to this without a ```Host``` header).

The original patch is now invalid. I'd be happy to provide a better one if someone can point me in the right direction. I think the environment should be populated for every client request regardless?

abeverley:
Just to clarify, Rusty's patch will fix the denial of service, but it will still leave a bug whereby ```host``` will not return the correct value under the circumstances described.

xsawyerx:
The first issue was handled and released. This is next.

veryrusty:
A variation of my suggestion + tests were merged from #930. I think that is sufficient to resolve this too. @abeverley any change you can verify if that is the case ?

A new request object **is** constructed for every new (external) request, based on the psgi env for the incoming HTTP request. I'm not aware of how your app is deployed, but based on your description above, you _may_ want to update to Moo 2.xx (if you haven't done so already) so fatal warning are not enabled by default.

abeverley:
Thanks @veryrusty. My app is deployed using a pretty standard Starman/Plack setup. Is that the reason that the request object is not being reconstructed for each request?

Thanks for the pointer about Moo 2. However, that would still leave me with the problem of `host` not being set correctly, if the first request happens to have been from a HTTP/1.0 client, so I'd still quite like to get to the bottom of this.

Of course, a quick and easy fix would be to force HTTP_X_FORWARDED_HOST using the proxy (which I'll probably do anyway), but I'm concerned that others will fall at this and be left with some difficult debugging and strange behaviour.

veryrusty:
@abeverley I have finally managed to reproduce the symptoms you described.. Pr to fix coming soon.

veryrusty:
See #934 :D

abeverley:
Awesome, that fixes it, thanks very much. That would have taken me months to figure out myself!

abeverley:
As for the original issue, I still think #930 needs some more work. Either a blank `host` when the header is blank (my preference), or better documentation to warn that `host` will contain the local host name in those circumstances. I'm happy working on either.

veryrusty:
Better docs!  They will help if you are using the built-in `behind_proxy` feature or using Middleware for header munging behind a proxy.

abeverley:
No problem, see #935.

PR #916, opened 2015-05-27
fix testing with cookies example to use full URLs

SysPete:
`$cookie_jar->add_cookie_header` is a noop if request only contains path.

Having spent some time banging my head against the wall I'd prefer if future users don't have to go through the same pain.


xsawyerx:
Merged, thanks! :+1:

PR #917, opened 2015-05-29
fix PUT request with JSON array in body

abdrabo:
PUTting JSON array fails with internal server error because the deserialized
data are not key/value pairs and can not be merged to the request->params.

The data are available in request->data but only key/value pairs can be merged
to request->params.


xsawyerx:
A different fix was applied which works on all parameters.

Thank you for the contribution anyway. :)

Issue #921, opened 2015-06-04
Dancer2-0.160002 test failure with Plack v1.0036


nanis:
I haven't tried to diagnose this, but here some information:

- I have a working 0.160001 install
- <pre>C:\...\Dancer2-0.160002> perl -MPlack::Response -E "say $Plack::Response::VERSION"
1.0036</pre>
- Will link to test output on CPANTesters when it becomes available.
- Below, I added a call to dump the `$plack` object right before the problem:

<pre>C:\Users\sinan\.cpanm\work\1433418642.3540\Dancer2-0.160002> perl -Iblib\lib t/classes/Dancer2-Core-Response/new_from.t
1..3                                                                                                                   
ok 1 - Dancer2::Core::Response->can(...)                                                                               
    # Subtest: new_from_array                                                                                          
    1..4                                                                                                               
    ok 1 - An object of class 'Dancer2::Core::Response' isa 'Dancer2::Core::Response'                                  
    ok 2 - Correct status                                                                                              
    ok 3 - Correct content                                                                                             
    ok 4 - All headers correct                                                                                         
ok 2 - new_from_array                                                                                                  
    # Subtest: new_from_plack                                                                                          
    1..5                                                                                                               
    ok 1 - An object of class 'Plack::Response' isa 'Plack::Response'                                                  
    # --- !!perl/hash:Plack::Response                                                                                  
    # body: Bar                                                                                                        
    # headers: !!perl/hash:HTTP::Headers::Fast                                                                         
    #   content-type: text/plain                                                                                       
    #   x-test: Val                                                                                                    
    # status: 200                                                                                                      
coercion for "headers" failed: Not an ARRAY reference at blib\lib/Dancer2/Core/Role/Headers.pm line 16.                
    # Child (new_from_plack) exited without calling finalize()                                                         
not ok 3 - new_from_plack                                                                                              
#   Failed test 'new_from_plack'                                                                                       
#   at c:/opt/perl-5.22.0/lib/Test/Builder.pm line 279.                                                                
# Looks like you failed 1 test of 3.                                                                                   
# Looks like your test exited with 255 just after 3.</pre>

<pre>Summary of my perl5 (revision 5 version 22 subversion 0) configuration:
   
  Platform:
    osname=MSWin32, osvers=6.3, archname=MSWin32-x64-multi-thread
    uname=''
    config_args='undef'
    hint=recommended, useposix=true, d_sigaction=undef
    useithreads=define, usemultiplicity=define
[...]

PR #922, opened 2015-06-04
Feature/http headers fast

veryrusty:
Plack v1.0035+ changed from using HTTP::Headers to HTTP::Headers::Fast, breaking our test suite.
HTTP::Headers::Fast has the same interface as HTTP::Headers, is still pure perl, but faster. Lets use it too!

Change to using HTTP::Headers::Fast. But allow both HTTP::Headers::Fast or HTTP::Headers objects in Dancer2::Core::Role::Headers to maintain back-compat with older Plack versions.

Resolves #921.

PR #923, opened 2015-06-05
Changes accordingly with new Plack release (mst++)

ambs:
It seems something changed on last Plack. This fixes it. Thanks to @shadowcat-mst

### DIFF
--- a/lib/Dancer2/Core/Role/Headers.pm
+++ b/lib/Dancer2/Core/Role/Headers.pm
@@ -5,6 +5,7 @@ package Dancer2::Core::Role::Headers;
 use Moo::Role;
 use Dancer2::Core::Types;
 use HTTP::Headers;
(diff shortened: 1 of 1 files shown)


ambs:
Hopefully fixes #921 with less code

veryrusty:
"something changed" is somewhat vague..

Plack moved to use HTTP::Headers::Fast as its more performant. We should do the same, which is what #922 is about. I'd be happy to see this combined into #922 if a few extra characters of code is all we are concerned about.

veryrusty:
Combined this into #922. The bless + isa check is cleaner than the regex update. Thanks @ambs!

ambs:
Sorry, @shadowcat-mst asked me to create a fix, and was in a hurry. Really did not have time to make it better explanation!

ambs:
@veryrusty++ as usual

PR #929, opened 2015-06-11
Default session_dir is appdir/sessions

hvoers:
Just a bug.
Can be worked around by putting an explicit session_dir in config.yml

### DIFF
--- a/Changes
+++ b/Changes
@@ -3,6 +3,9 @@
     [ DOCUMENTATION ]
     * Sanitize Changes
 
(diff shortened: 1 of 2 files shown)


veryrusty:
Thanks for the PR!

Unfortunately this can't be merged - it should cause test fails. However this PR does highlight there are none covering Dancer2::Core::Role::SessionFactory::File :(

@hvoers would you like write some tests, or extend some of the existing sessions tests to cover Dancer2::Session::YAML ?

PR #930, opened 2015-06-11
Check if ENV HTTP_X_FORWARDED_HOST is available before using it

melo:


lib/Dancer2/Core/Request.pm
```diff
@@ -186,7 +186,7 @@ sub BUILDARGS {
 sub host {
     my ($self) = @_;
 
-    if ( $self->is_behind_proxy ) {
+    if ( $self->is_behind_proxy and $self->env->{HTTP_X_FORWARDED_HOST} ) {
```
ambs: probably use `exists` instead of autovivificating it?

veryrusty:
Thanks @melo!  Merged with @ambs suggestion plus tests as 765261a.  :dancers: 
( This is _almost identical_ to the suggestion in #915 :smile: )

ambs:
:+1:

abeverley:
The problem with this is that if the header is missing (i.e. the client sent a HTTP/1.0 header without host), then the internal host name is returned in `host`.

I would have thought it would be better returning an empty string in these circumstances, which I presume is the behaviour you would get if you were *not* running behind a proxy and the client sent no host header.

abeverley:
See #935.

PR #931, opened 2015-06-11
Draw attention to precedence of parameter munging

Camspi:
In lieu of lobbying that the precedence from highest to lowest be route, post, query string, here is some documentation for the Manual to call attention to this not-so-obvious behavior that may have curious side affects.

Namely, when a route is defined as /:foo/bar, the author may expect (reasonably), that param('foo') will equal ':foo' in the route.  In all cases where 'foo' has not been POSTed, this is true.  However, per Dancer2::Core::Request the data munging prefers the POST parameters which makes the route matching parameters reliable only for method dispatch and not as a named parameters in and of themselves.

@xsawyerx and @mickeyn, thanks for hearing my original plea at YAPC::NA to get the precedence modified. :)

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -2530,7 +2530,16 @@ I<This method should be called from a route handler>.
 It's an alias for the L<Dancer2::Core::Request params
 accessor|Dancer2::Core::Request/"params($source)">. It returns a hash (in
 list context) or a hash reference (in scalar context) to all defined
-parameters. Check C<param> below to access quickly to a single parameter
+parameters.  
+
+By default, the values of C<POST> parameters overwrite the values of
+parameters defined by route declarations. Parameters defined by the route
+declarations overwrite query string parameters defined by C<GET> requests. See
+L<Dancer2::Core::Request params accessor|Dancer2::Core::Request/"params($source)">
+for optional arguments to C<params> to return values from specific
+sources (i.e., route, query string, post body).
+
+Check C<param> below to access quickly to a single parameter
 value.
 
 =head2 param


Camspi:
Four choices to resolve:
regex
regex with capture
splat
params('route')

^your verbal notes for yourself @xsawyerx :)

ambs:
:+1:

Camspi:
Published  Dancer2::Plugin::ParamKeywords to add a little syntactic sugar around consuming parameters from explicit sources.

xsawyerx:
Having had time to go over the code more and consider this (something I couldn't do on the spot), I think there's room to consider changing the order. It doesn't make sense to match using the route spec but override the variables created by it.

@veryrusty @ambs @mickeyn @shumphrey @yanick @racke, would any of you be adverse to changing the order of matching? It would address the following issue:

    post '/:name' => sub {
        my $name = param('name'); # $name = "nooo"
    };

    # curl -X POST -d 'name=nooo' http://localhost:5000/yesss

Here are the situations I can think of:

* Path defined with a named parameter (`/:name`) with data `name=Sawyer`: If we change the parameters order, it will be fine.
* Path defined without a named parameter (`/*`) with data `splat=Sawyer`: Now `params()` would contain a key called `splat` with the overwritten value instead of what was matched. That will also be fixed.

The only problem that remains, as far as I can tell, is that if you provide serialized data under a key that matches the named parameter in the path, that will get deleted. That's a problem.

What I'm thinking is either warning when these match, or even better, to warn a user to not use a named parameter (or a splat, even, I guess) when having data (rather than a "get" request).

This is a tricky topic.

xsawyerx:
Seems like this is how Rails/Rack handle it too: http://stackoverflow.com/questions/8510235/rails-query-parameter-vs-post-parameter/8512276#8512276

xsawyerx:
And another point of reference: https://dev.ckeditor.com/ticket/727

shumphrey:
if I were writing something with conflicting parameters, I *think* would expect the route/splat parameters to take precedence.
I am fine with the order changing.
As you say, tricky.

xsawyerx:
#993 was merged, which addresses this, and it documents this change in the migration manual.

Thanks for raising this issue, and sorry it took this long to handle.

PR #934, opened 2015-06-13
Bugfix/dispatcher scope cleanup

veryrusty:
If `Dancer2::Core::Dispatcher` is being used to dispatch over multiple apps, the dispatcher wasn't always ensuring that the internal state got cleaned up.

This could be triggered by `die`ing in a route or hook (see included test), or if fatal warnings were enabled (see description in #915).  I suspect this will clean up some of the with_return scoping issues that reported late 2014 (which I can't find the issue / prs for).


xsawyerx:
I didn't note it in the ticket, but this was merged! :)

PR #935, opened 2015-06-14
Add more detail to behind_proxy docs

abeverley:
This PR adds a little more detail about things to be careful of when using ```behind_proxy```.


xsawyerx:
Merged, thanks! :+1:

Issue #844, opened 2015-02-04
PUT request with JSON array in body stops the request processing


utrack:
Dancer2 version: current CPAN

PUTting an JSON array fails with error:

    Internal Server Error
    Not a HASH reference at /usr/share/perl5/site_perl/Dancer2/Core/Request.pm line 483.

`_body_params` is expected to be hashref, when serializer returns an arrayref. 
Some proper warning would do.

veryrusty:
When a serializer is set we try and serialize any data into the response. We really should handle deserializing any data type from the request body too.

If the deserailized data is a hashref, merge it into the request params. Otherwise we could
  * turn it into a key-value pair, and merge something like  `{ data => $deserialized_body_content }` into the request params, or
  * don't merge the deserialized data into the request params, leaving it in request->data. (my preferred option)
  * throw an error as @utrack suggested.

Does anyone else have an opinion on how this should be handled ?

corecache:
As for option 2, the original request-> data gets lost when beeing overriden by the deserialized data? or did you mean to put the deserialized data into a separate attribute?

veryrusty:
@corecache The "original" request body is always available via `request->body`.  `request->data` is a separate attribute that is populated with deserialized content

For the second option, an `if ( ref $data eq 'HASH' )` block wrapped around [lines 280-281 of Dancer2::Core::Request](
https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Request.pm#L280) should almost fix it (there is some documentation to clean up and a test or two to write as well :smile: )

corecache:
@veryrusty That would at least make sure that everything which should get merged down into _body_params is a hash. This still leaves people who want to parse a JSONArray with having to access request->data instead of request->params/params.

1) Is there a valid and convenient way to have JSONArray acessible via params? 
2) is it needed afterall? 
3) What would a solution look like? I can imagine sth like _body_params->{jsonarray/items/...} = $data in the case of $data is an ARRAY - but that feels somehow dirty as it fiddles with implicit parameter defintion during requests which could collide with user parameters.

xsawyerx:
I think we should start with defining what parameters are. In my opinion, they are essential key/value pairs. This means that any request to get parameters from a non-hash structure should fail. Having it available in `request->data` or `request->body` is fine to me and make sense.

This gives us a clear path of handling serialized content that isn't key/value pairs.

xsawyerx:
[...]

Issue #740, opened 2014-10-30
Replace param() keyword


xsawyerx:
`param` keyword is tricky. It says "please fetch the values of a single parameter key", while it's still susceptible to having multiple values vs. single values.

This means that:

    my $value = param('key');

might return a single value (if a single value was provided by the user) or a single value of arrayref (if multiple values were provided by the user).

This is not just confusing, it's a way to fuck up storage.

If you think arrays will save you, you're still screwed:

    my @values = param('key');

This will return one item in the array with the value if only one is sent, but one item in the array as a arrayref for the values if more than one value was sent.

Moving `params` to `Hash::MultiValue`, we can deprecate this keyword.

    # before
    my $value = param('key');

    # after
    my $value = params->get('key');

We don't actually have to remove `param`. We can still keep it for a long while with a warning.

We would like to get comments before actually doing it. This is why it's labeled as a question.
("We" = @mickeyn and I)

yanick:
What about `param_array()` as implemented in Dancer 1? It's probably isn't the most glamourous nor clever function name  ever, but what it doesn't have in pizzazz, I think it has in term of clarity and straight-forwardness...

xsawyerx:
I thought about this quite a bit. @mickeyn and I came up with one single keyword for all of it: `parameters`. This fits with `Plack::Request`:

    my $body_parameters = parameters('body'); # returns Hash::MultiValue
    my $route_parameters = parameters('route'); # returns Hash::MultiValue
    my $query_parameters = parameters('query'); # returns Hash::MultiValue
    my $all_parameters = parameters; # returns Hash::MultiValue

This allows me to actually use `Plack::Request`'s parsing with very little overrides. The only annoying part is reimplementing the body parsing (which included copying some code) to allow a situation of a serializer present.

And here's my problem: What do we do about serialized parameters? There are two questions remaining before I can provide this PR:

1. Given `{"foo":["bar","baz"]}`, should `parameters->get('foo')` return "bar" (which would happen in the case of `?foo=bar&foo=baz`) or should it return `[ "bar", "baz" ]`?
2. Given non-reference content (`foo`, which can happen with serializations like CBOR), what do we do? `Hash::MultiValue` won't work. Do we crash? Do we ignore it? Do we just deserialize it? What should `parameters()` return?

[...]

PR #937, opened 2015-06-15
Allow non-hashref parameters

xsawyerx:
Until now parameters had to be a key=value, the way HTTP requires.
However, when using serializers, we might find ourselves receiving
serialized input that is not a hash reference. CBOR is an example
of such a serializer.

Up until now we would crash somewhere internally when trying to
deserialize it. This commit changes it to silently ignoring it,
providing empty parameter values (since there is no "key" to
associate it with), but still making the deserialized data
available in "request->data".

This issue was discussed at GH #740 as part of work to move
Dancer2::Core::Request to be based more on Plack::Request and
adding Hash::MultiValue parameters.

This test uses CBOR::XS and Dancer2::Serializer::CBOR but does
not require them.

This fixes GH #844.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -477,8 +477,11 @@ sub _build_params {
 
     # and merge everything
     $self->{_params} = {
-        %$previous,                %{ $self->_query_params || {} },
-        %{ $self->_route_params }, %{ $self->_body_params  || {} },
+        map +( ref $_ eq 'HASH' ? %{$_} : () ),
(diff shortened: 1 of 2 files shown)


veryrusty:
Should we redo the test to use Dancer2::Serializer::YAML? Its part of the core and we have YAML.pm as a test dependency and it handles strings and arrayrefs.

Otherwise :+1: ! :dancers:

xsawyerx:
Added a test with `JSON`. I kept the `CBOR::XS`, optional, of course.

PR #938, opened 2015-06-15
Corrected link from Dancer2 Manual POD to params keyword.

ludovic-tc:
This is just to fix a link in the POD of Dancer2/lib/Dancer2/Manual.pod.


veryrusty:
Thanks @ludovic-tc!

Merged in a41197f0.

xsawyerx:
:+1:

Issue #936, opened 2015-06-15
error_template not rendered for implicit 404


mattchance:
When there are no matching routes or static files for a request and a 404 is thrown, the error_template set in the config is not used.

To reproduce:
https://gist.github.com/32544ef5d6972184050d

A request to /does-not-exist will result in error.tt being rendered, due to explicitly throwing a 404 error.
A request to /doesnt-exist however will not result in error.tt being rendered, instead a built-in error page.

veryrusty:
Hmmm, `Dancer2::Core::Error` is attempting to render the `error_template` for that second URL, but fails with this error (and hence falls back to the default error page):
```
Can't call method "params" on an undefined value at lib/Dancer2/Core/Role/Template.pm line 180.
```

xsawyerx:
Fix merged. Thanks for reporting! :+1:

PR #939, opened 2015-06-15
Bugfix/app set request

veryrusty:
Fix for #936. When rendering the 404 response, particularly when called from Dancer2::Core::Dispatcher, we were setting the request in the app object, but not in the engines.

When an error template was rendered, there was no request object in the template engine, so when the template attempted to merge the non-existant request + params it would throw an error, resulting in the generic 404 response being generated.

Calling `app->set_response` now sets the response attribute in the app and all its engine instances.


xsawyerx:
:+1:

Will merge today.

xsawyerx:
"And within 23 hours, it was indeed merged!" :)

Issue #926, opened 2015-06-08
session_dir default value is not "appdir/session"?


sherwind:
Hi,

I'm using uwsgi to run multiple Dancer apps on a dev server. It works fine except that today I noticed apps that don't specify session_dir create and share "sessions" directory under /etc/uwsgi.d -- the cwd before each app is launched. I already fixed this issue in uwsgi but my understanding is that the default session_dir should be "appdir/sessions". Is this a bug or my understanding of appdir is just wrong? Thanks 

    $ perldoc Dancer2::Session::YAML | grep -B1 default
       Files will be stored to the value of the setting "session_dir", whose
       default value is "appdir/sessions".

    $ dancer2 -a MyWebApp

    $ echo 'session: "YAML"' >> MyWebApp/config.yml

    $ ls
    MyWebApp        

    $ plackup MyWebApp/bin/app.psgi &>/dev/null &

    $ lwp-request -s -d http://localhost:5000
    200 OK

    $ ls
    MyWebApp        sessions

    $ grep -A4 'has session_dir' $(perldoc -l Dancer2::Core::Role::SessionFactory::File)
    has session_dir => (
        is      => 'ro',
        isa     => Str,
        default => sub { path( '.', 'sessions' ) },
    );

hvoers:
I think this is a bug, the code is wrong.

path( '.', 'sessions' ) should be path( setting('appdir'), 'sessions' )

hvoers:
I made a PR #929

veryrusty:
Thanks for reporting this @sherwind. Looks like a bug to me, `appdir/sessions` as the default is going to be less surprising that `./sessions`

@hvoers that suggestion won't work. `setting` is a DSL keyword, which will not be available within `Dancer2::Core::Role::SessionFactory::File`.  Looks like there are no tests against that role either :(

hvoers:
On Thu, 11 Jun 2015, Russell Jenkins wrote:

> @hvoers that suggestion won't work. setting is a DSL keyword, which will not be available within Dancer2::Core::Role::SessionFactory::File.

Mmm. 'path' is a DSL keyword and it works just fine.

@Sherwin Can you test this patch?

sherwind:
[...]

PR #940, opened 2015-06-16
Feature/appdir sessions

veryrusty:
Make the default `session_dir => appdir/sessions`, as documented in all existing file-based session stores. Ref #926.

More importantly, adds tests for ::SessionFactory::File, which has somehow been removed.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -187,8 +187,10 @@ sub _build_session_engine {
 
     Scalar::Util::weaken( my $weak_self = $self );
 
+    # Note that engine options will replace the default session_dir (if provided).
     return $self->_factory->create(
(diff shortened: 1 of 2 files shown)


xsawyerx:
:+1:

xsawyerx:
What happens if it's not defined?

veryrusty:
As in `$app->config->{appdir}` not defined? I guess its possible.

Given we use `path( $self->location, 'views' )` and `path( $self->location, 'public' )` as defaults in Dancer2::Core::App for the paths to views and public folders, using `path( $self->location, 'sessions' )` would be more consistent.

xsawyerx:
Agreed!

Alright, let the merge fest begin! :)

xsawyerx:
Merged! :+1:

PR #946, opened 2015-07-11
Improve hook/route handling when content is already set in response

### LINKED ISSUE #944: Error after returning content from before hook
Here is test case, which reproduces bug:

    use strict;
    use warnings;

    use Test::More;
    use Plack::Test;
    use HTTP::Request::Common;

    {
        package App; ## no critic
        use Dancer2;
        set serializer => 'JSON';
        hook before => sub {
            response->content({ foo => 'bar' });
            response->halt;
        };

        get '/' => sub {1};
    }

    my $test = Plack::Test->create( App->to_app );

    my $res = $test->request( GET '/' );
    ok( $res->is_success, 'Successful request' );
    is( $res->content, '{"foo":"bar"}', 'Correct content' );

It fails with error "Failed to serialize the request: hash- or arrayref expected (not a simple scalar, use allow_nonref to allow this) at /usr/local/booking-perl/5.18.2/site/lib/JSON.pm line 154. in (eval 106) l. 1". This happens because serialiser applies two times. First when we call content(…) from the hook and second call is from Dancer2/Core/App.pm from method _add_content_to_response, which is called from _dispatch_route:

    if ( $response->is_halted ) {
        return $self->_add_content_to_response(
            $response, $response->content,
         );
    }

veryrusty:
There were two cases where to copied the response content into a new var, only to set it straight back again into the response object. Apart from being loopy and unnecessary, when a serializer is defined, this would attempt to serialize the previously serialized content!

This Pr modifies the existing (private) method for preparing response content after hook and route execution, only setting response content if that is required, leaving existing content in the response alone. Fixes #944.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1396,10 +1396,8 @@ sub _dispatch_route {
     return $response;
 }
 
-sub _add_content_to_response {
+sub _prep_response {
     my ( $self, $response, $content ) = @_;
-
-    defined $content or return $response;
 
     # The response object has no back references to the content or app
     # Update the default_content_type of the response if any value set in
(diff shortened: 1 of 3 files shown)


xsawyerx:
#954 merged, contains these commits. Nice work, @veryrusty! :+1:

Issue #947, opened 2015-07-13
Failure to escape paths in regex patterns causes spurious failures on Windows


nanis:
Can be fixed by using `\Q$file\E` on line 22 of the test script `t\logger_console.t`.

See also [Tests failing due to directory separators are unforced errors](http://www.nu42.com/2014/11/tests-failing-due-to-path-separators.html).

<pre>t\logger_console.t ..................................... Unrecognized escape \l passed through in regex; marked by <-- HERE in m/core in t\l <-- HERE ogger_console.t l. 15/ at t\logger_console.t line 22.

t\logger_console.t ..................................... 1/? #   Failed test 'core message sent'
#   at t\logger_console.t line 22.
#                   '[test:4652] core @2015-07-12 21:57:26> core in t\logger_console.t l. 15
# '
#     doesn't match '(?^:core in t\logger_console.t l. 15)'
Unrecognized escape \l passed through in regex; marked by <-- HERE in m/debug in t\l <-- HERE ogger_console.t l. 15/ at t\logger_console.t line 22.

#   Failed test 'debug message sent'
#   at t\logger_console.t line 22.
#                   '[test:4652] debug @2015-07-12 21:57:26> debug in t\logger_console.t l. 15
# '
#     doesn't match '(?^:debug in t\logger_console.t l. 15)'
Unrecognized escape \l passed through in regex; marked by <-- HERE in m/info in t\l <-- HERE ogger_console.t l. 15/ at t\logger_console.t line 22.

#   Failed test 'info message sent'
#   at t\logger_console.t line 22.
#                   '[test:4652] info @2015-07-12 21:57:26> info in t\logger_console.t l. 15
# '
#     doesn't match '(?^:info in t\logger_console.t l. 15)'
Unrecognized escape \l passed through in regex; marked by <-- HERE in m/warning in t\l <-- HERE ogger_console.t l. 15/ at t\logger_console.t line 22.

#   Failed test 'warning message sent'
#   at t\logger_console.t line 22.
#                   '[test:4652] warning @2015-07-12 21:57:26> warning in t\logger_console.t l. 15
# '
#     doesn't match '(?^:warning in t\logger_console.t l. 15)'
Unrecognized escape \l passed through in regex; marked by <-- HERE in m/error in t\l <-- HERE ogger_console.t l. 15/ at t\logger_console.t line 22.

#   Failed test 'error message sent'
#   at t\logger_console.t line 22.
[...]

PR #948, opened 2015-07-13
GH #947: Tests fail due to unescaped file paths in regex pattern

nanis:
On Windows, `\` is the directory separator. Therefore, when `t\logger_console.t` is interpolated into a regex pattern, it becomes `tlogger_console.t`,
causing the match, and consequently, the tests in that file to fail.

Even on other platforms, the regex contains two unescaped dots, which, strictly speaking, is not correct.

The solution is to use `\Q$file\E` when interpolating a variable `$file` in to a regex pattern, and use a literal dot in the line number expression.


veryrusty:
:+1:  @nanis++

Merged. Thanks!

nanis:
Thank you very much for your work.

PR #950, opened 2015-07-16
Correctly handle multi byte unicode chars in param keys. Test the same.

patzim:
I did change the _decode() function to also decode hash keys and not only values. This seems to be how it should be at least for GET parameters. The function is also used for decoding other values though. I am not sure this change does not break any of those other uses. I would be grateful if someone with some deeper knowledge of Request.pm could have a look and verify I didn't break anything.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -342,12 +342,7 @@ sub _decode {
     }
 
     if ( ref($h) eq 'HASH' ) {
-        while ( my ( $k, $v ) = each(%$h) ) {
-            my $decK = _decode($k);
-            $h->{$decK} = _decode($v);
-            delete $h->{$k} if $k ne $decK;
-        }
-        return $h;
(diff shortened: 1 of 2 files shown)


veryrusty:
I'm in favour of decoding both the keys and values of params. Thanks for the Pr @patzim.

This needs reworking before merging though. On perl 5.20+, this generates the warning `Use of each() on hash after insertion without resetting hash iterator results in undefined behavior`.  There is no need to use `each` if we want to decode both the keys and values, a `map` will suffice. 

@patzim would you like to take another look at this?

patzim:
As proposed now using map to decode keys and values. (Didn't know the trick of using map with hashes. Always good to learn new things. :-)

veryrusty:
Looks good. :+1:  Thanks @patzim. Tis awesome that you learnt something new too! :dancers: 

I'll squish and merge later in the week.

veryrusty:
Merged as e3fe87e (squashed commits).

Thanks again @patzim!

Issue #959, opened 2015-07-22
Custom error template rendering broken by use of var in 'before' hook


aka-raj:
Dancer2 0.161

1) 500.tt rendered OK:

    any '/foo' => sub {
      die ('died here');
      template 'foo';
    }

2) inclusion of a before_template_render hook causes 500.tt to be ignored:

    hook before_template_render => sub {
      my $tokens = shift;
      var some_var => 21; # <= this seems to be culprit ??
    };

* show_errors: 0 -  msg = "Wooops, something went wrong"
* show_errors: 1 - msg = "died here at /home/raj/apps/error_test/bin/../lib/error_test.pm line 16"

commenting out 'var' in before_template_render hook restores proper 500.tt rendering function

aka-raj:
just realised use of 'var' in before_template_render hook probably not supported - substituting $tokens->{some_var} instead works fine

veryrusty:
May need to break this into two issues
  * Dancer2::Core::Error->_build_content has two `eval` blocks that render templates. If an error is thrown by either, we never log the actual error. That makes debugging harder than it should be.
  * The error being thrown when the custom template being rendered is
```
Function 'var' must be called from a route handler
```

If someone wants to look into this before I get a chance, it looks like where ever Dancer2::Core::App generates a Dancer2::Core::Error object, we need to ensure that `$Dancer2::Core::Route::REQUEST` and `$Dancer2::Core::Route::RESPONSE` are localised to the current request/response.

xsawyerx:
Fix in #962 merged. Thanks for opening the ticket! :)

PR #962, opened 2015-07-23
Bugfix: request scoping when rendering error templates

veryrusty:
From #959: If a route `die`d and you have a `before_template_render` hook that used keywords that need to be used inside a route handler, then rendering custom error templates would also die, giving you the default error template. This was inconsistent with `send_error`, which did render the custom error template.

This was another scoping issue; when Dancer2::Core::App generates 500 or 404 errors, the localized REQUEST/RESPONSE were no longer set.

As a bonus, if the error templates fail to render, the errors are now logged, which will make future debugging _much_ easier.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1416,6 +1416,9 @@ sub response_internal_error {
 
     # warn "got error: $error";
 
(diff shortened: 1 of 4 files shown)


xsawyerx:
:+1: Excellent!

xsawyerx:
Merged!

PR #964, opened 2015-07-25
Exceptions in before hooks renders custom error templates

### LINKED ISSUE #961: Custom handling of errors in core.app.before_request
Custom error template doesn't handle "Exception caught in 'core.app.before_request' filter: Hook error: [...]"

None of the init_error, before_error, after_error & on_route_exception hooks get called beforehand, but the 'before' hook does, but then request terminates with above message. 

Can core.app.before_request errors be handled in same way as route errors?

Full text message:
>Internal Server Error

>Exception caught in 'core.app.before_request' filter: Hook error: DBD::mysql::st execute failed: 
>[...] at /home/raj/apps/RequestForm/bin/../lib/RequestForm/DB.pm line 147.
> at /home/raj/perl5/lib/perl5/Dancer2/Core/App.pm line 962.
> at /home/raj/perl5/lib/perl5/Dancer2/Core/Role/Hookable.pm line 114.

veryrusty:
Catch exceptions in `before` hooks, allowing any defined custom error templates (or `views/500.tt`) to be rendered (rather than the generic internal server error page). Resolves #961.

This also triggered an issue in Dancer2::Core::Error where we checked if the error had an app, and then called app->request->uri_base;  errors can have an app without a request object.. fixed that too. :)

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1375,7 +1375,9 @@ sub build_request {
 sub _dispatch_route {
     my ( $self, $route ) = @_;
 
-    $self->execute_hook( 'core.app.before_request', $self );
+    local $@;
+    eval { $self->execute_hook( 'core.app.before_request', $self ); 1; }
+        or return $self->response_internal_error($@);
     my $response = $self->response;
 
     if ( $response->is_halted ) {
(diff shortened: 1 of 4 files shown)


xsawyerx:
Good job! :+1:

PR #967, opened 2015-07-29
fix docs

ambs:
the previous documentation used '$filename' as argument to upload, but it should be the field name.

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -1287,10 +1287,10 @@ Files are uploaded in Dancer2 using the class L<Dancer2::Core::Request::Upload>.
 The objects are accessible within the route handlers using the C<request>
 keyword and the C<uploads> attribute:
 
-    post '/upload/:file' => sub {
+    post '/upload' => sub {
         my $upload_dir   = "MyApp/UPLOADS";
-        my $filename     = params->{file};
-        my $uploadedFile = request->upload($filename);
+        my $uploadRequest = request->upload('file');
+        $uploadRequest->copy_to("$upload_dir/$filename");
     };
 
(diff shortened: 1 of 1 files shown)


ambs:
Can I merge this one? It is docs, so, no big deal, but would like a :+1:

veryrusty:
I did wonder if the param `file` and the form field `file` may cause confusion?
Possibly `name_of_upload_input` (or something less verbose) would be better.

However, I don't want to discourage correct docs, so merge away if you don't think its confusing.

ambs:
But aren't all form fields parameters?
That is, `file` is the same field in the two lines. But if you use the `param` method, you get the filename, but if you query the upload object, you get an object. Right?

But I can add some text to make it clear, if you like ;)

veryrusty:
There are two params named `file` in the example; one is a route param, the other is a body param (form field).

ambs:
And you say you are rusty :-P sorry, and thanks for pointing me to the route line, I didn't look to it with caution. probably I can change the upload route and remove the parameter, that is "useless" for the example. Anything against it?

veryrusty:
@ambs Feel free to totally rework the example :)

ambs:
@veryrusty, minor change. In vacations, so, trying to work the less possible O:-)
Would this be good enough for now? :-)

veryrusty:
@ambs that still not right; `$filename` isn't defined. Enjoy your vacation and I'll fix it ;)

veryrusty:
Merged as c848047. Thanks @ambs! :dancers:

ambs:
Gah,  stupid me. 
Thanks and sorry for the incompetence. 
Alberto 

-----Original Message-----

Issue #719, opened 2014-10-08
to_json and from_json are very expensive if called in a big loop


simonbrent:
Because they create a new Dancer2::Serializer::JSON object every time, calling to_json or from_json in a loop can add significant execution time.

I have an example with a loop iterating over about 9000 entries, each with a json string I need to turn into a hashref. In this case, calling Dancer2::from_json took about a second, while calling JSON::from_json took about 0.01 seconds.

Would you consider adding a faster method which didn't create the Serializer object every time?

Note: This was tested with Dancer 1, rather than Dancer2, but the code looks very similar

iftekhar25:
So, looking back at this patch, I wonder if I'm trying to be too clever by adding MooX::Singleton as a dependency, we can just as well make it a whole lot simpler. Feedback is welcome, happy to revise it.

xsawyerx:
@veryrusty's #972 was just merged, resolving this issue.

Thanks for all the discussion and work around it. :)

PR #969, opened 2015-07-31
lib/Dancer2/Serializer/JSON.pm: Issue # 719

iftekhar25:
PerlDancer/Dancer2#719

Performance improvement, cache the serializer object.

### DIFF
--- a/lib/Dancer2/Serializer/JSON.pm
+++ b/lib/Dancer2/Serializer/JSON.pm
@@ -8,14 +8,14 @@ with 'Dancer2::Core::Role::Serializer';
 
 has '+content_type' => ( default => sub {'application/json'} );
 
+my $s = Dancer2::Serializer::JSON->new;
+
 # helpers
 sub from_json {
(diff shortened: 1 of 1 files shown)


iftekhar25:
Some benchmarks:

https://gist.github.com/iftekhar25/7642c9722eeddaabda16

veryrusty:
Thanks @iftekhar25 for looking at this!

If speed is the main issue, using JSON::MaybeXS directly should out-perform the provided serialization helpers, however using XS modules is not always possible (and the core needs to be fatpackable).

The other serialization helpers for yaml and dumper also generate a new serialization object every time they are called. All three need the same issue to be addressed.

Your Pr is very much appreciated, and prompted me to think about it further. See #972 for my thoughts on how this could possibly be addressed. Thanks again!

xsawyerx:
PR #972 which addresses this was merged. Thank you for the push into solving it and the suggested solution. :+1:

PR #971, opened 2015-08-01
Fix route options matching

perlpunk:
some options weren't always validated because of
returning from the loop without resetting each iterator.
changed to for loop


veryrusty:
Another use of `each` that comes back to bite the unwary. @perlpunk++ 

Looks good to me, will merge shortly!  Thanks @perlpunk !!

veryrusty:
Merged as d572b52

Thanks again @perlpunk :dancers:

PR #972, opened 2015-08-05
Streamline the (to|from)_(json|yaml|dumper) serialization helpers.

### LINKED ISSUE #719: to_json and from_json are very expensive if called in a big loop

veryrusty:
The (to|from)_(json|yaml|dumper) helpers all create a new serializer
object and call (de)serialize on it. If you trace the actual calls, this
is roughly (pseudo code for serialize):

```
  Serializer->new->serialize(@_);
  # around modifier
  $self->execute_hook;
  original $self->serialize(...)
  $self->config;  (JSON only)
  do the serialization
  $self->log_cb if serialization failed.
```

However, as we have a new serializer object, there are no hooks, there
is no config, and the default lob_cb is `sub{1}`. ie. we make a lot of
method calls that do nothing after the expense of generating the
serializer object itself.

Instead, add logical checks that ensures that those methods only
get called if $self is blessed. This does not interfere with serializing
content at the end of a route, and allows (de)serialize to be called as
a class method. Resolves #719 and is an alternate to #969.

Another benefit is that serialization fails when called as a class
method will no longer be trapped and silently ignored, possibly resolving #644
and alternative to #647.

### DIFF
--- a/lib/Dancer2/Core/Role/Serializer.pm
+++ b/lib/Dancer2/Core/Role/Serializer.pm
@@ -4,6 +4,7 @@ package Dancer2::Core::Role::Serializer;
 use Moo::Role;
 use Try::Tiny;
 use Dancer2::Core::Types;
+use Scalar::Util 'blessed';
 
 with 'Dancer2::Core::Role::Engine';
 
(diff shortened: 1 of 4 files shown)


xsawyerx:
@veryrusty++

The only limitation we have here is that `__PACKAGE__` would always go for the class and won't allow inheritance. We can switch it to check whether a parameter is blessed or not but that doesn't necessarily mean it was called as a method, since you can serialized blessed variables too. Darn.

Anyway, it seems solid to me. @veryrusty, what do you think? Merging?

veryrusty:
The only place I used `__PACKAGE__` was in the methods that are called from the DSL serializaiton keywords. If you were to use a subclass of one of the core serializers, the methods the DSL calls shouldn't change. I'm not sure why one would subclass one of the core serializers; implementing a new serializer is straight forward. 

I'm for merging ;)

xsawyerx:
Merged! :+1:

PR #976, opened 2015-08-13
Short circuit $dsl->dsl on DSL import into plugins

ironcamel:
Plugins that use other plugins were trigging the DEPRECATED warning,
which was advising to use $dsl->dsl.

When importing the symbol 'dsl', compiled coderef returns the $dsl
object and does not raise the DEPRECATED warning.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -218,14 +218,17 @@ sub import {
                 $code->( $dsl, @_ );
             };
 
+            if ( $symbol eq 'dsl' ) {
+                $compiled = sub { $dsl };
+                $dsl_deprecation_wrapper = $compiled
+            }
+
             # Bind the newly compiled symbol to the caller's namespace.
             # As this may redefine a symbol, ensure the new coderef has
             # the same prototype signature.
             my $existing = *{"${plugin}::${symbol}"};
             my $prototype = prototype \&$existing;
             *{"${plugin}::${symbol}"} = Scalar::Util::set_prototype( \&$compiled, $prototype );
-
-            $dsl_deprecation_wrapper = $compiled if $symbol eq 'dsl';
         }
     }
 


ambs:
Would love to give feedback, but I cant understand the patch, so :-)

ironcamel:
@ambs this was a commit I cherry-picked (with permission) from @veryrusty's branch. Maybe he can give some more insight.

veryrusty:
First and foremost, this is a hack - its not worth considering if the reworking of the Plugin system by @yanick is close to being merged. Here's what's going on, and what this Pr is about:

`Dancer2::Plugin::_get_dsl` walks the call stack looking for something that `can('dsl')` that isn't the `$dsl_deprecation_wrapper`. When there is only one plugin in use, that's fine. When there are several, such at the example by @nunorc in #1008, each plugin adds a `dsl` method on import to the class.

In that example, any code that calls `_get_dsl`, such as the `plugin_setting` method skips the first caller that `can('dsl')`, but the next one (which has a different `$dsl_deprecation_wrapper`) triggers the deprecation warning to be thrown, even though there *IS* no call to `dsl`.  This is also responsible for the comment in #1008 regarding the order that the plugins are used.

This "hack" does two things
  * do not to add the deprecation warning on the `dsl` method that is imported, and
  * optimise the logic in the wrapper code for `dsl` keyword; there is no need to call `dsl->($dsl, @_)` as we can simply return the dsl object we already have.

ambs:
@yanick, what is the status of the module system?

I confess I disagree a little with @veryrusty. Applying a hack is worth if the module system is to take more than one week. I had a prior situation like this, when I suggested an hack to fix something, everybody said "no, we need to rewrite it" and... it took months before getting done.

So, please, @yanick, let us know if we can wait for it in a short term, or if applying and releasing the hack is acceptable to keep things working.

Cheers

veryrusty:
I doubt we'll get the plugins stuff cleaned up and merged before the dancer conference (its listed as an item for the hackathon).

@ambs I concur, better to merge rather than wait!

ambs:
OK, working on it.

ambs:
Erm, someone did it already?

ambs:
Merged, waiting from @nunorc or @ironcamel if this fixes the issues.

veryrusty:
Thanks @ambs !

( It certainly fixes warning in the example given by @nunorc )

ambs:
Thank *you*. Closing tickets, then :)

hvoers:
On Sat, 26 Sep 2015, Russell Jenkins wrote:

> I doubt we'll get the plugins stuff cleaned up and merged before the dancer conference

Me too.

Have you seen my posts on the users list?

http://lists.preshweb.co.uk/pipermail/dancer-users/2015-September/005017.html
http://lists.preshweb.co.uk/pipermail/dancer-users/2015-September/005018.html

In a real DSL you can just say 'whatever'.

'$dsl->whatever' is not a DSL.

Just my 2 cents,
Henk

PR #984, opened 2015-08-28
Use $^V for version in template token, not $]:

xsawyerx:
From perlvar:

       $OLD_PERL_VERSION
       $]      See "$^V" for a more modern representation of the Perl version
               that allows accurate string comparisons.

Before:

    5.018002

After:

    v5.18.2


ambs:
:+1:

PR #986, opened 2015-09-03
Replace File::Slurp (broken) with File::Slurper in Tutorial.pod

1nickt:
File::Slurp is broken, especially in its handling of encoding (see http://blogs.perl.org/users/leon_timmermans/2015/08/fileslurp-is-broken-and-wrong.html, https://rt.cpan.org/Public/Bug/Display.html?id=83126, https://rt.cpan.org/Public/Dist/Display.html?Status=Active;Name=File-Slurp). We shouldn't be recommending it in tutorials.


shumphrey:
:+1:

ambs:
Thanks!

PR #987, opened 2015-09-04
Captures should always return empty hashref:

xsawyerx:
captures() keyword would return undef if there were no captures.
That's bad. It should be consistent and always return hashref.


veryrusty:
:+1:

shumphrey:
:+1:

racke:
:+1:

Issue #739, opened 2014-10-30
params keyword should return Hash::MultiValue


xsawyerx:
We raised this problem in #726, which contains the entire discussion around it.

xsawyerx:
Fixed by #988!

PR #988, opened 2015-09-04
Provide keywords for Hash::MultiValue parameters

xsawyerx:
Finally it's landed.

This PR provides three new keywords:

    query_parameters
    body_parameters
    route_parameters

It returns a `Hash::MultiValue` object which you can use in order to get values without worrying about context.

    GET /?foo=bar&foo=baz

    get '/' => sub {
        # BEFORE:
        my $foo = params('query')->{'foo');
        ref $foo eq 'ARRAY' and $foo = $foo->[ $#foo ]; # we have to normalize $foo - yuck!

        # AFTER:
        my $foo = query_parameters->get('foo'); # always a single value
        my @foo = query_parameters->get_all('foo'); # always array, even if only one value
    };

<3

The reason we introduce these keywords instead of a `parameters` keyword is that we want users to make sure they know where they get the parameters and don't accidentally have overwritten merged parameters from multiple sources (and have to know how they are merged).

This handles the following tickets: #739, #740.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1123,7 +1123,7 @@ sub make_forward_to {
 
     $env->{PATH_INFO} = $url;
 
-    my $new_request = Dancer2::Core::Request->new( env => $env, body_is_parsed => 1 );
+    my $new_request = Dancer2::Core::Request->new( env => $env, body_params => {} );
     my $new_params = _merge_params( scalar( $request->params ), $params || {} );
 
     exists $options->{method} and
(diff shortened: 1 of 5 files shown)


xsawyerx:
@shumphrey @mickeyn @yanick @veryrusty @racke @ambs Would love to get some feedback on this. It's one of the next big things I want to push.

ambs:
Gah, this is a huge list of commits. Will take some time to review it :-/

xsawyerx:
Each commit is self-contained and only changes a little bit at a time.

mickeyn:
looks good to me and I really need this for my current project :)

@xsawyerx++

PR #989, opened 2015-09-04
Add option for providing skeleton

nunorc:
This option allows having custom skeleton directories for new applications. Although it's a very simple solution it can save a lot of work for anyone that creates a lot of applications and has to make the same changes every time (eg. use template toolkit instead of simple templates, add bootstrap).


xsawyerx:
:+1: from me.

One more step in the direction of profiles.

Would it be possible to add more documentation to this feature?

ambs:
:+1:

nunorc:
@xsawyerx thank you for your comment.

Just added a new commit that includes the new option in the "BOOTSTRAPPING A NEW APP" section of the manual. Let me know if it should be referenced anywhere else in the docs. Thank you.

ambs:
Seems good. Thanks! :+1:

PR #993, opened 2015-09-06
Route parameters should overwrite all other merged variables

xsawyerx:
When merging variables, body parameters overwrite query paraemeters,
but they would both overwrite the route parameters.

This commit makes sure (and adds a test to make sure) that route
parameters will not be overwritten.

This addresses GH #931, except it doesn't allow the query
parameters to overwrite the body.

Seems like Express.js does Query > Body, and Rails does Body > Query
and it's probably - if anything - not RFC compliant to begin with.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -387,8 +387,8 @@ sub _build_params {
         map +( ref $_ eq 'HASH' ? %{$_} : () ),
         $previous,
         $get_params,
-        $self->_route_params,
         $self->_body_params,
+        $self->_route_params,
     };
 
 }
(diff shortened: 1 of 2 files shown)


xsawyerx:
Should note that @shumphrey has given a :+1: to this solution on #931. Any comments from anyone before I merge it?

shumphrey:
I had a quick discussion with $work, we still agree that route > body > get. I think the scenarios where you'd get conflicting values would be rare, but in those cases, we agree with this PR

:+1:

veryrusty:
AFAIK there is no RFC to deal with this; merging the various sources of params was for convenience. If anyone _is_ having issues with clashing params, the approach in #988 provides a neater solution (in my opinion!).

This may be a breaking change for some people and is a change from D1. There should be some docs included too [ reworking #931 + additions to the migration pod should suffice :) ].

PR #996, opened 2015-09-07
Fix warning "Odd number of elements in anonymous hash" (loses params)

basbloemsaat:
If a route with an optional argument was called without the argument, eg.
```
get "/app/?:arg?"
```
was called as
```
GET /app
```
a warning "Odd number of elements in anonymous hash at .../Dancer2/Core/Request.pm line 345." was produced. This was caused by _decode on undef returned a null and the result is parsed in list context. This trivial change prevents this from occurring.

This also causes the params hash to be mangled. Eg.
```
get "/app/:a/:b/?:b?"
```
Called as
```
GET /app/1/2
```
has a high chance of losing params a and b. So any API implemented using optional params is probably affected.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -342,7 +342,7 @@ sub _decode {
     }
 
     if ( ref($h) eq 'HASH' ) {
-        return { map _decode($_), %$h };
+        return { map _decode($_) // undef, %$h };
     }
 
     if ( ref($h) eq 'ARRAY' ) {


veryrusty:
Thanks for the Pr @basbloemsaat !

We aim to maintain perl 5.8 compatibility; any chance you can rework to not use the defined-or operator?

basbloemsaat:
Will do

basbloemsaat:
Travis failing seems to be upstream btw, tests related to this are ok

xsawyerx:
Merged in https://github.com/PerlDancer/Dancer2/commit/64e0befe3eca8513ef15fb0f2f7572c5eb8886ea, thanks! :+1:

Issue #941, opened 2015-06-22
Callbacks for streaming asynchronous responses


xsawyerx:
While we have the `content` keyword for streaming content to the user as part of an asynchronous response, we have no way of knowing whether it worked or not.

The writer does not expose whether it was able to write or not. This means that if a user closed a connection before we called `done`, you will probably get a caught exception in your log that you can't do anything about.

While there are discussions on expanding the spec of PSGI to accommodate for that, maybe we should already think of syntax that - for now - will just eval the write operation instead. I've checked and this approach works, but I'm not sure about the DSL for it.

I'm mainly interested in hearing what everyone thinks a good syntax would be:

The simple case would be to have two callbacks for success and fail for the `content` keyword:

    content $more_data, \&success, \&fail;
  
Another would be to return a promise:

    my $p = content $mode_data, \&success, \&fail;
    $p->then(...);

or

    content( $more_data, \&success, \&fail )->then(...);

But that would require a promises module (I like `Promises` personally) which I don't know if we want to add to core. I'm thinking a plugin might help with this, or perhaps a module that upgrades regular callbacks into promises (and then maybe wrap that in a plugin or specialized DSL) - the **q** library has that.

Another option would be named callbacks:

    content $new_data, on_success => \&success, on_fail => \&on_fail;

Another would be a new keyword of some kind, but I'm not sure that will actually help.

Any other ideas I missed?

andrewsolomon:
I have no experience with Promises but having now read [Promises::Cookbook::GentleIntro](https://metacpan.org/pod/Promises::Cookbook::GentleIntro) I think it'd be a great addition.

My preference is to have named callbacks. Partly because of its readability, but also because this is a new terrain it leaves more room for change.

xsawyerx:
I think the first question to answer is: Should we return promises or not. That ties us into either Futures or Promises. There is an implementation of **q** in Futures but not for Promises. If we had that, we could say "we don't return promises, but you can use these modules to create promises from it".

We would need some sample code for it.

We can make the `content` keyword faster by using positional parameters and prototypes instead of named parameters, a la `AE`.

xsawyerx:
Reconsidering this, providing a callback for the entire `delayed` makes more sense. When a write fails, the entire `delayed` should fail. You should be able to do an eval for each `content` call, but otherwise have one `on_error` callback for the entire response.

How about this syntax:

    get '/' => delayed {
        content "hello";
        done;
    } on_error => sub {
[...]

PR #999, opened 2015-09-12
Add error handling for delayed responses

xsawyerx:
GH #941 mentions callbacks for delayed responses. Currently the
biggest one is error handling. This is what this commit adds.

It does this by introducing another argument to the "delayed"
keyword, "on_error".

    get '/' => sub {
        delayed {

        } on_error => sub {

        };
    };

The "on_error" option receives a sub. Since the "delayed" request
is already done within the context of a delayed response, there's
no need to use "delayed" again in the sub, just providing a regular
callback instead.

The tests don't fully work since the development server, for some
still unbeknownst reason, doesn't actually crash after we close
the socket and write again. In Twiggy it does. You can use this
test script which uses Twiggy:

    use strict;
    use warnings;
    my $caught_error;

    {
        package App::ErrorHandler; ## no critic
        use Dancer2;
        use AnyEvent;
        get '/log' => sub {
            delayed {
                flush;
                content "ping\n";
                done;
                content "failure\n";
            };
        };

        get '/cb' => sub {
            delayed {
                flush;
                content "ping\n";
                done;
                content "failure\n";
            } on_error => sub {
                my $caught_error = shift;
                core "Failed to stream to user: " . request->remote_address;
                1;
            };
        };
    }

    App::ErrorHandler->to_app;

Original tests marked as TODO until they work.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -889,7 +889,11 @@ sub send_file {
                 [ $res->status, $res->headers_to_array, $fh ]
             );
         };
+
+        Scalar::Util::weaken( my $weak_self = $self );
+
         $response = Dancer2::Core::Response::Delayed->new(
+            error_cb => sub { $weak_self->logger_engine->log(@_) },
             cb       => $cb,
             request  => $Dancer2::Core::Route::REQUEST,
             response => $Dancer2::Core::Route::RESPONSE,
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -893,7 +893,7 @@ sub send_file {
         Scalar::Util::weaken( my $weak_self = $self );
 
         $response = Dancer2::Core::Response::Delayed->new(
-            error_cb => sub { $weak_self->logger_engine->log(@_) },
+            error_cb => sub { $weak_self->logger_engine->log( warning => @_ ) },
             cb       => $cb,
             request  => $Dancer2::Core::Route::REQUEST,
             response => $Dancer2::Core::Route::RESPONSE,
--- a/lib/Dancer2/Core/DSL.pm
+++ b/lib/Dancer2/Core/DSL.pm
@@ -55,7 +55,7 @@ sub dsl_keywords {
         debug                => { is_global => 1 },
         del                  => { is_global => 1 },
         delayed              => {
-            is_global => 0, prototype => '&',
+            is_global => 0, prototype => '&@',
         },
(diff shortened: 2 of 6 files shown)


lib/Dancer2/Core/App.pm
```diff
@@ -889,7 +889,11 @@ sub send_file {
                 [ $res->status, $res->headers_to_array, $fh ]
             );
         };
+
+        Scalar::Util::weaken( my $weak_self = $self );
+
         $response = Dancer2::Core::Response::Delayed->new(
+            error_cb => sub { $weak_self->logger_engine->log(@_) },
```
veryrusty: The call to `logger_engine->log(@_)` is missing an argument. It needs both a *log level* and the log message.

veryrusty:
This looks good. @xsawyerx++
I'll take a final pass over it tomorrow. :)

lib/Dancer2/Core/App.pm
xsawyerx: Can you add a test plus a fix? Pretty please? :)

xsawyerx:
@veryrusty Thank you. :)

lib/Dancer2/Core/App.pm
xsawyerx: Pinging again on this, @veryrusty, I think I tested it when committing it and it worked. Basically when you call log, it just passes it on to the logger. Are you sure this is a mistake?

veryrusty: Sorry to go MIA on this. I'm sure its missing an arg.  
  * `error_cb` takes the error message as its only argument. 
  * The `log` method on all the logging engines take two args: "msg_level", and  "msg_content".

The error callback added to the `content` keyword in this PR  [two args for its logging call](https://github.com/PerlDancer/Dancer2/blob/feature/delayed-on-error/lib/Dancer2/Core/DSL.pm#L301-L303)

This is inside of `send_file`, which uses delayed responses to send a filehandle back to the PSGI engine. So `$Dancer2::Core::Route::WRITER->write` shouldn't be called in this use case; I'm not certain its possible to write a test case for it (at least I can't think of how at the moment).
 [ The fix would be: `error_cb => sub { $weak_self->logger_engine->log(warning => @_) }` ]

xsawyerx: I'll amend it to contain the fix. Thanks! Unfortunately the development server's failure is kind of weird. I couldn't write a proper test for it.

xsawyerx:
@veryrusty Added the fix! Thanks!

I'll be merging this soon.

PR #1007, opened 2015-09-20
Fix/no stack trace on client error

### LINKED ISSUE #1001: 404 presents runtime error
With latest Dancer version, when an unknown route is called, the 404 page shows a runtime error, with backtrace et al. Not sure this is desired.

To test, simple bootstrap a new app, plack it up, and try to access some strange route.

Cheers

veryrusty:
Only show the "full" error message (stack trace, env, etc.) if `show_errors` is true AND the error status is `5xx`. Resolves #1001.


xsawyerx:
Merged! :+1:

PR #1010, opened 2015-09-27
Dancer2::Plugin2

yanick:
yanick's prototype.


racke:
Got the following failure while running "dzil install" on your branch:

```
Can't locate Test/WWW/Mechanize/PSGI.pm in @INC (you may need to install the Test::WWW::Mechanize::PSGI module
) (@INC contains: t/lib /home/racke/dancer2/.build/2Yr7tNviuK/blib/lib /home/racke/dancer2/.build/2Yr7tNviuK/b
lib/arch /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/x86_64-linux /home/racke/perl5/perl
brew/perls/perl-5.22.0/lib/site_perl/5.22.0 /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/5.22.0/x86_64-lin
ux /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/5.22.0 .) at t/plugins2-1.t line 9.
BEGIN failed--compilation aborted at t/plugins2-1.t line 9.
# Looks like your test exited with 2 before it could output anything.
t/plugins2-1.t ......................................... 
Dubious, test returned 2 (wstat 512, 0x200)
Failed 6/6 subtests 
Can't locate Test/WWW/Mechanize/PSGI.pm in @INC (you may need to install the Test::WWW::Mechanize::PSGI module
) (@INC contains: t/lib /home/racke/dancer2/.build/2Yr7tNviuK/blib/lib /home/racke/dancer2/.build/2Yr7tNviuK/b
lib/arch /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/x86_64-linux /home/racke/perl5/perl
brew/perls/perl-5.22.0/lib/site_perl/5.22.0 /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/5.22.0/x86_64-lin
ux /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/5.22.0 .) at t/plugins2-2.t line 9.
BEGIN failed--compilation aborted at t/plugins2-2.t line 9.
# Looks like your test exited with 2 before it could output anything.
t/plugins2-2.t ......................................... 
```

racke:
Hm, is that just Dist::Zilla annoying me again?

```
Can't locate MooX/Types/MooseLike/Numeric.pm in @INC (you may need to install the MooX::Types::MooseLike::Numeric module) (@INC contains: /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/arch /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/x86_64-linux /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0 /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/5.22.0/x86_64-linux /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/5.22.0 .) at /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib/Dancer/Core/Types.pm line 13.
BEGIN failed--compilation aborted at /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib/Dancer/Core/Types.pm line 13.
Compilation failed in require at /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib/Dancer/Core/Runner.pm line 8.
BEGIN failed--compilation aborted at /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib/Dancer/Core/Runner.pm line 8.
Compilation failed in require at /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib/Dancer.pm line 13.
BEGIN failed--compilation aborted at /home/racke/dancer2/.build/nx9HggrEUg/Dancer-2.0/../blib/lib/Dancer.pm line 13.
Compilation failed in require at t/uri_for.t line 6.
BEGIN failed--compilation aborted at t/uri_for.t line 6.
t/uri_for.t .................... 
```

veryrusty:
@racke AutoPrereqs was disabled in #966. This work predates that change..

racke:
OK, finally it built. @yanick, it would be nice to say explicitly that `plugin_setting` is replaced by Moo attributes.

racke:
How do we do this with Plugin2?

```
register require_role  => \&require_role;
register requires_role => \&require_role;
```

yanick:
[This work predates that change..] Yup. This patch dates from March.

yanick:
@racke [register]

Do you mean how we define keywords?  That'd be described in the POD, at line 103 and following:

```
+ =head3 Adding keywords
+
+Keywords that the plugin wishes to export to the Dancer2 app must be defined in a 
+C<keywords> function:
+
+    sub keywords { qw/ add_smileys add_sad_kitten  / }
+
+Then each of the keyword is written as a function which first argument will
+be the plugin object itself.
+
+    sub add_smileys {
+        my( $plugin, $text ) = @_;
+
+        return join ' ', $text, $plugin->smiley;
+    }
+
+    # and then in the app
+
+    get '/' => sub {
+        add_smileys( "Hi there!" );
+    };
```

racke:
No my question was how do define two keywords using the same subroutine.

yanick:
[plugin settings] The documentation was mentioning it:

```
=head3 Inheriting from L<Dancer2::Plugin2>

The plugin must inherit from L<Dancer2::Plugin2>. The base clase provides the plugin with 
two attributes: C<app>, which is populated with the Dancer2 app object for which
the plugin is being initialized for, and C<config> which holds the plugin 
section of the application configuration.

```

But I'll add a about the base methods, so it's more apparent.

yanick:
@racke 

```
sub keywords { qw/ require_role require_roles  / }

sub require_role { ... }

*require_roles = *require_role;
# or
sub require_roles { &require_role }

```

racke:
Please fix this typo as well: The base clase

yanick:
For keywords that are synonyms, we could also have the convention:

```
sub keywords { 
  'align_left', [ 'align_center', 'align_centre' ], 'align_right'
}

sub align_center { ... }
```

I.e, an arrayref means synonyms, use the first one has the function to call

yanick:
Also, `sub keywords` could -- nay, should -- be wrapped in an utility function exported by D2::P2, such that we can do

```
plugin_keywords qw/ foo bar  baz /;
```

I'll have a patch shortly for that.

racke:
Tests are not exactly exhaustive :-/.

yanick:
@racke This was intended as a prototype to run against the team. I was not going to write gigabytes of tests before even knowing if you guys want this. If it gets a green flag, it'll be fleshed out.

racke:
@yanick - Fair enough, we need more experiments with this stuff. I also suggest to have a method for loading plugins instead of hiding that in an Moo attribute.

lib/Dancer2/Core/App.pm
```diff
@@ -34,6 +34,44 @@ with qw<
 
 sub supported_engines { [ qw<logger serializer session template> ] }
 
+sub with_plugins {
```
yanick: [ I also suggest to have a method for loading plugins instead of hiding that in an Moo attribute.]

Something like this?

racke: Yes, sounds good.

racke:
Next question: if the plugin adds a route and this route needs access to plugin config - how does that work?

racke:
Ok, what is the equivalent to the following code:

```perl
register_hook qw(login_required permission_denied);
```

That is important for porting Dancer2::Plugin::Auth::Extensible for example.

yanick:
@racke 

[if the plugin adds a route and this route needs access to plugin config - how does that work?]

Most easily:

```
    sub BUILD {
        my $plugin = shift;

        $plugin->app->add_route(
            method => 'get',
            regexp => '/do_stuff',
            code   => sub {  
                "The plugin says " . $plugin->config->{greeting};
            },
        );

    }
```

racke:
I quite like the Plugin2 approach, but without a `register_hook` equivalent I'm pretty stuck.

yanick:
[without a register_hook equivalent I'm pretty stuck.]

Working on it.

racke:
Any idea where this might come from, @yanick ?

```
t/01-basic.t ...... Variable "$plugin" is not available at (eval 135) line 1.
Variable "$plugin" is not available at (eval 137) line 1.
Variable "$plugin" is not available at (eval 139) line 1.
Variable "$plugin" is not available at (eval 141) line 1.
Variable "$plugin" is not available at (eval 143) line 1.
Variable "$plugin" is not available at (eval 145) line 1.
Variable "$plugin" is not available at (eval 147) line 1.
Variable "$plugin" is not available at (eval 149) line 1.
```

racke:
Also, the Plugin2 tests in Dancer2 are failing in a similar fashion:

```
t/plugins2-1.t ......................................... Variable "$plugin" is not available at (eval 95) line 1.
String found where operator expected at t/lib/poc.pm line 30, near "truncate_txt "hello there""
	(Do you need to predeclare truncate_txt?)
syntax error at t/lib/poc.pm line 30, near "truncate_txt "hello there""
Compilation failed in require at t/plugins2-1.t line 6.
BEGIN failed--compilation aborted at t/plugins2-1.t line 6.
t/plugins2-1.t ......................................... Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 
t/plugins2-2.t ......................................... Variable "$plugin" is not available at (eval 95) line 1.
[poc2:3146] error @2015-10-01 16:00:22> Route exception: Undefined subroutine &poc2::add_smileys called at t/lib/poc2.pm line 19. in /home/racke/dancer2/.build/hj5otcx3t2/blib/lib/Dancer2/Core/App.pm l. 1317
t/plugins2-2.t ......................................... 1/6 
#   Failed test 'GET /'
#   at t/plugins2-2.t line 15.
# 500
# Internal Server Error

#   Failed test 'Content is like "(?^:please)"'
#   at t/plugins2-2.t line 16.
#          got: "<!DOCTYPE html>\x{0a}<html lang="en">\x{0a}<head>\x{0a}  <meta ch"...
#       length: 491
#     doesn't match '(?^:please)'

#   Failed test 'Content is like "(?^:8-D)"'
#   at t/plugins2-2.t line 17.
#          got: "<!DOCTYPE html>\x{0a}<html lang="en">\x{0a}<head>\x{0a}  <meta ch"...
#       length: 491
#     doesn't match '(?^:8-D)'
# Looks like you failed 3 tests of 6.
t/plugins2-2.t ......................................... Dubious, test returned 3 (wstat 768, 0x300)
Failed 3/6 subtests 
t/plugins2-hooks.t ..................................... ok   
```

yanick:
Yup, there was  an eval we don't need anymore. Removed it, plugins2 tests are passing again.

racke:
So here is a proof of concept for a plugin using Plugin2:

https://github.com/PerlDancer/Dancer2-Plugin-Auth-Extensible/tree/topic/plugin2

It at least passes all the tests, but I'm sure it breaks untested features :-).
In general plugins written with Plugin2 will be cleaner and more object oriented,
as we got a real `$plugin` object.

How do we deal with backward compability? The syntax for "using" the plugin has
changed, so applications will break if a plugin switched to Plugin2.

Any ideas?

racke:
And apparently any plugin that you use inside your plugin needs to be based on Plugin2 as well:

```
Can't locate object method "new" via package "Dancer2::Plugin::DBIC" at /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/Dancer2/Core/App.pm line 64.
```

yanick:
On 2015-10-02 07:22 AM, Stefan Hornburg (Racke) wrote:
> So here is a proof of concept for a plugin using Plugin2:
> 
> https://github.com/PerlDancer/Dancer2-Plugin-Auth-Extensible/tree/topic/plugin2

Nice.

And it makes me think I can add a little bit of sugar and add a
`from_config` attribute argument to automatically grab the value from
the configuration, as we often do that.

> It at least passes all the tests, but I'm sure it breaks untested
> features :-).

Aaah, but just like the tree falling in the forest with no-one watching,
is it really a bug if it break something untested? ;-)

> How do we deal with backward compability? 

Notice that all of the old plugin tests still pass. The old plugins can
still live side-by-side with p2 plugins. They'll just ignore each other.

> The syntax for "using" the
> plugin has
> changed, so applications will break if a plugin switched to Plugin2.

I can fix that. Wait until after breakfast, and the patch will be there. :-)

Joy,
`/anick

yanick:
On 2015-10-02 07:37 AM, Stefan Hornburg (Racke) wrote:
> And apparently any plugin that you use inside your plugin needs to be
> based on Plugin2 as well:
> 
> |Can't locate object method "new" via package "Dancer2::Plugin::DBIC" at /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/Dancer2/Core/App.pm line 64.
> |

Yeeeeeah.... That's were my magic powers end. P1 plugins just don't play
well with others, period. My skills are not good enough to fix that. :-/

`/.

racke:
On 10/02/2015 03:01 PM, Yanick Champoux wrote:
> On 2015-10-02 07:37 AM, Stefan Hornburg (Racke) wrote:
>> And apparently any plugin that you use inside your plugin needs to be
>> based on Plugin2 as well:
>>
>> |Can't locate object method "new" via package "Dancer2::Plugin::DBIC" at /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/Dancer2/Core/App.pm line 64.
>> |
> 
> Yeeeeeah.... That's were my magic powers end. P1 plugins just don't play
> well with others, period. My skills are not good enough to fix that. :-/
> 
> `/.

OK, that is not a real problem. I just clone DBIC plugin and overhaul it :-).

Regards
	Racke

> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/PerlDancer/Dancer2/pull/1010#issuecomment-145014712>.
> 


-- 
Perl and Dancer Development

Visit our Perl::Dancer conference 2015.
More information on https://www.perl.dance.

yanick:
> I can fix that. Wait until after breakfast, and the patch will be there. :-)

Aaand it is. :-)

racke:
Thanks a lot, that works. Next stop is testing hooks in applications which is missing from Auth::Extensible tests.

racke:
The recipe for plugins using other plugins seems to work correctly as well :+1:

racke:
Another test balloon:

https://github.com/racke/Dancer2-Plugin-DBIC/commit/0fb0913d7778c390cd193d1d2f79c8b86bf16380

That was a quick change :-). It fails the last test in t/02-multiple-schemas.t
as it uses "set plugins => " twice.

cromedome:
That's freaking beautiful @racke! So much cleaner! :+1:

racke:
@cromedome: this is @yanick's candy - I'm merely the source code boy :-)

racke:
@yanick, do you have an idea where this error message comes from:

```
Overwriting existing sub 'TestApp::shop_country' with sub 'shop_country' exported by Dancer2::Plugin::Interchange6::Routes at t/routes/lib/TestApp.pm line 9.
```

yanick:
> |Overwriting existing sub 'TestApp::shop_country' with sub 'shop_country' exported by Dancer2::Plugin::Interchange6::Routes at t/routes/lib/TestApp.pm line 9.

Looks like the keyword get reimported many times. In
Dancer2::Plugin::Interchange6::Routes you might have to do

   use Dancer2::Plugin::Interchange6 ();

to stop the madness.

mickeyn:
sorry if I'm missing something, but I still see TODOs in the comments, aren't they blocking for the PR?
especially these: "# TODO probably wrong" :)

racke:
@mickeyn, we don't plan to merge the PR until we are pretty sure that is doing the right thing.

yanick:
On 2015-10-12 06:04 AM, Mickey wrote:
> sorry if I'm missing something, but I still see TODOs in the comments,
> aren't they blocking for the PR?

Not really.
https://github.com/PerlDancer/Dancer2/pull/1010#issuecomment-143772172

But now that I *DID* receive feedback, and have the thumb up of Racke
and Russel, I can clean up the code such that it'll work, and be
somewhat nice.

yanick:
On 2015-10-12 08:01 AM, Stefan Hornburg (Racke) wrote:
> we don't plan to merge the PR until we are pretty sure that is doing the
> right thing.

It might be a good idea to nail down what that means. Who is the "we",
and what are the checkmarks that need to be done such that we are
"pretty sure".

From Racker's excellent work, there is two things I agree with:

* Convert a few (let's use the magic number 10) plugins are see how well
that works.

* Add more tests, so that D2::P2 achieve over 90% coverage.

yanick:
Fleshing out the "pretty sure thing":

From Racker's excellent work, there is two things I agree with:

* Convert a few (let's use the magic number 10) plugins are see how well
that works.

* Add more tests, so that D2::P2 achieve over 90% coverage.

racke:
My goal is to make https://github.com/interchange/Dancer2-Plugin-Interchange6 work with Plugin2, which proves tricky with the code spread out in a couple files + using DBIC and Extensible  plugin.

SysPete:
@yanick I'm working on porting my [Dancer2::Plugin::DataTransposeValidator](https://github.com/interchange/Dancer-Plugin-DataTransposeValidator) plugin and wondering how to change plugin config on the fly within the test app. I currently have routes like this:
```
    post '/joined' => sub {
        set plugins => {
            DataTransposeValidator => {
                errors_hash => "joined"
            }
        };
        # do some stuff
    };
```
but I don't see how to achieve this using D2P2 - suggestions please.

SysPete:
and before I forget: dzil isn't installing the dep on MooX::ClassAttribute which I guess is an easy fix but I prefer not to touch dzil configs - I'm still a die-hard EUMM user.

racke:
On 10/13/2015 10:45 AM, Peter Mottram wrote:
> and before I forget: dzil isn't installing the dep on MooX::ClassAttribute which I guess is an easy fix but I prefer not to touch dzil configs - I'm still a die-hard EUMM user.
> 

Yes, that is fixed in Dancer2 master branch. So just install that module to get rid
of the errors.

Regards
	Racke


-- 
Perl and Dancer Development

Visit our Perl::Dancer conference 2015.
More information on https://www.perl.dance.

SysPete:
So now I have 6 apps in my main tests instead of one which gets around my earlier problem and I'm happy with that.

One other thing I tripped up on was $plugin->config returning undef when D2P's plugin_setting would return an empty hashref. Maybe this could be addressed?

yanick:
On 2015-10-13 05:47 AM, Peter Mottram wrote:
> So now I have 6 apps in my main tests instead of one which gets around
> my earlier problem and I'm happy with that.

Yeah, that's something we were discussing for another plugin as well.
The consensus right now is that one shouldn't modify the config itself
after plugins have been created (if not, it means a plugin would have to
always check back to the config for changes, which is a pain). So far,
that only seem to change things for tests, which should be okay. :cross
fingers:


> One other thing I tripped up on was $plugin->config returning undef when
> D2P's plugin_setting would return an empty hashref. Maybe this could be
> addressed?

I'll be! (see #1026)

yanick:
On 2015-10-13 04:45 AM, Peter Mottram wrote:
> and before I forget: dzil isn't installing the dep on
> MooX::ClassAttribute which I guess is an easy fix but I prefer not to
> touch dzil configs - I'm still a die-hard EUMM user.

#1027 created

racke:
@yanick, we also need this commit: 85f91f8. Otherwise dzil fails in case AnyEvent isn't installed.

yanick:
On 2015-10-13 08:58 AM, Stefan Hornburg (Racke) wrote:
> @yanick <https://github.com/yanick>, we also need this commit: 85f91f8
> <https://github.com/PerlDancer/Dancer2/commit/85f91f86d1fb6a8e05238e2fd4c15b97585fabc0>.
> Otherwise dzil fails in case AnyEvent isn't installed.

Done.

Thanks!

racke:
@yanick, maybe you can help me out with the following test:

https://github.com/PerlDancer/Dancer2-Plugin-Auth-Extensible/blob/topic/plugin2/t/02-as-plugin.t

It fails as follows:

```
t/02-as-plugin.t .. No Auth::Extensible realms configured with which to authenticate user at /home/racke/interchange/Dancer2-Plugin-Auth-Extensible/lib/Dancer2/Plugin/Auth/Extensible.pm line 1178.
[InsidePluginApp:28325] error @2015-10-13 15:29:30> Route exception: Can't locate object method "requires_login" via package "Dancer2::Plugin::Auth::Extensible" at (eval 152) line 17. in /home/racke/perl5/perlbrew/perls/perl-5.22.0/lib/site_perl/5.22.0/Dancer2/Core/App.pm l. 1322
```

I suppose I'm doing something wrong ...

ramortegui:
Hi @yanick, 

It seems the plugin configuration is not available on the BUILD method.

    sub BUILD {
      my $plugin = shift;
      use Data::Dumper;
      print Dumper($plugin->config); #prints $VAR1 = {};
    }

That's the reason of this message reported by @rake: 
t/02-as-plugin.t .. No Auth::Extensible realms configured with which to authenticate user at /home/racke/interchange/Dancer2-Plugin-Auth-Extensible/lib/Dancer2/Plugin/Auth/Extensible.pm line 1178.

veryrusty:
We try and keep the core fat-packable and `Sub::Attribute` involves XS. Does Plugin2 count as "core" ?

yanick:
On 2015-10-19 02:35 PM, Russell Jenkins wrote:
> We try and keep the core fat-packable and |Sub::Attribute| involves XS.

I can switch down to Attribute::Handlers.

> Does Plugin2 count as "core" ?

It could be. Or, if the few tweaks in D2::App are done, I could release
it as an independent module. So that by default D2 can support the
legacy system by default.

yanick:
On 2015-10-19 12:26 PM, Ruben Amortegui wrote:
> It seems the plugin configuration is not available on the BUILD method.
> 
> |sub BUILD {
>   my $plugin = shift;
>   use Data::Dumper;
>   print Dumper($plugin->config); #prints $VAR1 = {};
> }
> |
> 
> That's the reason of this message reported by @rake
> <https://github.com/rake>:
> t/02-as-plugin.t .. No Auth::Extensible realms configured with which to
> authenticate user at
> /home/racke/interchange/Dancer2-Plugin-Auth-Extensible/lib/Dancer2/Plugin/Auth/Extensible.pm
> line 1178.
> 

Yeah, we root-caused it earlier. The config was not in the app space, so
wasn't picked by the app, which in turn wasn't picked by its plugin.

Thanks!

`/.

veryrusty:
The current plugin implementation allows for the prototype of keywords to be specified. Is it possible to defined the prototype of the keywords imported by a plugin2 ?

yanick:
On 2015-10-19 02:53 PM, Russell Jenkins wrote:
> The current plugin implementation allows for the prototype of keywords
> to be specified. Is it possible to defined the prototype of the keywords
> imported by a plugin2 ?

It doesn't bother with prototypes right now, but it could be added.

yanick:
Sub::Attribute removed for Attribute::Handlers

Test coverage is now at 88%.

ramortegui:
Hi @yanick, one question, how can I enable ( or disable) routes when my plugin is loading with Plugin2?

eg. I was thinking something like this ( getting access to the configs ) :

    sub BUILD {
      my $plugin = $shift;
      my $app = $plugin->app;
      my $settings = $self->config;
      if( $settings->{default_routes} ne '0' ){
        $app->add_route => {
          method ......
          ......
        }
      }
    }

yanick:
On 2015-10-21 10:24 AM, Ruben Amortegui wrote:
> Hi @yanick <https://github.com/yanick>, one question, how can I enable (
> or disable) routes when my plugin is loading with Plugin2?
> 
> eg. I was thinking something like this ( getting access to the configs ) :
> 
> |sub BUILD {
>   my $plugin = $shift;
>   my $app = $plugin->app;
>   my $settings = $self->config;
>   if( $settings->{default_routes} ne '0' ){
>     $app->add_route => {
>       method ......
>       ......
>     }
>   }
> }
> |

Yup, that's pretty much it. :-)

You have the gist of it. The only thing I could add is to sprinkle a
little bit of attribute sugar on top:


```
has default_routes => (
  is => 'ro',
  from_config => sub { 0 },
);

sub BUILD {
  my $plugin = shift;

  $plugin->app->add_route( ... ) if $plugin->default_routes;

   # and so on, and so forth

}

```

ramortegui:
Hi @yanick, I was looking into t/plugins2-from-config.t, and when I try to get an attribute of the plugin on the BUILD method the test fails. 

I just added the next code inside the BEGIN bloque:

    sub BUILD{
      my $self = shift;
      my $one = $self->one;
    }

The config isn't  available at this point.

Any idea?

yanick:
On 2015-10-21 02:09 PM, Ruben Amortegui wrote:
> Hi @yanick <https://github.com/yanick>, I was looking into
> t/plugins2-from-config.t, and when I try to get an attribute of the
> plugin on the BUILD method the test fails.
> 
> I just added the next code inside the BEGIN bloque:
> 
> |sub BUILD{
>   my $self = shift;
>   my $one = $self->one;
> }
> |
> 
> The config isn't available at this point.
> 
> Any idea?

Yup. Makes total sense once you look close enough. :-)


If you do something with `one` in the BUILD, what happens is:

In `MyApp`, when you hit the line `use Dancer2::Plugin:FromConfig;`,
the plugin class create the instance of the plugin to inject into the app.

Creating the instance means that underneath a
`Dancer2::Plugin::FromConfig->new` call is being made...

... Which in turn uses `BUILD` ...

... `BUILD` calls on the `one` attribute ...

... and, *BAM*, indeed the config is not yet available because it's
defined in the next few lines.


This won't happen in the regular case, as the config will be read from
inside the config.yml file and not everything will be mashed inside the
same file. If you want to see it work, you can do:

    BEGIN {
    set plugins => {
        FromConfig => {
            one => 'un',
            two => {
                three => 'trois',
            }
        }
    };
    }

    use Dancer2::Plugin::FromConfig;


That will ensure that the config is set at compile time, and that the
plugin instance will be created right after its values have been populated.


Makes sense? :-)

ramortegui:
Thanks @yanick makes sense.

I have some issues with DBIC and the plugin(that uses DBIC -and works fine ) that I'm writing.  The problem is when I'm using DBIC outside my plugin.  So I wrote this test to load two plugins.  But it seems that is overwriting the subroutines.

    use strict;
    use warnings;
    use Test::More;

    {
      package Dancer2::Plugin::Plugin1;
      use Dancer2::Plugin2;
      BEGIN{
        has one => (
          is => 'ro',
          default => sub { 'uno' },
        );
        plugin_keywords 'one';
      }
    }

    {
      package Dancer2::Plugin::Plugin2;
      use Dancer2::Plugin2;
      BEGIN{
        has two => (
          is => 'ro',
          default => sub { 'dos' }
        );
        plugin_keywords 'two';
     
      }
    }

    {
        package MyApp;

        use Dancer2;
        use Dancer2::Plugin::Plugin1;
        use Dancer2::Plugin::Plugin2;

        Test::More::is one() => 'uno', 'from plugin1';
        Test::More::is two() => 'dos', 'from plugin2';
    }

    done_testing();


and this is the result:

    t/plugins2-uses-plugin2.t .. Overwriting existing sub 'MyApp::two' with sub 'two' exported by Dancer2::Plugin::Plugin2 at t/plugins2-uses-plugin2.t line 35.
    Overwriting existing sub 'MyApp::one' with sub 'one' exported by Dancer2::Plugin::Plugin2 at t/plugins2-uses-plugin2.t line 35.
    t/plugins2-uses-plugin2.t .. 1/? 
    #   Failed test 'from plugin1'
    #   at t/plugins2-uses-plugin2.t line 37.
    #          got: undef
    #     expected: 'uno'
    # Looks like you failed 1 test of 2.
    t/plugins2-uses-plugin2.t .. Dubious, test returned 1 (wstat 256, 0x100)
    Failed 1/2 subtests 
    
    Test Summary Report
    -------------------
    t/plugins2-uses-plugin2.t (Wstat: 256 Tests: 2 Failed: 1)
      Failed test:  1
      Non-zero exit status: 1
    Files=1, Tests=2,  1 wallclock secs ( 0.03 usr  0.00 sys +  0.41 cusr  0.01 csys =  0.45 CPU)
    Result: FAIL

yanick:
On 2015-10-21 04:39 PM, Ruben Amortegui wrote:
> I have some issues with DBIC and the plugin(that uses DBIC -and works
> fine ) that I'm writing. The problem is when I'm using DBIC outside my
> plugin. So I wrote this test to load two plugins. But it seems that is
> overwriting the subroutines.

You found a bug! \o/ Well, many of them.

I've updated the repo with the fix, and some more tests to replicate
your usecase. Can you try it and let me know if it works better for you now?

Hopeful,
`/anick

ramortegui:
Yes @yanick, it works! Thanks!

yanick:
On 2015-10-22 07:08 PM, Ruben Amortegui wrote:
> Yes @yanick <https://github.com/yanick>, it works! Thanks!

WOOHOO! \o/

racke:
Hooks will be a problem for existing installations, as you have to say 

`hook 'plugin.myplugin.myhook'`

instead of just

`hook myhook`

racke:
Also hooks aren't really covered in the Plugin2 POD.

yanick:
On 2015-10-24 03:02 AM, Stefan Hornburg (Racke) wrote:
> Hooks will be a problem for existing installations, as you have to say
> 
> |hook 'plugin.myplugin.myhook'|
> 
> instead of just
> 
> |hook myhook|

Yes. And it's terrible that there is nothing, absolutely nothing we can
do about it.

o___o

Seriously, though:

1. the already existing plugins of already existing installations won't
change.

2. Isn't that exactly the behavior of P1?
https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Plugin.pm#L146

yanick:
On 2015-10-24 03:05 AM, Stefan Hornburg (Racke) wrote:
> Also hooks aren't really covered in the Plugin2 POD.

Fixed.

xsawyerx:
I haven't replied on this issue, but talking to the devs personally, I gave my :+1:.

@yanick, amazing work.

abeverley:
@yanick, thanks for your efforts with this - great progress.

I'm in the process of migrating Dancer2::Plugin::LogReport. I've got most of it working, but a few questions/comments:

1. (possibly more of a Moo question): I was using ```sub import {}``` to import a bunch of functions into the Dancer2 app, but I'm not sure how to do this with P2, as ```import()``` is never called. I realise keywords would normally be used for this, but in this case I want to "pass through" a whole bunch of functions from the parent module, and just add the Dancer specific items as keywords. Does that make sense and is there an easy solution to this?

2. I'd like to be able to access the template engine during ```BUILD()``` or soon after. I couldn't do this with the original architecture, but it would be nice to be able to do it with P2. I realise this is a bit chicken and egg, and from what I can tell the plugin is built before the template engine, but I wonder whether there could be some sort of solution, like a "post app build" hook or something. I can access the template engine during a subsequent keyword call, but that's a bit hacky and a little late too.

3. Minor point. I've found myself writing ```$app->add_hook(Dancer2::Core::Hook->new(...``` a few times. Is it worth having a shortcut, a bit like ```DSL::Hook```. In fact, should the full DSL be available to the plugin somehow?

Thanks,

Andy

yanick:
On 2015-10-24 06:27 PM, Andy Beverley wrote:
> @yanick <https://github.com/yanick>, thanks for your efforts with this -
> great progress.

Awww, thank you. ^.^

>  1.
> 
>     (possibly more of a Moo question): I was using |sub import {}| to
>     import a bunch of functions into the Dancer2 app, but I'm not sure
>     how to do this with P2, as |import()| is never called. I realise
>     keywords would normally be used for this, but in this case I want to
>     "pass through" a whole bunch of functions from the parent module,
>     and just add the Dancer specific items as keywords. Does that make
>     sense and is there an easy solution to this?

The most easy solution is simply to export the functions via
`plugin_keywords`, and just ignore the first argument in `@_`, which
will be the plugin object itself. Or, if you want to make that
transparent to the exported function as well:

```
package Dancer2::Plugin::Foo;

use Dancer2::Plugin2;

sub im_exported { ... }

plugin_keywords im_exported => sub {
	shift; # lose the '$plugin' argument
	im_exported(@_);  # or could be 'goto &im_exported'
};

```

>  2.
> 
>     I'd like to be able to access the template engine during |BUILD()|
>     or soon after. I couldn't do this with the original architecture,
>     but it would be nice to be able to do it with P2. I realise this is
>     a bit chicken and egg, and from what I can tell the plugin is built
>     before the template engine, but I wonder whether there could be some
>     sort of solution, like a "post app build" hook or something. I can
>     access the template engine during a subsequent keyword call, but
>     that's a bit hacky and a little late too.


I don't know the specifics of what you're trying to do, but I'd use a
lazy attribute to achieve that post-build type of behavior:


```
has template => (
  is => 'ro',
  lazy => 1,
  default => sub {
     # peusdo-code. Don't remember the right call
     $_[0]->app->engines('template');
  }
);

sub mykeyword :PluginKeyword {
	my $plugin = shift;

	# 'template' will be instantiated on the
	# first call to 'mykeyword', and so the
	# template will be there
	my $template = $plugin->template;

	...
}

```

Hmm... I have to check. I'm pretty sure there might be another way as
well. I'll get back to you on that.

>  3.
> 
>     Minor point. I've found myself writing
>     |$app->add_hook(Dancer2::Core::Hook->new(...| a few times. Is it
>     worth having a shortcut, a bit like |DSL::Hook|. In fact, should the
>     full DSL be available to the plugin somehow?

The DSL can't be made available. That's probably something I should blog
about. But in a very tight nutshell: the DSL is applied to the namespace
and the *class* at buildtime, and the application -- which the DSL needs
-- is only tied to the *object* at runtime, and can be tied to several
objects. Trying to define at class-level, but bind at object-level is
the big oooopsie of Plugin1 that Plugin2 tries to fix.

This being said, you're not entierely SOL. You can mitigate the pain by
doing:

```
has '+app' => (
   handles => [ 'add_hook' ],
);
```

And at least you only have to write `$plugin->add_hook` instead of
`$plugin->app->add_hook`.


Hope that helps!
`/anick

yanick:
On 2015-10-26 11:07 AM, Yanick Champoux wrote:
> Hmm... I have to check. I'm pretty sure there might be another way as
> well. I'll get back to you on that.

I was thinking about engine build-time hooks. It's not there right now,
but...

https://github.com/PerlDancer/Dancer2/issues/1047

I'll be working on that. :-)

abeverley:
On Mon, 2015-10-26 at 09:37 -0700, Yanick Champoux wrote:
> I was thinking about engine build-time hooks. It's not there right 
> now, but...
> 
> https://github.com/PerlDancer/Dancer2/issues/1047
> 
> I'll be working on that. :-)

Awesome, thanks Yanick.

For the record, the motivation is that I want my plugin to be able to
add a filter to Template::Toolkit.

Of course, such a hook will also allow applications to do the same
thing.

Something like:

    $app->template_engine->engine->context->define_filter(...)

(obviously having tested that TT is the template engine)

Thanks for the other comments - I'll reply to them soon.

Andy

veryrusty:
Added a PR against @yanick's branch to remove Test::WWW::Mechanize::PSGI as a test dependency (replaced with Plack::Test + HTTP::Request::Common).

yanick:
@veryrusty patch pulled in. :+1:

xsawyerx:
I added this commit: https://github.com/PerlDancer/Dancer2/commit/847d94930b538b748479075d0c22ec87df673a74 to get rid of `MooX::ClassAttribute`.

xsawyerx:
I added the following commits, which implement the old syntax in this new implementation:

* https://github.com/PerlDancer/Dancer2/commit/d545b758732924ef93b127a5d12edfc653c486f5
* https://github.com/PerlDancer/Dancer2/commit/29c563e2374785ca3341d5cc5509532936dc0841
* https://github.com/PerlDancer/Dancer2/commit/9f0418ccba951b8fb0c1904e0fb84d42d91d9307
* https://github.com/PerlDancer/Dancer2/commit/5d99189a2d8838c76bfa27c6975d446e954e98c0
* https://github.com/PerlDancer/Dancer2/commit/fb3ee55dfd2f5aaeee815b6eacbf532c46cb5b7f

@yanick, Do you mind if I open a new PR from the branch on this repo?

xsawyerx:
Added:

* 4bbdb1629a61a46380df3a48f0f06c15614df523: What I *think* is a small fix. @yanick, verify please.
* 454f2ca8332b655798031c3136f982e161f376fd: `execute_plugin_hook` so you don't need to call the full hook name. Discussed here as well.

racke:
@xsawyerx: Thanks a lot for your work on having a drop-in replacement for current Plugin.pm.

BTW, we are getting the following test failure.

```
racke@argus:~/dancer2$ prove -vl t/plugin2/hooks.t 
t/plugin2/hooks.t .. 
1..6
ok 1
ok 2
ok 3
[PoC:17958] error @2015-11-05 10:08:03> Exception caught in 'core.app.after_request' filter: Hook error: Hook 'plugin.foodetector.foo' does not exist at /home/racke/dancer2/core/lib/Dancer2/Core/App.pm line 590.
 at /home/racke/dancer2/core/lib/Dancer2/Core/App.pm line 1025. in /home/racke/dancer2/core/lib/Dancer2/Core/App.pm l. 590
not ok 4

#   Failed test at t/plugin2/hooks.t line 60.
ok 5
not ok 6

#   Failed test at t/plugin2/hooks.t line 63.
#          got: 'nope'
#     expected: 'hooked'
# Looks like you failed 2 tests of 6.
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/6 subtests 

Test Summary Report
-------------------
t/plugin2/hooks.t (Wstat: 512 Tests: 6 Failed: 2)
  Failed tests:  4, 6
  Non-zero exit status: 2
Files=1, Tests=6, 30 wallclock secs ( 0.04 usr  0.00 sys +  0.42 cusr  0.02 csys =  0.48 CPU)
Result: FAIL
```

yanick:
On 2015-11-04 02:07 PM, Sawyer X wrote:
>   * <https://github.com/PerlDancer/Dancer2/commit/fb3ee55dfd2f5aaeee815b6eacbf532c46cb5b7f>
> 
> @yanick <https://github.com/yanick>, Do you mind if I open a new PR from
> the branch on this repo?

@sawyer Go ahead. If it's not too onerous, I'd prefer if the d2p1
compatibility code was in its own  sub-branch until we get most of the
plugin test suites passing.

yanick:
On 2015-11-05 03:08 AM, Sawyer X wrote:
>   * 4bbdb16
>     <https://github.com/PerlDancer/Dancer2/commit/4bbdb1629a61a46380df3a48f0f06c15614df523>:
>     What I /think/ is a small fix. @yanick <https://github.com/yanick>,
>     verify please.

	Yup. That was me being distracted. Good catch.

>   * 454f2ca
>     <https://github.com/PerlDancer/Dancer2/commit/454f2ca8332b655798031c3136f982e161f376fd>:
>     |execute_plugin_hook| so you don't need to call the full hook name.
>     Discussed here as well.

I've put a comment with the commit. I think the `caller` juggling can be
entirely side-stepped.

xsawyerx:
@racke Fixed! All tests pass, including new syntax and old one.

av8r:
Hi

Any chance to have it merge for Xmas ?

yanick:
¯\_(ツ)_/¯

xsawyerx:
Unfortunately merging it is a very complicated task. It will require either finishing the compatibility layer (I've been working on that but other priorities got in the way) or breaking compatibility substantially or branching off to yet another implementation (and having two Dancers and several Dancer plugin implementation). I'm really against the last two options, but I can't be certain the first will be done by xmas.

I'll work heavily on it this week and hopefully so.

I appreciate the patience. We want to get this done right.

racke:
On 12/21/2015 11:01 AM, Sawyer X wrote:
> Unfortunately merging it is a very complicated task. It will require either finishing the compatibility layer (I've been working on that but other priorities got in the way) or breaking compatibility substantially or branching off to yet another implementation (and having two Dancers and several Dancer plugin implementation). I'm really against the last two options, but I can't be certain the first will be done by xmas.
> 
> I'll work heavily on it this week and hopefully so.
> 
> I appreciate the patience. We want to get this done right.
> 

It certainly can be done by Xmas of some year :-).

Regards
	Racke


-- 
Perl and Dancer Development

Visit our Perl::Dancer conference 2015.
More information on https://www.perl.dance.

xsawyerx:
So, uhh.. it seems to work, except for two things:

1. `use Dancer2` + `use Dancer2::Plugin` breaks. We don't currently detect this.
2. `use Dancer2::Plugin::MyPlugin 'my', 'specific', 'subs'` breaks. We currently detect this and ask you to change it.

Everything else *seems* to work. We throw a bunch of warnings for deprecated behavior now. We could probably throw even more, but with time. :)

Please to test! I will be testing a few modules and will email the user list for any plugin authors to do the same. Thank you.

SysPete:
@xsawyerx I created new tests and removed `use Dancer2` from [Dancer2::Plugin::DataTransposeValidator](https://github.com/interchange/Dancer-Plugin-DataTransposeValidator) and I'm seeing the expected deprecation warning:
`Using DSL in plugins is deprecated (setting). at ./t/d2plugin.t line 49.`
which is fair enough but this is not so good:
`Use of uninitialized value $app_plugin_config in hash dereference at /home/apm/git/Dancer2/lib/Dancer2/Plugin.pm line 263.`

In addition my pod coverage tests are now useless...
```
# Coverage for Dancer2::Plugin::DataTransposeValidator is 6.7%, with 14 naked subroutines:
#       ClassHooks
#       PluginKeyword
#       dancer_app
#       execute_plugin_hook
#       hook
#       keywords
#       on_plugin_import
#       plugin_args
#       plugin_setting
#       register
#       register_hook
#       register_plugin
#       request
#       var
```
I can cope with this since it is RELEASE_TESTING only but is kinda sucky.

Otherwise it does at least pass all tests whilst still being old-style D2 plugin.

xsawyerx:
@SysPete The problem with the POD tests is that you effectively **do** get those subroutines now. I can try and remove some of them but can't remove all of them. `PluginKeyword`, for instance, is provided by the new system. There's no way around that. I can try and hide them all behind `AUTOLOAD` but that really is an escape hatch for *unsupported* old plugin system keywords, so I'd rather not have additional hackery there just to accommodate a test.

Also, @dams had secret plans to remove my `AUTOLOAD` hackery to begin with (which I'm hoping he'll actually go through with!) :)

xsawyerx:
I've added this plugin to #1078.

xsawyerx:
@SysPete removed the uninitialized warnings you received.

dams:
Wow this is not a small PR :)

abeverley:
Sorry for the slow reply on this:

> > I was using |sub import {}| to import a bunch of functions into the 
> > Dancer2 app, but I'm not sure how to do this with P2, as |import()| 
> > is never called. I realise keywords would normally be used for this, 
> > but in this case I want to "pass through" a whole bunch of functions 
> > from the parent module, and just add the Dancer specific items as 
> > keywords. Does that make sense and is there an easy solution to 
> > this?
> 
> The most easy solution is simply to export the functions via
> `plugin_keywords`, and just ignore the first argument in `@_`, which
> will be the plugin object itself. Or, if you want to make that
> transparent to the exported function as well:
> 
> ```
> package Dancer2::Plugin::Foo;
> 
> use Dancer2::Plugin2;
> 
> sub im_exported { ... }
> 
> plugin_keywords im_exported => sub {
> 	shift; # lose the '$plugin' argument
> 	im_exported(@_);  # or could be 'goto &im_exported'
> };
> 
> ```

I'm still struggling with this. I've found that import() is being called
in my plugin, but if it exists, then the rest of my plugin isn't
properly importing (BUILD isn't called, and my keywords aren't
exported).

I can't use Yanick's suggestions above, as the import() of the depended
-upon module has so many options and so many functions that I can't
realistically replicate them in my plugin (and even if I could it would
be a lot of duplicate code).

I'm going to carry on trying to work out what is going on here, but if
anyone has any ideas/suggestions in the mean time, then please let me
know.

Thanks,

Andy

abeverley:
> I'm still struggling with this. I've found that import() is being called
> in my plugin, but if it exists, then the rest of my plugin isn't
> properly importing (BUILD isn't called, and my keywords aren't
> exported).

I'm gradually getting to the bottom of this. It seems that my ```import``` sub is overriding the one from ```Try::Tiny```, which is causing the standard import functionality to not run. I've found that I can use ```_exporter_validate_opts()``` instead, which appears to work, but I am not sure whether by doing so I am abusing that particular function.

SysPete:
@xsawyerx having fun with Dancer2::Plugin::Shutdown which composes in a Moo role and explodes in an ugly way:

```
Can't use string ("-1") as a symbol ref while "strict refs" in use at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Role/Tiny.pm line 361.
        Role::Tiny::_install_methods("Moo::Role", "Dancer2::Plugin::Shutdown", "Dancer2::Plugin::Role::Shutdown") called at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Role/Tiny.pm line 92
        Role::Tiny::apply_single_role_to_package("Moo::Role", "Dancer2::Plugin::Shutdown", "Dancer2::Plugin::Role::Shutdown") called at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Moo/Role.pm line 271
        Moo::Role::apply_single_role_to_package("Moo::Role", "Dancer2::Plugin::Shutdown", "Dancer2::Plugin::Role::Shutdown") called at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Role/Tiny.pm line 196
        Role::Tiny::apply_role_to_package("Moo::Role", "Dancer2::Plugin::Shutdown", "Dancer2::Plugin::Role::Shutdown") called at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Role/Tiny.pm line 201
        Role::Tiny::apply_roles_to_package("Moo::Role", "Dancer2::Plugin::Shutdown", "Dancer2::Plugin::Role::Shutdown") called at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Moo/Role.pm line 263
        Moo::Role::apply_roles_to_package("Moo::Role", "Dancer2::Plugin::Shutdown", "Dancer2::Plugin::Role::Shutdown") called at /home/apm/.perlbrew/libs/perl-5.20.3@main/lib/perl5/Moo.pm line 39
        Moo::with("Dancer2::Plugin::Role::Shutdown") called at lib/Dancer2/Plugin/Shutdown.pm line 8
        require Dancer2/Plugin/Shutdown.pm called at t/000.t line 11
        Webservice::BEGIN() called at lib/Dancer2/Plugin/Shutdown.pm line 0
        eval {...} called at lib/Dancer2/Plugin/Shutdown.pm line 0
Compilation failed in require at t/000.t line 11.
        Webservice::BEGIN() called at t/000.t line 11
        eval {...} called at t/000.t line 11
BEGIN failed--compilation aborted at t/000.t line 11.

```
extract from Dumping `$stash` just before `lib/perl5/Role/Tiny.pm line 361`:
```
  plugin_keywords => *Dancer2::Plugin::Shutdown::plugin_keywords,
  plugin_setting => -1,
  register => *Dancer2::Plugin::Shutdown::register,
```
So how does plugin_setting end up being `-1` ? Very weird.

shadowcat-mst:
Class::Load is obsoleted by Module::Runtime - and in fact is just a thin wrapper around the latter

Speaking as somebody who's reimplemented the Catalyst handling for them twice, attributes are a really terrible idea.

I can't see where keywords are added to the DSL object.

The idea, originally, for Dancer2 plugins as sukria and I planned them, was to have the plugin able to extend the app by applying roles to things - i.e. plugins should not, normally, be defining keywords directly. They should instead be defining methods that can be added to the app object, or to e.g. request classes or whatever, which then get curried versions of themselves exported via the DSL.pm code - which provides a map of what it needs to export specifically to enable this (i.e. basically a plugin becomes a sort of htt://p3rl.org/CatalystX::RoleApplicator like thing under the hood but with nice sugar so your users don't have to deal with Package::Variant etc. directly). 

That way a plugin can, if it wants to, override a core DSL keyword, and the re-export of the $dsl should take care of replacing that without running into a problem.

Note that this then also means that a standard $dsl can be exported into the plugin namespace, avoiding the weirdness where people end up having to try and call methods on a passed-in $dsl object. All that's required to make that work is to 'local' the relevant $app for the plugin's $dsl as a call passes into the plugin.

So, basically, converting it to being a Moo class is totally a start, but now you need to throw even more Moo at it and then we'll get somewhere.

xsawyerx:
I think I'll need to read this several times (or once when I'm focused) to fully understand this.

abeverley:
Just to clarify for anyone else looking at this, the development is no longer on the repo/branch of this PR. It is at a branch of the same name on the PerlDancer repo: https://github.com/PerlDancer/Dancer2/tree/plugins-yanick

abeverley:
And more discussions are at the issue that is already linked above (#1078)

racke:
Heads up: the code has now merged into master.  Any leftovers which prevent us to close the issue?

xsawyerx:
Probably #1138. It doesn't have to be solved but it has to be resolved. :)

abeverley:
> Heads up: the code has now merged into master. Any leftovers which prevent us to close the issue?

Yeah, last time I checked this breaks ```Dancer2::Plugin::LogReport```. Since then, I haven't had time to look any further into it. Has anything changed that may have mitigated that?

abeverley:
> Yeah, last time I checked this breaks Dancer2::Plugin::LogReport

Can I request that this PR does not make itself into a Dancer2 release until this is resolved please?

SysPete:
@abeverley plugin2 has been merged into master branch so we now need to get remaining issues resolved.

For `Dancer2::Plugin::Auth::Extensible` I'm reworking it for new style plugin2 with all compat code removed and once I have it working will PR the DBIC provider. The plan is to release a new D2PAE at the same time that the Dancer2 release with plugin2 hits the CPAN. Trying to rehack horrible compat code into plugin2 for this use case is probably not worth the effort considering how few plugins are affected.

Are you OK to follow the same route with LogReport?

abeverley:
> For Dancer2::Plugin::Auth::Extensible I'm reworking it for new style plugin2 with all compat code removed and once I have it working will PR the DBIC provider.

Great, thanks.

> Trying to rehack horrible compat code into plugin2 for this use case is probably not worth the effort considering how few plugins are affected.
> 
> Are you OK to follow the same route with LogReport?

More than happy with that approach. I just want to avoid a situation where I install the latest Dancer2 release and end up with an unusable project. I need to pull the latest code for both and then see where we're at. Basically, the original P2 code worked fine, it was some of the later work that caused some problems.

SysPete:
Just ran into another issue with plugin2:

I've been working on improving Dancer2::Session::DBIC and the final feature added was the ability to use Dancer2::Plugin::DBIC:

https://github.com/interchange/Dancer2-Session-DBIC/commit/bc4e9aaa6d68ff03de0cfaf32f4d839e96cf3b4d

This works fine with old plugin but with plugin2 fails horribly:

https://github.com/interchange/Dancer2-Session-DBIC/issues/10

I decided to try passing the app into the session engine in `_build_session_engine` in ::App so I could try using plugin2's with_plugin method but that just ends up failing in exactly the same way as issue 10.

The only way to make things work in the end was to import `Dancer2` and `Dancer2::Plugin::DBIC` into `Dancer2::Session::DBIC` and use D2PD's `schema` keyword but this just feels completely wrong.

Any thoughts as to a sensible direction to take?

hvoers:
@SysPete  I would
`use DBICx::Sugar qw(schema);`

SysPete:
@hvoers many thanks - works perfectly. Now just need to get Dancer2::Plugin::Auth::Extensible::Provider::{Database|DBIC} to work with plugin2.

xsawyerx:
So we have two remaining problems:

* @SysPete's work on getting D2::P::Auth::Extensible::Provider::{Database|DBIC} working with D2P2.
* @abeverley's project

Am I right? Let's sort it out!

SysPete:
I pushed this terrible commit to D2PAE for Provider::Database:

https://github.com/PerlDancer/Dancer2-Plugin-Auth-Extensible/commit/f656775f76008fdeacb6e4e685b261b2dbcc5035

I was hoping it would make someone scream. That commit needs to be consigned to a brown envelope and lost in a dark dungeon. But it does make one big point: I can't work out how to get access to Dancer2::Plugin::Database DSL from Provider::Database.

Suggestions?

SysPete:
And in case someone reads my comment above regarding `Dancer2::Session::DBIC` I'd like to get access to `Dancer2::Plugin::DBIC` DSL in there too. Right now I'm doing something utterly revolting by importing `DBICx::Sugar` but that is a hack too terrible to ponder since it only actually works if `Dancer2::Plugin::DBIC` has already been imported and its `schema` keyword has been called. Gets me to where I want to be but I have to hide under the blanket of shame again. :unamused:

yanick:
On 2016-03-29 06:49 AM, Peter Mottram wrote:
> I can't work out how to get access to Dancer2::Plugin::Database DSL from
> Provider::Database.
> 
> Suggestions?

I was going for the obvious "Don't, use
`$app->plugin('Foo')->whatever_method` instead", but you probably are
trying to make it work in the transitory P1/P2 context. In which case,
uh, I don't know. I really don't know.

`/.

SysPete:
On 29/03/16 15:21, Yanick Champoux wrote:
> On 2016-03-29 06:49 AM, Peter Mottram wrote:
> > I can't work out how to get access to Dancer2::Plugin::Database DSL from
> > Provider::Database.
> >
> > Suggestions?
>
> I was going for the obvious "Don't, use
> `$app->plugin('Foo')->whatever_method` instead", but you probably are
> trying to make it work in the transitory P1/P2 context. In which case,
> uh, I don't know. I really don't know.
>
> `/.

I've given up trying to make it work for p1/p2 compat and am trying to
make it work just for p2. Part of the problem is that the Provider
classes are not actually part of the plugin itself so I tried passing
the Dancer2::Plugin::Auth::Extensible plugin object as an attribute to
the Database provider's constructor then in the provider I have:

```
has dancer2_plugin_database => (
    is => 'ro',
    lazy => 1,
    default => sub {
        $_[0]->plugin->app->with_plugin('Dancer2::Plugin::Database');
    },
    handles => { database => 'database' },
);
```
and then in provider methods we have this:
```
my $database = $self->database($settings->{db_connection_name});
```
and all should be cool except that we end up with these errors (which I
remember seeing somewhere a long time ago):
```
Route exception: Can't find originating plugin at (eval 296) line 17.
Use of uninitialized value $Dancer2::Plugin::CUR_PLUGIN in string at
/home/apm/git/Dancer2/lib/Dancer2/Plugin.pm line 247.
```

I've tried all sorts of things but I'm still stuck at this point.

yanick:
On 2016-03-29 10:20 AM, Peter Mottram wrote:
> Route exception: Can't find originating plugin at (eval 296) line 17.
> Use of uninitialized value $Dancer2::Plugin::CUR_PLUGIN in string at
> /home/apm/git/Dancer2/lib/Dancer2/Plugin.pm line 247.

Yeah, `$CUR_PLUGIN` is called as part of 'plugin_setting'. Maybe find
where `plugin_setting` is called from, and change that for a
`$plugin->config`?

SysPete:
On 29/03/16 16:55, Yanick Champoux wrote:
> On 2016-03-29 10:20 AM, Peter Mottram wrote:
> > Route exception: Can't find originating plugin at (eval 296) line 17.
> > Use of uninitialized value $Dancer2::Plugin::CUR_PLUGIN in string at
> > /home/apm/git/Dancer2/lib/Dancer2/Plugin.pm line 247.
>
> Yeah, `$CUR_PLUGIN` is called as part of 'plugin_setting'. Maybe find
> where `plugin_setting` is called from, and change that for a
> `$plugin->config`?

Doh! Thanks for the reminder @yanick. Sometimes I need someone else to
state the obvious even though I've already created PRs in other plugins
for the same reason.

@bigpresh: now I just need a new release of Dancer2::Plugin::Database
with my latest PR (pretty please) so I can get this thing ready.

hvoers:
On Tue, 29 Mar 2016, Peter Mottram wrote:

> And in case someone reads my comment above regarding `Dancer2::Session::DBIC` I'd like to get access to `Dancer2::Plugin::DBIC` DSL in there too. Right now I'm doing something utterly revolting by importing `DBICx::Sugar` but that is a hack too terrible to ponder since it only actually works if `Dancer2::Plugin::DBIC` has already been imported and its `schema` keyword has been called. Gets me to where I want to be but I have to hide under the blanket of shame again. :unamused:

It's not that bad. I had a BEGIN block in my app allready.

Just add:

DBICx::Sugar::config( config->{plugins}->{DBIC} );

SysPete:
I think its finally time we closed this one. Any objections?

cromedome:
👍  to close.

racke:
Please go ahead :+1:

yanick:
yup, I think this one is done. :-)

PR #1012, opened 2015-09-29
Introduce ':nopragmas' import option

xsawyerx:
On a discussion on p5p, the idea of Dancer, Moose, et. al. clobbering
the caller has been raised as an issue. The best way to deal with
this is having the modules provide an ability to not import pragmas
into the caller.

Instead of "no_warnings" configuration options (like in Dancer 1)
or specific import flags for controlling specific pragmas and how
they are imported, we introduce a specific flag for cleaning all
pragmas, and the user controls what to do.

    use Dancer2 ':nopragmas'; # no pragmas, only syntax

This *might* seem similar to 'use Dancer2 ':syntax'" but the concept
of *that* was to prevent the creation of a server. In Dancer2, since
we're using Plack, this is superfluous. Instead, the ':nopragmas'
makes it clear what the distinction is.

### DIFF
--- a/lib/Dancer2.pm
+++ b/lib/Dancer2.pm
@@ -24,12 +24,18 @@ sub import {
     my ( $class,  @args   ) = @_;
     my ( $caller, $script ) = caller;
 
-    $_->import::into($caller) for qw(strict warnings utf8);
-
     my @final_args;
+    my $clean_import;
     foreach my $arg (@args) {
+        # ignore, no longer necessary
+        # in the future these will warn as deprecated
         grep +( $arg eq $_ ), qw<:script :syntax :tests>
             and next;
+
+        if ( $arg eq ':nopragmas' ) {
+            $clean_import++;
+            next;
+        }
 
         if ( substr( $arg, 0, 1 ) eq '!' ) {
             push @final_args, $arg, 1;
@@ -37,6 +43,9 @@ sub import {
             push @final_args, $arg;
         }
     }
+
+    $clean_import
+        or $_->import::into($caller) for qw<strict warnings utf8>;
 
     scalar @final_args % 2
       and die q{parameters must be key/value pairs or '!keyword'};
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -2151,9 +2151,9 @@ I<not> create an application with the name C<MyApp::Private>.
 =item * B<:nopragmas>
 
 By default L<Dancer2> will import three pragmas: L<strict>, L<warnings>,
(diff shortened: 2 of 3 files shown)


xsawyerx:
Any comments, @skington?

veryrusty:
I've always liked the opinionated "use Dancer2" implies "use strict, warnings and utf8" approach.

The Pr looks fine; the "all or nothing" approach ':nopragmas' adds is a simple way to achieve flexibility, but I'm personally not sold on the need.

skington:
I think the code example is in the wrong order. My objection to Dancer importing pragmas was that it clobbers any variation of said pragmas that's already been set up: if e.g. you've said
```perl
use strict;
use warnings;
no warnings 'experimental::smartmatch';
```
then without the ':nopragmas' switch Dancer2 would turn the experimental smartmatch warning back on. So I think the code example should also be in that order.

xsawyerx:
@veryrusty added your phrasing, thank you. :)

xsawyerx:
@skington added per your suggestion.

ribasushi:
This may be tangential, but there is this: https://metacpan.org/pod/Devel::DefaultWarnings. One could use it in an attempt to determine if there is anything to clobber, and act accordingly. @haarg mentioned there are some edge cases he wasn't happy about, but still - this may be an avenue to explore as opposed to the proposed "user needs to do stuff" change.

Cheers

xsawyerx:
I think that catering to the basic user, which Moose has done successfully for a long time, while making it easy for a more advanced user to find what they need in order to override stuff they don't like, is the simplest and most deterministic approach.

I'm afraid of going into the world of "make it import stuff, sort of, but only for warnings, mostly, sort of, depending on what happens".

veryrusty:
Thanks @ribasushi, that's not tangental at all!  A solution that only imported warnings if they were at their 'defaults' would be my preference.

@haarg could you comment on what the edge cases for Devel::DefaultWarnings are?

haarg:
The cases I'm concerned about are things like `use experimental 'signatures';`, which would cause Devel::DefaultWarnings to detect the warnings as non-default.  However, in the future if signatures graduate to stable status, the experimental warning would be removed, and Devel::DefaultWarnings would no longer detect non-default warnings.

xsawyerx:
That might be a time to revise this, but until then, let's continue with this PR.

veryrusty:
@haarg Thanks for the explanation.

veryrusty:
The simplicity of this PR wins.  

Squished and merged as d515925.   :dancers:

Thank you to everyone for their input.

Issue #990, opened 2015-09-04
Deprecated examples on Dancer2::Plugin


ambs:
A simple fix is needed in https://metacpan.org/pod/Dancer2::Plugin to remove that part on 'module compatibility' that is deprecated.

cromedome:
Are you talking about the "Deprecation note" between the "register_plugin" and "plugin_args" sections? If so I will cut that out and submit a pull request shortly.

xsawyerx:
Fixed by #1020. Thanks!

PR #1015, opened 2015-09-29
Removed deprecated functionality from example plugin.

cromedome:
Per https://github.com/PerlDancer/Dancer2/issues/990, register_plugin()
no longer should state what versions of Dancer the plugin is to be
registered for. Also removed the deprecation note - leaving it in would
be potentially confusing for new plugin authors.


xsawyerx:
Replaced with #1020.

PR #1017, opened 2015-10-01
Doc tweaks to make it easier for new users to run Dancer2 apps.

cromedome:
A question was raised on IRC about how to run a Dancer2 app. They cited
the Deployment manual, which instructed users to run "perl bin/app.psgi"
to start a Dancer2 application in standalone mode. After some discussion
with Sawyer, it was determined that not only will this not work, but it
is not a good practice going forward.

My initial patch was going to be to simply replace "perl" with
"plackup", but after reading the entire deployment guide, it seemed
friendlier to novice users to make running standalone the first thing in
the Deployment manual. Since auto-restarting is also a
development-friendly feature, I moved that to second in the list. The
order of items in the remainder of the document has remained unchanged.

I also drew attention to the fact that auto-reloading is not recommended
in production environments.

### DIFF
--- a/lib/Dancer2/Manual/Deployment.pod
+++ b/lib/Dancer2/Manual/Deployment.pod
@@ -12,6 +12,41 @@ __END__
 Dancer has been designed to be flexible, and this flexibility extends to your
 choices when deploying your Dancer app.
 
+=head2 Running stand-alone
+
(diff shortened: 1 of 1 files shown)


xsawyerx:
Merged, thanks! :+1:

PR #1021, opened 2015-10-02
Better explain auto-reloading options for users of Dancer.

cromedome:
I ran into some performance issues using Shotgun due to the number of
modules my application loads and the number of static resources it can
server on a typical request. I found a good middle ground using -r and
-R options to Plack, and I thought other users might benefit from what I
had learned.

This pull request explains how the Plack reloader and the Shotgun works,
and explains what problems users might experience in using them. It's my
hope that users reading this will be able to better decide for
themselves what works best with their app/development process.

### DIFF
--- a/lib/Dancer2/Manual/Deployment.pod
+++ b/lib/Dancer2/Manual/Deployment.pod
@@ -24,21 +24,61 @@ Point your browser at it, and away you go!
 This option can be useful for small personal web apps or internal apps, but if
 you want to make your app available to the world, it probably won't suit you.
 
-=head2 Auto Reloading with plackup and Shotgun
+=head2 Auto Reloading the Application
 
-To edit your files without the need to restart the webserver on each file
-change, simply start your Dancer2 app using L<plackup> and
-L<Plack::Loader::Shotgun>:
+While developing your application, it is often handy to have the server 
+automatically reload your application when changes are made. There are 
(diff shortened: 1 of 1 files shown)


girlwithglasses:
A couple of points you could add:

- use `-M<Module::Name> -M<ModuleTwo::Name>` to preload modules that are not going to change so that Shotgun does not re-evaluate them in every request.

- if you are using Dancer2's new asynchronous capabilities, Shotgun kills Twiggy; I found this out after numerous experiments yesterday trying to get the async demo code working.

cromedome:
Thanks for the suggestions, I will be happy to incorporate them.

Do you happen to know if the other auto-reload options have the same effect on Twiggy?

Jason A. Crome / CromeDome
crome-plated: Web Development and Custom Programming
http://crome-plated.com

cromedome:
Updated the PR. Does that look better?

girlwithglasses:
Looks good to me! Sorry, I didn't do any comprehensive checking of Twiggy, Shotgun, and the other reloader options, but I'd guess that the conflict is Shotgun-specific as Shotgun sets `psgi.streaming` to `0`.

xsawyerx:
Way to go, folks! @cromedome++ @girlwithglasses++ !

xsawyerx:
Merged!

@girlwithglasses, I credited you with your handle, but if you prefer to get recognized under any other name, please let me know!

PR #1024, opened 2015-10-05
Update documentation about public dir

sdeseille:
Update documentation about the fact that static_handler option have to
be set to true when public_dir location is set to something different
from $appdir/public or not declared in $ENV{DANCER_PUBLIC}.

It is a workaround before @cromedome fix that in Dancer2 code ;)


xsawyerx:
Merged, thanks! :+1:

PR #1034, opened 2015-10-16
Change default cookie name from 'dancer.session' to 'dancer2.session'.

cromedome:
I'm not sure this is a bug per se, but since so many other things were
renamed to be Dancer2, the cookie name seemed like an oversight. For the
sake of being consistent, I changed the name.

This discussion prompted the change:
http://paste.perldancer.org/1kccHcbfIsYNX

In this case, there was a Dancer1 and Dancer2 app on the same host using
the default cookie, and inadvertently ended up sharing the same cookies
(not desirable). Racke posted a solid workaround, but this seemed like
something we should also fix.

### DIFF
--- a/lib/Dancer2/Core/Role/SessionFactory.pm
+++ b/lib/Dancer2/Core/Role/SessionFactory.pm
@@ -42,7 +42,7 @@ has log_cb => (
 has cookie_name => (
     is      => 'ro',
     isa     => Str,
-    default => sub {'dancer.session'},
+    default => sub {'dancer2.session'},
 );
 
 has cookie_domain => (
@@ -261,7 +261,7 @@ configuration. See L<Dancer2::Config/Session-engine>.
 
 The name of the cookie to create for storing the session key
 
-Defaults to C<dancer.session>
+Defaults to C<dancer2.session>
 
 =attr cookie_domain
 
--- a/t/error.t
+++ b/t/error.t
@@ -23,7 +23,7 @@ my $env = {
     SERVER_PROTOCOL   => 'HTTP/1.1',
     REMOTE_ADDR       => '127.0.0.1',
     HTTP_COOKIE =>
-      'dancer.session=1234; fbs_102="access_token=xxxxxxxxxx%7Cffffff"',
+      'dancer2.session=1234; fbs_102="access_token=xxxxxxxxxx%7Cffffff"',
     HTTP_X_FORWARDED_FOR => '127.0.0.2',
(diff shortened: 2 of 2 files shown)


xsawyerx:
Agreed here.

xsawyerx:
Merged, thanks! :+1: 
And thanks to pokki and @racke!

racke:
It seems to me that this change would drop existing sessions if you upgrade Dancer2, which would
be a bad surprise.

xsawyerx:
Hmm... tricky. I thought of this more of as a security fix, but you're right. I'll patch it back and release quickly. How do you suggest fixing it?

racke:
Good question, what about setting session_name in the skeleton to the correct value and add a note on startup for existing applications without session_name setting?

xsawyerx:
We could do that and start throwing warnings on the old behavior. The problem is that it will force everyone to change the configuration until that configuration is default.

racke:
Yeah. Maybe just recommending dancer2_session as best practice?

SysPete:
First of all pokki would have had (almost) the same problem if running two independent D2 apps rather than one D1 and on D2 even if the default cookie name was changed. Agree that adding session_name to skel is a good idea but IMO best practice is to name the session cookie something app-specific or completely generic. For a single app domain I generally set it to 'session' and when running multiple apps will use things like 'session-admin' and 'session-shop'.

I agree with @racke that changing the default is now a bad idea. I'll create a PR for skel changes with some text so we have something to discuss.

xsawyerx:
I've released a version that revert this change. Considering we see this change, as this PR presents it, to be the wrong approach, should we close this PR?

SysPete:
@xsawyerx now #1074 exists I think this one can be closed.

racke:
On 12/17/2015 10:25 AM, Sawyer X wrote:
> I've released a version that revert this change. Considering we see this change, as this PR presents it, to be the wrong approach, should we close this PR?
> 

Yes, let's close it.

Regards
	Racke

-- 
Perl and Dancer Development

Visit our Perl::Dancer conference 2015.
More information on https://www.perl.dance.

fgabolde:
> First of all pokki would have had (almost) the same problem if running two independent D2 apps rather than one D1 and on D2 even if the default cookie name was changed.

(I'm pokki.)

"almost" the same problem is right: since D1 and D2 use different session ID formats, D2 (or some D2 session engines? it's been a while) actually throws an exception when it receives a D1 cookie. This is how I first noticed it.

Anyway, thanks everybody for the fix.

PR #955, opened 2015-07-18
views use TT2's INCLUDE_PATH 

veryrusty:
Previously we were usually passing absolute paths for templates (and have ABSOLUTE => 1 set) when rendering output with TemplateToolkit, which worked for most use-cases.

However, as reported in #951 (with further discussion on IRC), if you set a path 'fragment' for a view, such as
```
  set views => 'foobar';
```
The both Dancer2::Template::TemplateToolkit and TT2's code for handling multiple INCLUDE_PATHs were *both* concatenating the views directory with the template name to be rendered. (i.e. looking for a template like `foobar/foobar/index.tt`).

This Pr updates the following to resolve #951
  1. Let TT2 do the template searching, ( override our default methods for `view_pathname` and `layout_pathname` ), and
  2. Sets the default INCLUDE_PATH given to TT2 to be a coderef, allowing the `views` setting to be modified at runtime.

Issue #951, opened 2015-07-16
config key "views" maybe broken with tt?


wchristian:
15-07-15@17:07:09 (Mithaldu) what is the correct way to configure custom template toolkit paths with dancer2?
15-07-15@17:07:35 (Mithaldu) setting views: meep/marp just causes TT to look for meep/marp/meep/marp/my.tt
15-07-15@17:12:38 (Mithaldu) i guess telling TT to INCLUDE_PATH: . works too, if a little inelegant

rleir:
Are you still seeing the problem? (please close the issue if not). In my test, it seems correct to me:

set views => path(dirname(__FILE__), 'meep/marp');

.. lib/Cfoo/meep/marp/hello.tt is not a regular file or reference at /usr/local/share/perl/5.20.1/Dancer2/Core/Role/Template.pm line 128.\n"}

wchristian:
Repro steps:

- osname=MSWin32, osvers=6.3, archname=MSWin32-x86-multi-thread-64int
- uname='Win32 strawberry-perl 5.18.4.1 #1 Thu Oct  2 12:10:29 2014 i386'
- dancer2 0.162000 installed 
- download https://dl.dropboxusercontent.com/u/10190786/micro.zip
- unpack
- perl app.pl
- curl localhost:3000
- error @2015-10-01 21:30:03> Route exception: Failed to render template: file error - rami\my.tt: not found at C:/strawberry/perl/site/lib/Dancer2/Core/Role/Template.pm line 128.
- remove comments on engine config
- perl app.pl
- curl localhost:3000 
- works

veryrusty:
The example by @rleir has an absolute path, whereas the example from @wchristian has a relative path for `views`.  Pr #955 has the details even if the solution there is not the preferred one.

rleir:
@veryrusty Thanks for #955. @wchristian Does this work for you?

wchristian:
The explanation in #951 is fine, though i'll have to leave it up to you, whether you wish to close this based around having one unaccepted pull request.

veryrusty:
Please leave the issue open.

veryrusty:
Fixed by #1037. Thanks @wchristian for raising the issue!

PR #1037, opened 2015-10-19
views do NOT use TT2's INCLUDE_PATH directive

veryrusty:
Previously we were (usually) passing absolute paths for templates (and have ABSOLUTE => 1 set) when rendering output with TemplateToolkit, which worked for most use-cases.

However, as reported in #951 (with further discussion on IRC), if you set a path 'fragment' for a view, such as

```
  set views => 'foobar';
```

The both Dancer2::Template::TemplateToolkit and TT2's code for handling multiple INCLUDE_PATHs were both concatenating the views directory with the template name to be rendered. (i.e. looking for a template like foobar/foobar/index.tt).

Alternate (and simpler) approach to solve #951 (and PR #955) - don't set INCLUDE_PATH and let D2 to the path concatenation.

### DIFF
--- a/lib/Dancer2/Core/Role/Template.pm
+++ b/lib/Dancer2/Core/Role/Template.pm
@@ -6,7 +6,6 @@ use Dancer2::Core::Types;
 use Dancer2::FileUtils qw'path';
 use Carp 'croak';
 
-use Data::Dumper;
 use Moo::Role;
 with 'Dancer2::Core::Role::Engine';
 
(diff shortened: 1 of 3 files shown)


cromedome:
:+1:

xsawyerx:
:+1: 

Merge away.

veryrusty:
Merged as 3224d527f. Thanks all!

bigpresh:
I'm very late to this, but I think this may have been a mistake - I think it would be more user-friendly to let TT2 deal with it.  If I'm not mistaken, this means that all apps that used e.g. `views => 'views'` then expected that process `foo.tt` would load `$appdir/views/foo.tt`, and that a `[% PROCESS 'bar.tt' %]` within `'foo.tt` would load `$appdir/views/bar.tt`, would now need changing, as they wouldn't work after this change?

I'm not sure but I think PR #955 may have been a better solution?

xsawyerx:
@bigpresh can you please open a new issue and reply to #1093 with it?

This issue is closed and no one is monitoring it, I think.

abeverley:
Moving discussion to #1093

PR #1038, opened 2015-10-20
Import plugins with Dancer2::Plugins into application

zurborg:
This feature request allows everyone to use `Dancer2::Plugins` as plugin loader

No big deal, just syntactic sugar.

``` perl
use Dancer2::Plugins qw/
    Adapter
    Ajax
    Auth::Tiny
/;
```

I had to workaround `Import::Into` ... and came to a crazy solution:

``` perl
eval("package $caller { use $plugin } ");
```

Looks weird, is weird, but works good.


xsawyerx:
:+1: Neat idea!

Why did you have to work around `Import::Into`? Couldn't you have you used it directly?

zurborg:
I've done some experiments with Import::Into but that didn't work properly. I think there is an issue with on_plugin_import. So instead fixing this issue I found out that my workaround could help. But the `package $namespace { ... }` syntax seems not known to older perl versions (previous 5.16? uhm...) and that should be fixed, indeed.

xsawyerx:
@zurborg I'll try implementing it with `Import::Into` again. Maybe I'll get lucky.

xsawyerx:
It seems like this cannot be implemented using Import::Into.

The form of `package $name { ... }` is relatively new and we want to remain 5.8.x compatible, so if you could revise that to `package $name; use $plugin_class;` - that would be great.

preaction:
How were you using Import::Into? Your problems may have something to do with https://github.com/PerlDancer/Dancer2/pull/1136

xsawyerx:
I don't remember. I'll try again and report back.

SysPete:
:+1: Since this is sugar that has no impact on any other D2 code I'm in favour of merging this. Also, quite a neat thing to have available when you're importing a pile of plugins.

cromedome:
I like this, thanks! 👍

cromedome:
I cannot get the tests to run for this. Not sure if it is an issue with the feature or the tests. I will investigate.

xsawyerx:
Reviewing it many years later, I'm not a fan anymore, for two reasons:

1. It makes code less grep-able and difficult to statically analyze
2. It's a hack to just avoid writing `use Dancer2::Plugin::` and that seems like a hack that makes code less maintainable for a marginal gain.

I'm sorry for the work you put in, but I suggest we don't merge this.

zurborg:
No problem - it was just a hack, nothing important :-)

PR #1040, opened 2015-10-21
Forward crashes with serializer JSON

basbloemsaat:
By cloning the request into a new request in line 1130 of Dancer2::Core::App a pointer to HTTP::Message::PSGI::$input, a function with a filehandle, becomes stale.
Upon second read no data is returned by the function, but the content length indicates not all was read: an error is generated.

    Route exception: Bad Content-Length: maybe client disconnect?

By deleting the CONTENT_LENGTH header on cloning, this is circumvented.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1126,6 +1126,7 @@ sub make_forward_to {
     my $env = { %{ $request->env } };
 
     $env->{PATH_INFO} = $url;
+    delete($env->{CONTENT_LENGTH});
 
     my $new_request = Dancer2::Core::Request->new( env => $env, body_params => {} );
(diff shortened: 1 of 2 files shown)


veryrusty:
@basbloemsaat++  Nicely diagnosed! 

PR looks good (I'll add a comment to the code before merging). 

Thanks again!

basbloemsaat:
@veryrusty thanks!

veryrusty:
Merged in commit 7399d37. I added comment in Dancer2::Core::App as to why we delete CONTENT_LENGTH and tweaked the commit message. Merged during DancerConf 2015!

Thanks again @basbloemsaat !

xsawyerx:
@basbloemsaat++
@veryrusty++

PR #1042, opened 2015-10-22
Add Digest::SHA to prereqs

veryrusty:
Dancer2::Core::Role::SessionFactory depends on Digest::SHA, which was
added to core in perl v5.9.3. Explicitly add it as a dependency for perl
5.8.x support.


xsawyerx:
Was just about to :+1: it. :)

PR #1044, opened 2015-10-23
added a small utility method: method_is_safe

vanHoesel:
instead of having to check if a request is a safe by checking is_get . . . 

just call method_is_safe

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -311,14 +311,22 @@ sub deserialize {
     return $data;
 }
 
-sub secure    { $_[0]->scheme   eq 'https' }
-sub uri       { $_[0]->request_uri }
-sub is_head   { $_[0]->{method} eq 'HEAD' }
-sub is_post   { $_[0]->{method} eq 'POST' }
-sub is_get    { $_[0]->{method} eq 'GET' }
-sub is_put    { $_[0]->{method} eq 'PUT' }
-sub is_delete { $_[0]->{method} eq 'DELETE' }
-sub is_patch  { $_[0]->{method} eq 'PATCH' }
+sub secure     { $_[0]->scheme   eq 'https' }
+sub uri        { $_[0]->request_uri }
+sub is_head    { $_[0]->{method} eq 'HEAD' }
+sub is_post    { $_[0]->{method} eq 'POST' }
+sub is_get     { $_[0]->{method} eq 'GET' }
+sub is_put     { $_[0]->{method} eq 'PUT' }
+sub is_delete  { $_[0]->{method} eq 'DELETE' }
+sub is_patch   { $_[0]->{method} eq 'PATCH' }
+sub is_options { $_[0]->{method} eq 'OPTIONS' }
+sub is_trace   { $_[0]->{method} eq 'TRACE' }
+
+# RFC 7231 HTTP/1.1 Semantics and Content
+#section 4.2.1 Common Method Properties - Safe Methods
+sub method_is_safe {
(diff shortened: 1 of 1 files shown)


xsawyerx:
You're also adding `is_options` there. Honestly, I don't like these methods. Maybe we should introduce a plugin that adds these comfortable methods. I can hack something up and upload it for reference.

vanHoesel:
it is fine, i needed it mostly in a plugin for conditional request - you can discard the whole thing

vanHoesel:
And yes....

http://tools.ietf.org/html/rfc7231#section-4.3.7 OPTIONS

it's part of the HTTP specs - only a bit difficult to implement - but that might be possible to detect by inspecting the routes.

rleir:
I am curious to know the use case. Option seems to be limited to a very small niche of use cases. When I am designing an API, I intend it to be quite static once I have clobbered the bugs. If I want the API to be discoverable, there seem to be better ways (see the link below). See also Leonard Richardson, "RESTful Web APIs".

The core of Dancer2 should not attempt to cover every RFC. You can build plugins and modules as desired if you want to "scratch your itch". Would you close this issue (when you get a moment, thanks!)

https://www.mnot.net/blog/2012/10/29/NO_OPTIONS

vanHoesel:
@rleir:
I am no discussing the use case of OPTIONS, I was only requesting for a convenience method, which is following the HTTP spec... and explains when a method is safe and when not. It has a 'exact description' of what a safe-method is, one of them is 'OPTIONS'

Dancer2 does support OPTIONS (use case or not) and for all other methods, it does provide the `is_get`, `is_put` and so on, but not for `is_options` - and THAT is what the RFC demands.

Even if the convenience method `methode_is_safe` from the RFC is not something you like to pull in, you still should, for the sake of completeness, pulling in the `is_options` method.
To be honest, if I would ever get into a situation where I would had to check what method I am using (inside a `any` for example) I would be really frustrated if I could do a bunch of if-elsif-else with `is-get`, `is_put` and so on and then get an error because I intuitively used `is_options`. That sucks

rleir:
Now I understand better why you want it. Maybe a plugin?

vanHoesel:
I had to do it myself in my own plugin... for i needed to release my code.

But for the sake off completeness, please please please add the `is_options`

SysPete:
I have to agree with @vanHoesel regarding `is_options` since Dancer2 supports `OPTIONS` and we have the `options` keyword so it makes sense to add `is_options`.

We don't support `TRACE` so `is_trace` is pointless in core. IMHO `method_is_safe` also has no place in core since 'safe' in this case has specific meaning within the context of whatever protocol is being implemented in a plugin.

xsawyerx:
I'm willing to merge the `is_options`.

So I think we are in agreement here:
* Merge `is_options`.
* No merge for `is_trace`.
* No merge for `method_is_safe`.

@vanHoesel Would it be possible to submit the `is_options` separately? If you do, I'll merge that right away.

Meanwhile closing this since it contained multiple changes, only one of which is approved.

vanHoesel:
it's okay...

a shame though that `is_trace` it didn't make it....

haven't tested it, but I will assume that

any '/' => sub { is_trace }

would return 'true' if done with the TRACE method - didn't check - but unless sp[ecifacally blocked, it will simply roll in this


happy coding

xsawyerx:
@vanHoesel Are you submitted the `is_option`?

vanHoesel:
and passed the check with travis... correct, isn't it?

xsawyerx:
I don't follow.

PR #1045, opened 2015-10-23
Small cleanups to the request class

veryrusty:
Two small cleanups:
 * remove an unused var
 * no need to explicitly call `$self->data` within new() as the `body_parameters` method has already done it (one line before).


xsawyerx:
:+1:

veryrusty:
Merged as 09807c2  (fixed typo in commit message)   :dancers:

Issue #1052, opened 2015-11-03
serializer + before_serializer weird behavior


mickeyn:
tried to set the ref to be serialized in the before_serializer hook rather than the route handler, didn't work so well.

here's a minimized version to show the problem:
code: http://pastebin.com/5Gg3CSQS
output: http://pastebin.com/wmcFwbaZ

bug?

xsawyerx:
Nope.

    package BUG;
    use Dancer2;
    set serializer => 'JSON';

    hook before_serializer => sub {
        $_[0] = { blah => 'blah' };
    };

    get '/' => sub {
        print STDERR "do nothing in the route handler\n";
    };

mickeyn:
thanks @xsawyerx ;)

so what happens here is that $content is passed to the hook (undefined value), since @_ is aliasing the same input variable (which is later passed to the serializer) we can set it there.
nice.

xsawyerx:
@mickeyn++ Thanks for adding an explanation of how it works!

mickeyn:
ok, checked and this doesn't really work (@xsawyerx - we only checked a generalized function call outside this problem scope).

problem 1:
since the hook is executed through an intermediate execute_hook which copies @_ into @args and passes *it* down to the hook code, $_[0] is not an alias to the original $content (that's created in 'around serialize').

(minor) problem 2:
there's a check on $content and it's length, so the route handler has to return a value in any case to reach the hook execution.
(that can potentially be solved by moving the content check till after the hook - so I'm not worried about it).

I'm not blocked by this, solved my problem with a completely different approach, but this still bugs me :)

though I agree this is not a bug, it's still an open question -
how can I replace the content sent to the serializer after the route handler has run?
(I do believe the proper place should be the 'before_serializer' hook)

xsawyerx:
@mickeyn Is this fixed by #1053?

xsawyerx:
I guess so. :)

PR #1053, opened 2015-11-05
Allow before_serializer to set content (using the alias in @_)

mickeyn:
To make it even more meaningful, this change also makes sure
before_serializer will always be called when 'serializer' is
set.

The check against the content being empty is done after the
before_serializer hook is called.

I also altered the checks against the content's length.
It's likely that such check will produce warnings in future
Perl versions, as this will check length of stringified refs
when content is a ref (very common).

### DIFF
--- a/lib/Dancer2/Core/Response.pm
+++ b/lib/Dancer2/Core/Response.pm
@@ -107,12 +107,14 @@ has content => (
 around content => sub {
     my ( $orig, $self, $content ) = @_;
 
-    @_ == 2 and return $self->$orig;
+    $self->serializer
+        and $content = $self->serialize($content);
+
+    $content or return $self->$orig;
 
     $self->serializer
         or return $self->$orig( $self->encode_content($content) );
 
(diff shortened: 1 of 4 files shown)


mickeyn:
tied to this discussion: #1052

xsawyerx:
This commit also does other stuff:

* Change a `length` check to a `ref` check.
* Remove a content `length` check entirely.

We talked in person of the latter, but why the former?

mickeyn:
You are referring to another length check (we discussed the one on $content). the change that you mention here is for the passed hook name, checking if it's defined and that it has length is wrong as this check will pass for a reference which is then stringified to be used as a hash key (sure it won't match existing hooks but like in the other 'length' case, relies on ref stringification which is wrong and may lead to warnings in future Perl releases - I had a discussion with Rafael about it and he was surprised it doesn't already).
The suggested check for having a value and it not being a ref seems more sensible to me.

xsawyerx:
Merged, thanks! :+1:

PR #1058, opened 2015-11-13
factor out the slurp function used in two tests

szabgab:


xsawyerx:
This is taken from [Template::Tiny](https://metacpan.org/pod/Template::Tiny) verbatim, so we need to keep it the way it is, unaltered. [03_samples.t](https://metacpan.org/source/ADAMK/Template-Tiny-1.12/t/03_samples.t#L69) and [04_compat.t](https://metacpan.org/source/ADAMK/Template-Tiny-1.12/t/04_compat.t#L65).

Closing. Sorry.

PR #1060, opened 2015-11-15
External Skeleton issues

szabgab:


xsawyerx:
Merged, while supporting `.gitignore`, etc. which the user might explicitly have in their skeleton directory.
Thanks! :+1:

Issue #1062, opened 2015-11-15
dancer2 -s skel -a My::App does not work


szabgab:
```dancer2 -s skel -a My::App``` does not work

I have to put ```-a``` first:

```dancer2  -a My::App  -s skel``` works

IMHO the order of options should not matter.

xsawyerx:
Agreed.

xsawyerx:
#1063 merged, fixes this. Thanks for the report!

PR #1063, opened 2015-11-16
Update backward compat for the -s option

nunorc:
Add the -s option to the backward compatibility conditional in dancer2 script, this should fix issue #1062 . Also add help for this option in the script man.

### DIFF
--- a/script/dancer2
+++ b/script/dancer2
@@ -7,7 +7,7 @@ use warnings;
 use Dancer2::CLI;
 
 # backward compatibility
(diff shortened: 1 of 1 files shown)


xsawyerx:
It's not to *maintain backwards compatibility*, it's the *backwards compatibility conditional*. It's a layer that support compatibility, but seems to practically prevent support for something if it's not mentioned there too.

xsawyerx:
Merged, thanks! :+1:

Issue #1056, opened 2015-11-12
(bug?) should we fall back in body parsing when deserialization fail?


mickeyn:
When we fail to deserialize body content, we fall back to the "regular" body parsing (through Plack::Request) - this may lead to unexpected results (we're reading the content not as intended).

Shouldn't we settle for an error / empty body when we have a serializer set and we fail to deserialize the body?

example:

``` perl
package MyApp;
use Dancer2;
set serializer => 'JSON';
post '/' => sub { +{ xxx => body_parameters->get('xxx') } };
1;
```

with valid JSON:

```
$ curl -q -X POST -d'{"important":"# DANGER!!! NEVER set xxx/yyy/etc. (i.e. yyy=1;xxx=1;...)"}' http://0:3000
{"xxx":null}
```

with invalid JSON (removed the last '}'):

```
$ curl -q -X POST -d'{"important":"# DANGER!!! NEVER set xxx/yyy/etc. (i.e. yyy=1;xxx=1;...)"' http://0:3000
{"xxx":"1"}
```

mickeyn:
forgot to mention this also comes with server log errors for the failed deserialization (as expected).

veryrusty:
@mickeyn++ I'm not keen on the current behaviour and would like to see this cleaned up (or at least simplified). I've been considering (but lacked the tuits to try) reversing the current order, and doing (roughly):
  * If the request is `application/x-www-form-urlencoded` or `multipart/form-data`, then let HTTP::Body or similar do its thing, otherwise
  * let the serializer engine (if defined) do its thing, which includes die on failure.

mickeyn:
@veryrusty: I think it's bad if the server dies on failure to deserialize esp. in the case of invalid user input.

veryrusty:
@mickeyn I meant "let the serialization die".  One day I'll learn to English ;)

I've taken a first pass at implementing what I proposed above. I'll clean it up and push it for review later in the week.

mickeyn:
@veryrusty cool!

xsawyerx:
@veryrusty Is there a PR for that branch anywhere?

veryrusty:
@xsawyerx not yet, I got distracted with seasonal festivities.  I need to go back over it; from memory its the docs that are lacking :(

cwallacetlc:
[...]

PR #1064, opened 2015-11-17
Added DSL keyword request_data.

mickeyn:
`request_data` will return the body of the request in either its
plain (body) form or deserialized (data) form.

This is needed as body_parameters is now a HMV which is suited
to represent parameters and not other forms of data.

Example:
In an app with a JSON serializer, we cannot use body_parameters
to distinguish between a body like '{"data":[]}' and '{}'.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -742,6 +742,15 @@ If you need to access the body of the request, you have to use this accessor and
 should not try to read C<psgi.input> by hand. C<Dancer2::Core::Request>
 already did it for you and kept the raw body untouched in there.
 
+=method body_data
+
+Returns the body of the request in data form, making it possible to distinguish
+between C<body_parameters>, a representation of Web parameters
+(L<Hash::MultiValue>) and other forms of content.
+
+If a serializer is set, this is the deserialized request body. Otherwise this is
+the decoded body parameters (if any), or the body content itself.
+
 =method content
 
 Returns the undecoded byte string POST body.
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -745,7 +745,7 @@ already did it for you and kept the raw body untouched in there.
 =method body_data
 
 Returns the body of the request in data form, making it possible to distinguish
-between C<body_parameters>, a representation of Web parameters
+between C<body_parameters>, a representation of the request parameters
 (L<Hash::MultiValue>) and other forms of content.
 
 If a serializer is set, this is the deserialized request body. Otherwise this is
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -609,6 +609,14 @@ sub _set_route {
 
 sub route { $_[0]->{'route'} }
 
+sub body_data {
+    my $self = shift;
+    return $self->data if $self->serializer;
+    $self->_body_params;
+    return $self->{_body_params} if keys %{ $self->{_body_params} };
+    return $self->body;
+}
+
 1;
 
 __END__
--- a/lib/Dancer2/Core/DSL.pm
+++ b/lib/Dancer2/Core/DSL.pm
@@ -433,7 +433,10 @@ sub route_parameters { shift; $Dancer2::Core::Route::REQUEST->route_parameters(@
 
 sub request_data {
     my $req = $Dancer2::Core::Route::REQUEST;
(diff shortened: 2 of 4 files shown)


veryrusty:
If the request body has been parsed, eg. its encoded as `application/x-www-form-urlencoded` or `multipart/form-data`, then for consistency I'd be expecting those parsed params (that's `request->{_body_params}`,  not the H::MV `body_parameters` )  be returned by this new keyword rather than the raw request body.

Wonder if its better returning something like (pseudo code)
```
( $request->data if request->has_data ) // 
( $request->{_body_params} if keys %{$request->{_body_params} } ) // 
$request->body
```

mickeyn:
@veryrusty++

I don't mind the check, though has_data doesn't exist and I do want to invoke serialization, so I rather keep that check on serializer.
Also, what about invoking the body parsing as well?

like:
```perl
sub request_data {
    my $req = $Dancer2::Core::Route::REQUEST;
    return $req->data if $req->serializer;
    $req->_body_params;                                                                                                              
    return $req->{_body_params} if keys %{ $req->{_body_params} };
    return $req->body;                                                                                                                      
}
```

veryrusty:
We already deserialize the body when the request object is created.

For an app with a serializer defined, when deserialization fails the request body then gets parsed provided HTTP::Body can handle it (`application/x-www-form-urlencoded` and `multipart/form-data` content types, plus some others). Hence the `has_data` check in my pseudo code (and its trivial to add).

[ Linking this to #1056 ]

mickeyn:
yes, that behavior is discussed separately as you mentioned my other issue :)
 I consider the fallback to be dangerous and I don't want to rely on it.

the reason I check for `serializer` is that the `data` method will simply return anyway if it's not set (look at `deserialize`) and I don't wish to further attempt to parse the content in other ways if I was expecting it to be serialized.

cromedome:
What's left to finish with this? Seemed like we were close and then it died on the vine. @veryrusty? @mickeyn?

mickeyn:
sorry for the late reply. I'll have to find the time for it, will try to do so in the coming days.

cromedome:
@mickeyn++

mickeyn:
resolved the conflicts. there are tests and a documentation in place.... shall we merge it?

cromedome:
This is :+1: from me. If @veryrusty is good with it, I will merge this ASAP for the upcoming release.

Thanks @mickeyn!

veryrusty:
I'd like to propose one change: the code for the `request_data` keyword is all based on the request but implemented in the DSL class, it would be cleaner if the dsl method just called a method on the request.

How about we add a new method to the Request class (naming things is hard, it doesn't have to be `body_data`):

```
sub Dancer2::Core::Request::body_data {
    my $self = shift;
    return $self->data if $self->serializer;
    $self->_body_params;
    return $self->{_body_params} if keys %{ $self->{_body_params} };
    return $self->body;
}
```

Then the DSL keyword calls that: `sub request_data { $Dancer2::Core::Route::REQUEST->body_data }`

mickeyn:
@veryrusty++ 

Totally agree, will make the adjustment.

mickeyn:
@veryrusty can you please check again?

veryrusty:
Now that Request->body_data is documented, this is a :+1: from me!    ;)

lib/Dancer2/Core/Request.pm
```diff
@@ -742,6 +742,15 @@ If you need to access the body of the request, you have to use this accessor and
 should not try to read C<psgi.input> by hand. C<Dancer2::Core::Request>
 already did it for you and kept the raw body untouched in there.
 
+=method body_data
+
+Returns the body of the request in data form, making it possible to distinguish
+between C<body_parameters>, a representation of Web parameters
```
racke, line 748: Web parameters sounds weird to me and leaves wondering what they are. These come from a form or as "request" parameters, don't they?

cromedome: I was thinking request parameters might be a better fit here.

mickeyn: I'm happy with any phrasing you prefer

veryrusty: "request parameters" is better phrasing. Docs updated.

racke:
Good to go :+1:

cromedome:
Thanks for your work, everyone! Merged!

mickeyn:
@cromedome++ thanks for the follow up

PR #1070, opened 2015-12-10
404 messages shouldn't display the requested path without escaping

cdmalon:
Currently, a request for a nonexistent path results in a 404 error that displays the requested path, without escaping.

So, if you request

    http://myserver/nonexistent_path<strong>crazy</strong>

you will see "crazy" in bold.  This opens up HTML injection possibilities.  This pull request simply removes the requested path from the information sent to the 404 template.


SysPete:
My view is that including untrusted data in a template without escaping it is a failure of the template / template engine. I personally use a template engine which escapes by default so this issue does not affect me but this patch *does* since I use the message token in my 404.

Time for me to look further...

SysPete:
So this only happens when  `show_errors` is set to true which should never be used in a production environment. My gut feeling is that we should `_html_encode($self->message)` in `Dancer2::Core::Error::default_error_page` anyway.

Thoughts anyone?

veryrusty:
@SysPete I agree with your gut feel.

cromedome:
@SysPete agreed.

SysPete:
SInce I have already submitted PR #1071 to rectify as per my earlier comment I'm going to close this PR.

@cdmalon many thanks for catching this issue.

PR #1071, opened 2015-12-11
_html_encode $self->message in default_error_page

SysPete:
This is an alternative fix for GH PR #1070


xsawyerx:
Merged, thanks! :+1:

PR #1074, opened 2015-12-17
add sample session config to skel - see PR #1034 comments

SysPete:
This is probably too much. Perhaps a smaller section plus improved pod elsewhere would be better. Starting the discussion so we can close #1034 .


xsawyerx:
:+1:

veryrusty:
:+1:

veryrusty:
Merged as 6348aaa48d51e099fb374d5478f2f8a977938c93.

@SysPete++

PR #1075, opened 2015-12-22
fix Template Toolkit INCLUDE_PATH error

mccheung:
Hi,   
looks you forget set Template Toolkit INCLUDE_PATH value. Please merge and update. 

Thanks

### DIFF
--- a/lib/Dancer2/Template/TemplateToolkit.pm
+++ b/lib/Dancer2/Template/TemplateToolkit.pm
@@ -28,6 +28,8 @@ sub _build_engine {
     $tt_config{'END_TAG'} = $stop_tag
       if defined $stop_tag && $stop_tag ne '%]';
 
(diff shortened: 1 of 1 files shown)


veryrusty:
Thanks for the PR, but is there a particular issue or problem you are trying to address @mccheung ?

Typically Dancer2 provides an absolute path to the template to TT2 so `INCLUDE_PATH` is ignored. For relative paths, if `INCLUDE_PATH` is set then both Dancer2 and TT2 will attempt to append path fragments, resulting in templates not being found.

mccheung:
Can you show me the absolute path setting for TT2 ? which file?

veryrusty:
@mccheung Thanks for raising your PR.

Until #1093 was raised, I was missing some extra detail as to what was going on. From the next release, we are now setting `INCLUDE_PATH` again, allowing relative views that include further templates with relative paths.

Issue #818, opened 2014-12-28
Chained hooks


xsawyerx:
I would love to have a hook that provides an _around_, rather than a _before_ or _after_.

The _before_ hook is meant to perform some checks and we're utilizing the **halt** flag after a _before_ hook to check whether the hook asked to stop a request, simply because the _before_ hook wasn't intended for that.

What people really need is an _around_ hook. It will allow you to add information, send the request along (or not), and applying changes on a return, just like a Plack middleware.

Then we could deprecating changing the response (and stopping the request) from the before hook - which I honestly abhor. A lot of our problems have to do with having a response before there's a response, most of it for the _before_ hook.

What do you think?

PR #1077, opened 2015-12-24
Make sure '$@' can be used as an argument to execute_hook

ambs:
In a plugin I had

````
eval { # something }
if ($@) {
  $dsl->execute_hook('jwt_exception' => $@);
};
`````
and with current version, the hook code receives an empty string. But if I do

````
eval { # something }
if ($@) {
  my $x = $@;
  $dsl->execute_hook('jwt_exception' => $x);
};
`````
it will work. The hook will receive $@ contents.

This patches fixes this. But not sure if it will break the functionality added in this commit -- https://github.com/PerlDancer/Dancer2/commit/8e86ce682f2b4699451a3e8c025f7045f6bbe4da

### DIFF
--- a/lib/Dancer2/Core/Role/Hookable.pm
+++ b/lib/Dancer2/Core/Role/Hookable.pm
@@ -96,8 +96,7 @@ sub has_hook {
 
 # Execute the hook at the given position
 sub execute_hook {
-    my $self = shift;
-    my $name = shift;
+    my ($self, $name, @args) = @_;
 
     $name and !ref $name
         or croak "execute_hook needs a hook name";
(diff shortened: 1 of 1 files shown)


xsawyerx:
I'm not sure if this is something we should fix, for two reasons:

1. You should not expect code to keep `$@` unchanged. Since `$@` is not localized by default, you need to save it aside or localize it yourself (although, if you localize it yourself, code that was reached from your code might still change it). Effectively, unless you use `$@` immediately or save it to another variable, you *always* risk action at a distance.

2. There is a major benefit in sending `@_` directly to subs, especially hooks. It allows them to change arguments, since `@_` is aliased. This means we can have **around**-like hooks without having to rewrite anything. This relates to #818.

ambs:
The 2. was my fear, too. as I said in IRC, happy to workaround in the module. Just wanted to be sure this was expected. Thanks!

xsawyerx:
@ambs No worries. :)

PR #1081, opened 2015-12-27
Remove duplicate entry from contributors list

nunorc:
Remove duplicate entry from contributors list.


xsawyerx:
Merged, thanks! :+1: 
(And sorry for only seeing it now...)

PR #1082, opened 2015-12-28
Bugfix/zero response

veryrusty:
The change to allow `before_serializer` hooks in commit 5697359 introduced a side effect for any code "getting" the response content. This reverts the changes to the `around content` modifier made in that commit, resolving #1079 (while keeping @ambs test from #1080).

However this undoes desired functionality from #1053 for a `before_serializer` being able to change the response content in the case when no route returns content for the response. To trigger the serializer hooks in this case, set the response content attribute to the default (an empty string) when constructing the PSGI response (rather than inject the empty string in the PSGI response as done previously).

( I'm keeping @mickeyn to his word for removing that around modifier )

### DIFF
--- a/lib/Dancer2/Core/Response.pm
+++ b/lib/Dancer2/Core/Response.pm
@@ -179,9 +179,12 @@ sub to_psgi {
     Plack::Util::status_with_no_entity_body($status)
         and return [ $status, $self->headers_to_array($headers), [] ];
 
+    my $content = $self->content;
     # It is possible to have no content and/or no content type set
-    # e.g. if all routes 'pass'. Apply defaults here..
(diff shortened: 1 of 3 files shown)


ambs:
:+1: by me. Probably @mickeyn  and @xsawyerx  should comment on this.

Also, sorry for pushing, but it would be great if you could include this patch before the next release.

Happy New Year.

mickeyn:
Fine with me.
Happy New Year.

veryrusty:
Merged as 4969c34712d3034e3a3c76ac4c457ee84c5b9058

Thanks everyone!

xsawyerx:
:+1:

Issue #33, opened 2012-01-26
make uri_for receive a name for a route


tmueller:
This is something for the wishlist, not an issue itself.

Right now uri_for converts a relative path to an absolute path and attaches parameters if necessary.

In case you have to change a path in your controller, you have to search all your templates, to find where this path was used and change your templates accordingly.

My idea would be to provide a name for a route and use that name to identify the route to generate a uri. Something along the lines:

```
# in the controller
get '/product/:id' as 'product' => sub { ... }

# OR
get '/product/:id' called 'product' => sub { ... }


# meanwhile in the template ...
<a href="[% uri_for('product', { id => 5 }) %]">...</a>
<a href="[% uri_for('product', { id => 6, special_price => 1 }) %]">...</a> 

# ... becomes
<a href="/product/5">...</a>
<a href="/product/6?special_price=1">...</a>
```

The Benefit would be, that we can change the route in our controller without having to search our templates for where this route was used. Like so: 

```
# in the controller
get '/product/:id/:special_price?' as 'product' => sub { ... }


# no change in the template is required, but now ...
<a href="[% uri_for('product', { id => 6, special_price => 1 }) %]">...</a>  

# ... would yield
<a href="/product/6/1">...</a>
```

The idea is, to decouple the uris in templates from routing in the controller, so that you are more flexible to change your routes. And to make uri generation more comfortable.

xsawyerx:
I've been playing with this tonight, and given #1088 merging, the [following gist](https://gist.github.com/xsawyerx/e0095c87e0da26196f61) provides a simple implementation of it.

I would be happy to put this in a plugin, or possibly in core, but we would need to resolve two things:

* The new keyword for defining these (went with `named_route` for the example).
* Whether to use `uri_for` or provide a new `uri_for` method. I think the main one can be used if we're adding this to core.

There are two additional cases I'm not sure how and if to handle:

1. `get` creates a `GET` and `HEAD`. The latter is ignored. Correct, I'm assuming. Am I missing something?
2. What should this behave with `any`? Currently it will warn and return empty. Maybe should `croak`?

yanick:
First and foremost, I like the idea. Makes things a little more tidy
then the uris accumulate. :-)

Thoughts and suggestions:

* If you want to incoporate this into core, 
you don't even need a registery of the routes like
in the gist, but can add a 'named_route' attribute
to the route itself.

* You don't really need to filter out the HEAD when it's called
[...]

PR #1088, opened 2016-01-01
Return route objects when defining new ones

xsawyerx:
This commit adds the ability to receive all the route objects
created with the call to the DSL that creates them.

This is helpful if you want to know which objects you just
created. An example could be a plugin that creates a route (or
receives a route created) and wants to inspect it or alter it.

Future thoughts:

* What if we had hooks for route creation?
* What if routes had unique IDs?

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1019,6 +1019,8 @@ sub add_route {
     my $method = $route->method;
 
     push @{ $self->routes->{$method} }, $route;
+
+    return $route;
 }
 
 sub route_exists {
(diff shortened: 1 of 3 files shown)


veryrusty:
We don't specify the return values of most keywords, however returning the route object(s) makes sense here. @xsawyerx++

We should document and add tests before merging though. ;)

racke:
Sounds like a good and useful idea - @xsawyerx ++

xsawyerx:
I'll add some tests and docs. As the ticket now reflects, I want this for what I'm playing with in #33. :)

xsawyerx:
@racke Added a test showcasing it.

I prefer not to add documentation for it because:

1. It's implementation detail.
2. It's experimental.

Can I merge it now? :)

racke:
Good, go ahead!

veryrusty:
:+1:

veryrusty:
Rebased and merged as a73a8b24. @xsawyerx++

:dancers:

PR #1089, opened 2016-01-01
Be more accurate in regexp parsing

xsawyerx:
The current regexp match in a route was checking for a comment
with single characters that compose the words "token", "splat",
and "megasplat". What we want is to match them *exactly*.

I'm assuming the brackets for character classes ("[]") were an
oversight.

Tests all still pass.


veryrusty:
@xsawyerx++

veryrusty:
:+1:

veryrusty:
Merged as 2ed1573ec0d46adc4bf821dcb056ef2b3275385a.

@xsawyerx++ again :dancers:

Issue #1086, opened 2015-12-31
Requirement for encode_base64url() in MIME::Base64 implies version 3.11


wyoung:
This function wasn't added to MIME::Base64 until version 3.11, but the version requirement isn't explicit in Dancer2. On trying to run Dancer2 on my CentOS 5 system, which ships MIME::Base64 version 3.07, you get an obscure "internal server error" complaint in the browser when trying to view even the sample Dancer2 app. ("dancer2 -a app")

xsawyerx:
:+1: Nice discovery!

veryrusty:
Our `dist.ini` has contained a minimum version for MIME::Base64 since Feb 2013:
```
[Prereqs]
...
MIME::Base64 = 3.13
```

@wyoung how did you install Dancer2 on that CentOS5 system ?

wyoung:
@veryrusty: I believe I used cpanm, and it couldn't have been any earlier than December 2013, when I first started evaluating Dancer. I don't have access to the machine to check versions right now, since I'm home for the long weekend, but it was probably more recently than that.

I have a CentOS 5 test VM here, and I installed Dancer2 via cpanm on it. It replaced the platform-default version of MIME::Base64 with 3.15, the current version.

Because the platform ships with an older version, though, this means the problem could be that a later Perl RPM package update would replace the cpanm-provided version. Therefore, I recommend that you reinforce the version requirements in your dist.ini in the `use` statements as well:

    use MIME::Base64 v3.110;

I realize that this is a violation of the DRY principle, but it is a necessary guard.

xsawyerx:
@wyoung No reason not to add it. Would you like to submit a PR? I'll merge it right away.

PR #1091, opened 2016-01-04
Mirrored the version numbers in dist.ini into *.pm

wyoung:
The dist.ini checks are installation time only, whereas the "use" calls
in *.pm are runtime only, so we need both for proper safety.  Otherwise,
a platform Perl module that is older than required may be updated by
cpan/cpanm, then be reverted by a later platform Perl upgrade.  At that
point, Dancer may start doing strange things which are hard to diagnose.
It's better if the Dancer app fails hard when this happens, telling
exactly what broke.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -49,6 +49,15 @@ match = dist.ini
 ;[Test::EOL]
 
 
+; WARNING: When adding or updating a version number here, remember to
+; change the "use" directives in the *.pm files.  This may seem to be
+; a violation of the DRY principle, but these are installation-time
+; checks only.  We also want run-time checks, since the platform Perl
+; may be upgraded after the installation of Dancer, rolling back any
+; CPAN-installed Perl modules to their platform-provided versions.  We
+; want Dancer to stop running if that happens, not bull forward blindly,
+; causing difficult to interpret errors further down the line.  Thus,
+; we oppose one good design principle with another: fail fast.
 [Prereqs]
 Moo = 2.000000
 Moo::Role = 0
--- a/dist.ini
+++ b/dist.ini
@@ -50,14 +50,8 @@ match = dist.ini
 
 
 ; WARNING: When adding or updating a version number here, remember to
-; change the "use" directives in the *.pm files.  This may seem to be
-; a violation of the DRY principle, but these are installation-time
-; checks only.  We also want run-time checks, since the platform Perl
-; may be upgraded after the installation of Dancer, rolling back any
-; CPAN-installed Perl modules to their platform-provided versions.  We
-; want Dancer to stop running if that happens, not bull forward blindly,
-; causing difficult to interpret errors further down the line.  Thus,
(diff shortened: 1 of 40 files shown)


wyoung:
For the most part, this is a mechanical fix for issue #1086 but there are two "creative" bits that one
merging this in should take a look at:

1. Added a big block comment to dist.ini explaining why the version numbers are in 2+ places now.
It basically says the same thing as the checkin comment above.

2. Added a "use Plack v1.3.500" call to Dancer::Core::Runner purely for the version check. I'm not sure
if this is the best place to add it, but no other part of the code says "use Plack", so I made a wild guess.
I tried to avoid it, but neither Plack::Runner nor Plack::Builder (the other main Plack modules that do get
"use"d directly) are versioned, so there's nothing to check against.

I do not have a Dancer2 web app to test this against, but cpanm installed it here, which means all the
tests passed.

xsawyerx:
I thought you were going to submit a PR just for the Email module - the one that you snagged on.

I don't feel comfortable forcing every module to require the specific version needed of every module. This makes development exhausting and removes the entire purpose of a packaging system and a distribution metadata.

wyoung:
As I said in the checkin and dist.ini comments, you need at least two version checks because there are two cases: installation time and run time. Just because the required package versions are correct at installation time doesn't mean they are still correct 6 months later at run time, due to platform Perl upgrades which may in fact *downgrade* a required module.

Maybe you'd be more comfortable if we just had the checks in exactly two places, possibly at the top of Dancer2.pm. This would mean adding "unnecessary" `use` calls, but that can be explained away with a comment.

SysPete:
I have to agree with @xsawyerx here: adding version requirements to every `use` throughout the distribution creates a maintenance nightmare that will come back and bite us on a regular basis. Sorry but :-1: from me on this PR as it is.

wyoung:
@xsawyerx, @SysPete - I've updated my master tip with the versioned `use` calls consolidated into Dancer2::Core, making for a much smaller diff.

xsawyerx:
Unfortunately this is still the same workaround. This time we're forcing the main module to load a bunch of stuff it *shouldn't* just to be able to handle when someone works around the proper packaging.

Let me be more accurate in what I meant in my response in the original ticket:

You encountered a situation with a single module requiring a minimum version. Even though the answer is simply "we package it and in the packaging we provide the minimum version of every requirement we have", we're willing to load a minimal specific version of that module. It's a single change, for a specific module and in order to be accommodating, I'm okay with that.

I don't feel comfortable with either doing this for *every module* in *every class* or forcing the main module to load *every module* necessary. I'm not happy introducing a dependency in the main module on all modules (who would even remember that? Not me...) or providing two different places to require minimum version, just to accommodate people who work around the proper packaging.

I'm sorry, but this doubles our efforts for no major gain and I would be unhappy to incur this additional cognitive load and technical complexity on contributors and core team alike.

Issue #1013, opened 2015-09-29
before_template(_render) hook broken since 0.16200


b10m:
While upgrading to 0.16200 I noticed plugins that use the before_template hook die with:
> [TestApp:23715] error @2015-09-29 18:20:25> Route exception: Can't call method "params" on an undefined value at /usr/local/perlbrew/perls/5.19.3/lib/site_perl/5.19.3/Dancer2/Core/Role/Template.pm line 180. in /usr/local/perlbrew/perls/5.19.3/lib/site_perl/5.19.3/Dancer2/Core/App.pm l. 1273

Example to replicate this:

App (test) code:

```perl
use Plack::Test;
use HTTP::Request::Common;

{
    package TestApp;
    use Dancer2;

    hook before_template => sub {
        # do something, or nothing...
    };

    set template => 'template_toolkit';

    get '/' => sub {
        template 't.tt', { hi => "hello" },
    };
};

my $test = Plack::Test->create( TestApp->to_app );
my $res = $test->request(GET "/");

print $res->content;
```

With a very basic template named ```views/t.tt```

```perl
[% hi %]
```
The problem, I guess, is the order of operations here. When I move the ```set template => 'template_toolkit';``` above the ```hook```, all works as expected, but that'd means I have to use constructs like this to use plugins:

```perl
BEGIN {
    use Dancer2;
    set template => 'template_toolkit';
};
use Dancer2::Plugin::Something::That::Adds::A::Hook;
```

Not ideal.

b10m:
Oh, the real issue comes from the fact that ```$self``` doesn't have a ```->request``` (yet) in ```Dancer2::Core::Role::Template::_prepare_tokens_options```

xsawyerx:
The risk of allowing templates to be rendered outside a request is that you might want to render a template that requires something available in a request (the request object, the request stash, etc.).

I'm not sure why calling the `set template` after the `before` hook causes a crash, though...

xsawyerx:
I see the problem happens because setting the `before` hook clears the variables used later in the template. I'm trying to solve it without having to allow calling `template` outside a request.

xsawyerx:
I have a fix for review in #1092. @b10m++ @SysPete++

b10m:
:+1: Seems to work for me. @xsawyerx++ for the digging (I know how horrible it must've been to debug this ;)

xsawyerx:
[...]

PR #1092, opened 2016-01-04
Do not cache available engines early

xsawyerx:
GH #1013 discovered a problem which can be summarized to: When changing an engine (template, logger, session, serializer) *after* declaring a hook, the engine will not maintain the current request object during request-time.

    package App;
    use Dancer2;
    hook before => sub {1};
    set template => 'simple';
    ...

    # template cannot be rendered because there is no request

What happens is a simple matter of race condition which I'll try to explain.

Normally, when a request is run through the dispatcher, it will instantiate a lazy attribute (`defined_engines`) which lists the available engines for the App. It will then ask them to set
their current active request to the one it received.

That value is then stored in the attribute and is never recalculated.

So far this is just by design.

However, if that value gets calculated ahead of time, by some other action taking place, when the dispatching process reaches that code, it will use the available engines that were created previously. These engines might not be relevant anymore. This is what's happening.

When calling a hook (`hook $name => sub {...}`), the application will call any hook candidates in order to have them add a hook as well. This in turn reaches for the currently available engines, as they are possible hook candidates too. If that value wasn't populated before, it will be now. This means that all the currently-defined engines will be caluclates as the available ones (never to be recalculated).

Enter the race condition manifestation:

When the hook is called, it creates the `defined_engines`, which will include whatever engines are configured as of that moment (such as the template engine). Later the template is reconfigured. However, the `defined_engines` attribute stays the same.

During a request, the Dancer dispatcher tries to set the current request on all available engines, including the original template engine that was available when the hook inflates the value the
dispatcher is using to decide on the available engines.

Thus, the dispatcher sets the old value.

Fixing it:

In order to fix it, we turn the attribute to a subroutine. We call it every time we need to dispatch and keep it for the remainder of the request dispatching. This means that every dispatching now calls this subroutine, which only creates an arrayref of all engine objects, so it's negligible, IMHO.

Additional comments:

* I'm not sure this applies to the serializer engine too. If it does, it will solve it for the serializer engine too.

* Theoretically, all engines could use the request variable `$Dancer2::Core::Route::REQUEST` to reach the current request at any given time, but that would be an implementation detail. Perhaps we should introduce that in the role. That might allow removing an attribute and a few ugly methods for setting the current one. However, I'm not inclined to do that at the moment, since it will incur more depth of knowledge in maintaining the core, and that's never a good thing.


veryrusty:
@xsawyerx++ Looks good. ( But damn; one failing test case :( )
If you need to sleep, I'll try and sort that out!

veryrusty:
:+1:  that was one nasty race condition to chase down.

xsawyerx:
:+1: @veryrusty for always covering my ass. :)

SysPete:
:+1: @xsawyerx great work! I cherry-picked these commits into my local plugins-yanick branch and can now get more plugins working with plugin2. Any chance of getting this into a release soon so I can starting pushing PRs for the affected plugins?

xsawyerx:
Yeah, I can merge this and push a new release today.

I can also rebase the plugins branch to refresh it. (I'd rather do that than cherry-pick or merge.)

veryrusty:
Merged as 56d38c114288c34738227506c6026c986e7b4093.

@xsawyerx++ Thanks for chasing that race condition down!

xsawyerx:
@SysPete Pushed the **plugins-yanick** branch after a rebase. It includes the fix to #1013. You will need to delete your local copy and fetch&&checkout the new one:

    git br -D plugins-yanick
    git fetch
    git checkout -b plugins-yanick origin/plugins-yanick

Issue #1093, opened 2016-01-06
Unable to include relative TT templates from within another template


abeverley:
#1037 removed the default include path for TT templates. Whilst there were good reasons for this, it now means that by default a relative template cannot be included from another template. E.g. any application that was doing ```[% PROCESS another_template.tt %]``` will now be broken.

The simple fix is to define ```INCLUDE_PATH``` in the TT config. This is undesirable though, as it requires an absolute path.

Can I suggest as a solution that a new configuration option is added to the template_toolkit configuration to enable the default include path as ```$self->views``` so as to save hard-coding paths into the config?

If the above sounds okay, I will be happy to supply a PR.

Thanks,

Andy

abeverley:
So, I'm going to try and summarise the issues and potential solutions:

1. Using ```set views => 'myviews'``` with a relative path did not work. This was because ```config->{views}``` was normally configured as an absolute path (built by Dancer on start-up). The config value was passed both to TT for its include path and also used to generate absolute path names passed to TT for each page render. When set to a relative path, Dancer was passing the constructed relative path filename, and TT was searching its (same) include path, leading to duplicate paths (e.g. ```myviews/myviews/index.tt```

2. The recent solution (#1037) fixes the above problem of a relative template directory, but has 2 problems: firstly the original subject of this issue (unable to include files within templates, or needing to include them with a different path, breaking existing applications), but also does not work anyway when an application is deployed using FastCGI (and possibly others?), as FastCGI-deployment means that the app does not find the templates unless the paths are passed as absolute values (I've not been able to work out exactly why this is, but is obviously something to do with working directories when deployed in that manner).

For the reasons outlined in the last part of para 2, I believe that absolute paths should always be passed to TT. Therefore, I can see only 2 solutions:

1. Revert the original patch (#1037) and always use ```set => views``` with an absolute path. The documentation already states this, although the example could be changed from ```set views => path(dirname(__FILE__), 'templates');``` to ```set views => config->{appdir}.'myviews'```

2. Revert the original patch (#1037) and provide a new function to set a relative template path. Of course, the existing ```set``` functionality can't be configured to be "intelligent" as it's a generic function to change configuration.

Incidentally, I think #955 also has problems, as it would still result in relative paths being used for TT, which has problems as described above.

xsawyerx:
@veryrusty You're being assigned to this because you know way more than I do what's going on with this.

@abeverley, excellent write-up!

abeverley:
[...]

PR #1095, opened 2016-01-09
Fix/tt2 relative paths

veryrusty:
Fix for the Dancer2::Template::TemplateToolkit to allow both relative `views` and the further processing of relative templates.

Rough summary: let TT2 use its INCLUDE_PATH searching mechanisms to do any path concatenation and tests for template existence. Without this, any 'relative' templates will not be found when using INCLUDE or PROCESS directives.

Reverts the code change from #1037 (but keep the tests from that PR).

Adds the commits from #955 that build a dynamic INCLUDE_PATH generator for the TT2 object. 

An additional configuration option is added (not in #955), allowing an extra `include_path` (note the lower case) to be prepended to the list of paths added to the INCLUDE_PATH of the TT2 object. By adding in an (absolute) path for `include_path` into, say your production environment's config should mitigate some deployment options not behaving when relative views are used (in development).
Note that this isn't documented - but then again, none of the TemplateToolkit config options seem to be documents (or my search foo is lacking tonight). 

Refactor the test for a view or layout `pathname` existing into a new method. The default is still
a simple `-f $pathname` test.  D2::Template::TemplateToolkit overrides this, delegating the test to the current Template::Context. This allows
  * relative paths for views, and further templates to be processed that are relative to that. (resolving #1093)
  * allows the use Template::Provider classes that pull templates from other data sources (like S3 or a database).

This will also resolve #1075.

### DIFF
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -68,9 +68,9 @@ sub _build_error_template {
 
     # look for a template named after the status number.
     # E.g.: views/404.tt  for a TT template
+    my $engine = $self->app->template_engine;
     return $self->status
-      if -f $self->app->template_engine
-          ->view_pathname( $self->status );
+      if $engine->pathname_exists( $engine->view_pathname( $self->status ) );
 
     return;
 }
--- a/lib/Dancer2/Core/Role/Template.pm
+++ b/lib/Dancer2/Core/Role/Template.pm
@@ -108,6 +108,11 @@ sub layout_pathname {
     );
 }
 
+sub pathname_exists {
+    my ( $self, $pathname ) = @_;
+    return -f $pathname;
+}
+
 sub render_layout {
     my ( $self, $layout, $tokens, $content ) = @_;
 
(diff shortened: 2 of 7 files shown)


xsawyerx:
My head just hurts from trying to fathom it. :)
If it works, it works! :+1: 

@abeverley, could you take a peek?

abeverley:
Thanks @veryrusty, +1 from me.

Tests conducted:

- fastcgi, no custom ```views```, ```PROCESS``` template (my use-case)
- fastcgi, custom relative ```views```, ```include_path``` configured, ```PROCESS``` template
- running locally on port 3000, custom relative ```views```

veryrusty:
@abeverley++ Thanks heaps for doing the extra testing! :beers:

abeverley:
No, thank you @veryrusty!

xsawyerx:
@veryrusty++
@abeverley++

I think we're ready for a merge?

veryrusty:
I think so. Merge away :dancers:

veryrusty:
Merged. Thanks everyone!

PR #1098, opened 2016-01-11
Perl58 fixes - gh1096 & 1097

SysPete:
Perl v5.8 fixes and isa fixes

- skip named capture tests on Perl <v5.10
- stop using Safe::Isa's $_DOES
- some isa checks did not die on failure

### DIFF
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -164,9 +164,12 @@ has serializer => (
     is        => 'ro',
     isa       => Sub::Quote::quote_sub(q{
         use Safe::Isa;
-        $_[0]
-            ? $_[0]->$_DOES('Dancer2::Core::Role::Serializer')
-            : 1;
+        # Perl5.8 not supported for $_DOES (yet) so use $_can + does
+        if ($_[0]) {
+            $_[0]->$_can('does')
+              && $_[0]->does('Dancer2::Core::Role::Serializer')
+              or die "does not have role Dancer2::Core::Role::Serializer";
+        }
     }),
     builder   => '_build_serializer',
 );
--- a/lib/Dancer2/Core/Response.pm
+++ b/lib/Dancer2/Core/Response.pm
@@ -23,18 +23,20 @@ use overload
 has headers => (
     is     => 'ro',
     isa    => Sub::Quote::quote_sub(q{
+        # HTTP::Headers::Fast reports that it isa 'HTTP::Headers',
+        # but there is no actual inheritance.
         use Safe::Isa;
-        $_[0]->$_isa('ARRAY')          ||
-        $_[0]->$_isa('HTTP::Headers') ||
-        $_[0]->$_isa('HTTP::Headers::Fast')
+        $_[0]->$_isa('HTTP::Headers')
+            or die "is not HTTP::Headers or HTTP::Headers::Fast"
     }),
     lazy   => 1,
     coerce => sub {
         my ($value) = @_;
         # HTTP::Headers::Fast reports that it isa 'HTTP::Headers',
         # but there is no actual inheritance.
-        return $value if blessed($value) && $value->isa('HTTP::Headers');
-        HTTP::Headers::Fast->new( @{$value} );
+        $value->$_isa('HTTP::Headers')
+          ? $value
+          : HTTP::Headers::Fast->new(@{$value});
     },
     default => sub {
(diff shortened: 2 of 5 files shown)


xsawyerx:
I suggest making the `$_DOES` possible on 5.10 and above, and revert to `can` on 5.8 and under.

SysPete:
@xsawyerx: I'm still wondering what the special case is that makes DOES useful to us if we are just checking the inheritance hierarchy which isa performs perfectly well. We just want to be sure we have HTTP::Headers(::Fast) or subclass or aref we can use as args?

veryrusty:
One other use of `$_DOES` that needs to be reworked too: https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Error.pm#L168

veryrusty:
`HTTP::Headers` and `HTTP::Headers::Fast` are classes, not roles. The use of `$_isa` in the response class is appropriate (in my opinion).

Though we should keep the use of `$_DOES` in Dancer2::Core::Error (as that is checking for a role).

SysPete:
Now my brain has woken up I just realised there is a bigger problem with both of these isa checks using `->$_DOES`: they don't die.

An isa check that does not die does not do anything other than consume cpu cycles. Let's look at some tests:

```perl
use Test::More tests => 4;
use Test::Exception;

use Dancer2::Core::Error;
use Dancer2::Serializer::JSON;
use JSON;

my $error;

lives_ok { $error = Dancer2::Core::Error->new } "Error->new() lives";

dies_ok { $error = Dancer2::Core::Error->new(show_errors => []) }
"Error->new( show_errors => [] ) dies";

lives_ok {
    $error =
      Dancer2::Core::Error->new(serializer => Dancer2::Serializer::JSON->new)
}
"Error->new(serializer => Dancer2::Serializer::JSON->new) lives";

dies_ok {
    $error = Dancer2::Core::Error->new(serializer => JSON->new)
}
"Error->new(serializer => JSON->new) dies";
```

* Test 1 is just a simple check that we can call `new` with no args.
* Test 2 is checking `isa => Bool,` and passes since [] is not a bool so new dies.
* Test 3 passes a valid serializer arg and passes.
* Test 4 should pass since we know JSON does not have the role 'Dancer2::Core::Role::Serializer' and so the isa check should die.

Commit e73154a equally fails to fix the problem since it should be doing:
```perl
isa => Sub::Quote::quote_sub(q{
    use Safe::Isa;
    $_[0]->$_isa('HTTP::Headers') || die "is not HTTP::Headers or HTTP::Headers::Fast"
}),
```

Please feel free to jump in here and tear this to shreds since it could be my caffeine levels are just too low.

racke:
Looks like the original code as the tests are broken anyway!

veryrusty:
@SysPete++  The Moo docs are clear that the `isa` checks need to `die`for the type validation to fail.

We already have our own type library built on MooX::Types::MooseLike::Base, which allows for
```
isa => AnyOf[ ArrayRef, InstanceOf[ 'HTTP::Headers', 'HTTP::Headers::Fast' ] ],
```

https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Role/Logger.pm#L60-L62
is one more instance of an isa check that does not die on failure.

SysPete:
@veryrusty looks like these isa checks used to use Types::MooseLike but were changed to quote_sub inlined for speed which is where the implicit die got lost. For any hot code paths I'm in favour of using inlined isa checks since the speed boost over Types::MooseLike is significant.

In other projects I've switched to Type::Tiny for speed reasons since it inlines type checks when possible and has a nice XS go faster stripe available too. Perhaps this is something we could consider for the future.

In the meantime will start a hunt for more of these isa bugs. One day maybe I'll find the time to start writing thousands of unit tests.

xsawyerx:
Notice that our type system was doing it wrong and wouldn't catch HTTP::Headers::Fast as HTTP::Headers: 197fad960593a2e5f20cfb951c0a6ff2131accea.

I'm willing to try Type::Tiny. Right now our type checks is what's slowing us down the most.

SysPete:
New issue #1101 created for Type::Tiny discussion.

SysPete:
@xsawyerx: commits fixing all additional issues raised in comments to this PR have been added so I consider this PR ready for merge.

xsawyerx:
@SysPete Excellent!

xsawyerx:
Merged, thanks! :+1:

Issue #1046, opened 2015-10-25
In-config no_server_tokens does not work.


slashbeast:
Hello,

It seems that the no_server_tokens if put in config is completely ignored.

Adding 

    no_server_tokens: 1

to config.yml or environments/*.yml does not work. Dancer still sends Server header with it's version.

As per documentation

    no_server_tokens (boolean)
    If set to true, Dancer2 will not add an "X-Powered-By" header and also append the Dancer2 version to the "Server" header. Default to false - adding.

This seems to be also wrong, as there's no X-Powered-By at all, just the Server header with version.

Headers with config.yml 'no_server_tokens: 1'

    HTTP/1.1 200 OK
    Server: Perl Dancer2 0.161000
    Content-Type: text/html; charset=UTF-8
    Content-Length: 5304
    Date: Sun, 25 Oct 2015 19:38:39 GMT
    Connection: keep-alive

Headers when starting application with DANCER_NO_SERVER_TOKENS=1 env variable:

    HTTP/1.1 200 OK
    Content-Type: text/html; charset=UTF-8
    Content-Length: 5304
    Date: Sun, 25 Oct 2015 19:39:49 GMT
    Connection: keep-alive

Meaning, setting it thru env variable does works.

dev-lang/perl-5.22.0
dev-perl/Dancer2-0.161.0

cromedome:
The code supports this. If you look here:
https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Runner.pm#L99-L101

the config files are never consulted. I'll start looking at this.

slashbeast:
Hello,

Have you managed to do anything regarding this issue? Should the documentation be changed or dancer2 aligned to support that?

xsawyerx:
@cromedome ping. :)

cromedome:
Sorry @slashbeast, $day_job has been in the way of getting this done. For now, perhaps we should patch the docs for now and fix it in an upcoming release. What do you think?

xsawyerx:
This is considered a bug. I'll work on it.

xsawyerx:
Fix provided in #1100.

xsawyerx:
#1100 merged, I consider it fixed, and closing the issue.

@slashbeast Thank you for reporting!

slashbeast:
I totally forgot about this one and just got notifaction that it's fixed.

Many thanks for fixing it!

xsawyerx:
@slashbeast++

:smile:

PR #1100, opened 2016-01-11
Check `no_server_tokens` in config file, fixes GH #1046:

xsawyerx:
GH #1046 raises the issue that the `no_server_tokens` configuration option was not picked up by the application and it would still add the tokens header.

This was caused by checking the Runner for the option (which it does try to set), but ignoring the configuration which is loaded by the config.

Now we're sending the new response object a boolean representing whether to add this special additional header. I don't like it very much for several reasons, but it seems the optimal solution. Said reasons:

1. What if you're using something else to construct the headers? You might not get this one.

2. Since there is no clear way to simply automatically add headers that aren't determined from the response object, we have to create a special attribute for only one specific case, which kind of sucks and is not really scalable.

However, this really is a special case because it's the one extra header we have that has nothing to do with a response, but is determined by the outside world (in this case a configuration option).

Further thoughts:

We used to have Config class (which D1 still has) but it caused more problems since it was a global and applications aren't global and can have different configurations. I'm not sure there's anything to be done to make the configuration more decoupled and cleaner.

The Runner has some defaults without reading the configuration file and the App then reads the configuration file and merges it with the configuration from the Runner as the `settings` method.

1. This means that we keep reading the configuration when someone accesses the `settings` method, which might be correct because then the config can change.

2. The ConfigReader role provides a `settings` attribute which we override. The DSL updates the `config` attribute. This is all pretty weird. That means they can be out of sync and while some components get one, the others might get the other.

Not really sure what to do about this, but not urgent, so meh.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -348,6 +348,7 @@ around _build_config => sub {
 sub _build_response {
     my $self = shift;
     return Dancer2::Core::Response->new(
+        server_tokens => !$self->config->{'no_server_tokens'},
         $self->has_serializer_engine
             ? ( serializer => $self->serializer_engine )
             : (),
--- a/lib/Dancer2/Core/Response.pm
+++ b/lib/Dancer2/Core/Response.pm
@@ -104,6 +104,12 @@ has content => (
     clearer   => 'clear_content',
 );
 
+has server_tokens => (
+    is      => 'ro',
+    isa     => Bool,
+    default => sub {1},
+);
+
 around content => sub {
     my ( $orig, $self ) = ( shift, shift );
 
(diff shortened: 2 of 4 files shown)


xsawyerx:
@PerlDancer, I can't just merge it until I get two approvals. :)

(I mean, I **can**, but I'd rather get approvals :smile:)

SysPete:
:+1: for merging this one since I certainly don't want that config option to get lost. At the same time I think perhaps we should open a new issue since @xsawyerx's description of some of the issues involved in this PR make me a little uncomfortable.

xsawyerx:
@SysPete I don't think any of those are active problems. What issues would you open?

cromedome:
:+1: on this from me too.

All the points you have made @xsawyerx have touched upon things I have run into looking into my rehash of configuration. We've discussed it to some extent on the mailing list but I think there is still more to hash out. 

Not critical, like you said :) But I will revisit my last proposal for config with this in mind. I'd like to get back to finishing this soon.

SysPete:
@xsawyerx I'm really thinking about config rehash as per @cromedome comment as I was wondering if that might change things. Ignore me for now and let's wait and see how it pans out.

xsawyerx:
If anyone wants to open any issues from this, go nuts. :)

Meanwhile, can we merge this?

SysPete:
:+1: Merge away.

xsawyerx:
Merged!

PR #1105, opened 2016-01-16
Correct Try Block & Template Issues

symkat:
We've found routes through auto_page to throw 500 errors.

We tracked these issues down first to the `try {}` block in `Dancer2::Template::TemplateToolkit` missing its semicolon and `use Try::Tiny`.

After correcting this we found the paths from `Dancer2::Handler::AutoPage` need their preceding `/` removed to succeed for the template calls to function.  It is entirely possible a better solution needs to be found for 'Make AutoPage work w/ ::Template::TemplateToolkit' that works for all the templating engines.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -88,6 +88,8 @@ Import::Into = 0
 Safe::Isa = 0
 Hash::Merge::Simple = 0
 App::Cmd::Setup = 0
+; Used in Dancer::Template::TemplateToolkit
+Try::Tiny = 0
 
 [Prereqs / Recommends]
 ; Serializers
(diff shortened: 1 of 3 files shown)


veryrusty:
@symkat++ Thanks for the PR!  (and damn, that was my change; sorry about that. )

We should use an `eval`  instead of introducing Try::Tiny as a dependency, and I need to think over the autopage change.

xsawyerx:
This is great. Try::Tiny is less of a dependency problem (I'm sure something is using it upstream) but a major performance hit.

xsawyerx:
Merged #1108. Thanks!

Issue #1106, opened 2016-01-18
auto_page fails to find Template Toolkit templates


nunorc:
With an application using Template Toolkit engine, and auto_page set, dancer fails to find the templates in the views directory.

    $ dancer2 gen -a Test
    $ cd Test
    (edit config.yml to use Template Toolkit engine, and set 'auto_page: 1')
    $ echo test > views/test.tt
    $ plackup bin/app.psgi
    $ curl localhost:5000/test

This seems to work fine with the default simple template engine.

Dancer2 version 0.166000

nunorc:
@ambs

PR #1108, opened 2016-01-18
Bugfix/autopage template toolkit

veryrusty:
Bugfixs for autopage when using the template_toolkit template engine.

  * Ensure pathname formed from joining of `views` to `request->path` do not have a double slash as a result of that join (See #1106)
  * Remove the use of `try`, replacing it with a simpler `eval`.
  * Better tests for autopage (will test the tiny and template_toolkit engine if available)

This is an alternative to #1105. @symkat++ for that Pr - it made isolating the problem much easier! :dancer:

### DIFF
--- a/lib/Dancer2/Template/TemplateToolkit.pm
+++ b/lib/Dancer2/Template/TemplateToolkit.pm
@@ -76,13 +76,11 @@ sub layout_pathname {
 
 sub pathname_exists {
     my ( $self, $pathname ) = @_;
-    my $rc = 0;
-    try {
+    eval {
         # dies if pathname can not be found via TT2's INCLUDE_PATH search
         $self->engine->service->context->template( $pathname );
-        $rc = 1;
-    }
-    return $rc;
+    };
+    return $@ ? 0 : 1;
 }
 
 1;
--- a/lib/Dancer2/Template/TemplateToolkit.pm
+++ b/lib/Dancer2/Template/TemplateToolkit.pm
@@ -76,11 +76,13 @@ sub layout_pathname {
 
 sub pathname_exists {
     my ( $self, $pathname ) = @_;
-    eval {
+    local $@;
+    my $rc = eval {
         # dies if pathname can not be found via TT2's INCLUDE_PATH search
         $self->engine->service->context->template( $pathname );
+        1;
     };
-    return $@ ? 0 : 1;
+    return $rc;
 }
 
 1;
--- a/lib/Dancer2.pm
+++ b/lib/Dancer2.pm
@@ -286,6 +286,7 @@ things:
     Jonathan Scott Duff
     Julien Fiegehenn
     Julio Fraire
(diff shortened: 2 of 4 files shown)


lib/Dancer2/Template/TemplateToolkit.pm
```diff
@@ -76,13 +76,11 @@ sub layout_pathname {
...
-    my $rc = 0;
-    try {
+    eval {
         # dies if pathname can not be found via TT2's INCLUDE_PATH search
         $self->engine->service->context->template( $pathname );
-        $rc = 1;
-    }
-    return $rc;
+    };
+    return $@ ? 0 : 1;
```
symkat: You might want to localize $@.

symkat:
:+1: Happy to have helped, thank you for getting this fixed up so quickly.

shadowcat-mst:
@veryrusty your use of eval is unsafe; you should be producing a return value from it ala

    my $rc = eval { $self->engine->service->context->template( $pathname ); 1 };

would be ok - if you'd already localised $@. Testing $@ directly the way you're doing is an antipattern that can lead to bugs - which was the exact reason why Try::Tiny was written, and Dancer2 already depends on it via Plack - so unless this particular function shows up at the top of a Devel::NYTProf profile, I think using eval here is premature micro-optimisation with a risk of bugs.

Also, please can you make sure to add to your CONTRIBUTORS section

    Kaitlyn Parkhurst (SYMKAT) <k.parkhurst@shadowcat.co.uk>

shadowcat-mst:
@veryrusty your use of eval is unsafe; you should be producing a return value from it ala

    my $rc = eval { $self->engine->service->context->template( $pathname ); 1 };

would be ok - if you'd already localised $@. Testing $@ directly the way you're doing is an antipattern that can lead to bugs - which was the exact reason why Try::Tiny was written, and Dancer2 already depends on it via Plack - so unless this particular function shows up at the top of a Devel::NYTProf profile, I think using eval here is premature micro-optimisation with a risk of bugs. If not, at the very least add the 'local' as kaitlyn pointed out and switch to the return code based version as I requested, and make a mental note that you must always do both when writing eval statements in future.

Also, please can you make sure to add to your CONTRIBUTORS section

    Kaitlyn Parkhurst (SYMKAT) <k.parkhurst@shadowcat.co.uk>

and to your SPONSORS section

    HMDS laboratory St. James's Institute of Oncology, Leeds via Shadowcat CPAN Support

(my apologies for the customer having a long name ... but OTOH this would be Dancer2 helping to run a cancer lab, so I'm hoping the coolness outweighs the length ;)

veryrusty:
Thanks for the feedback @shadowcat-mst

ambs:
Looks good to me. @xsawyerx, can we get another finger up?

xsawyerx:
:+1: Cutting a release for it!

xsawyerx:
Merged. Thanks everyone!

PR #1110, opened 2016-01-23
Remove logical tests of $@ as eval failures

veryrusty:
Using a logical test of $@ after an `eval` is considered an antipattern.
Many of the uses of `eval` in the core had been reworked to no longer do
that, this commit cleans up the rest.

Also localizes $@ before every `eval`, ensuring we don't replace a previous
error message.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -370,6 +370,7 @@ sub _build_session {
 
         # if we have a session cookie, try to retrieve the session
         if ( defined $session_id ) {
+            local $@;
             eval  { $session = $engine->retrieve( id => $session_id ); 1; }
             or do { $@ and $@ !~ /Unable to retrieve session/
                         and croak "Fail to retrieve session: $@" };
@@ -968,6 +969,7 @@ sub compile_hooks {
                 $Dancer2::Core::Route::RESPONSE->is_halted
                     and return;
 
+                local $@;
                 eval  { $hook->(@_); 1; }
                 or do {
                     $app->cleanup;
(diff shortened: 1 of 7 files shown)


veryrusty:
Hmmm, Travis failed on perl 5.12 and 5.10. Have to go chasing that in the 'morrow.

SysPete:
Testing eval vs. try+catch I'm seeing an additional cost of around 3µs per try+catch on my laptop so 100 uses of try/catch in a single request (seems like a lot) is going to cost only 0.3ms. I wonder if @shadowcat-mst's remarks about premature optimisation are worth considering given what appears to be a very small gain. We obviously want to make Dancer as fast as possible but I think we should perhaps start with reliable code and then try to address these try/catch uses one at a time.

veryrusty:
Thanks for taking a look @SysPete.  I've concur; we need to start with reliable code and optimise use-cases as needed. This PR shouldn't get merged (I'll close it shortly).

On perl versions before 5.14, the boilerplate code needed to correctly deal with `$@` being clobbered when using `eval` is messy (see the Try::Tiny docs for the gory detail).

If Try::Tiny is slow, you'd think there would be other options on CPAN. A quick scan and some benchmarking gave (and there may well be others):
  * `TryCatch` an order of magnitude faster that Try::Tiny. But is XS (plus Moose types).
  * `Try::Catch` is a trimmed down version of `Try::Tiny` and is faster (about half way between Try::Tiny and TryCatch) but is currently perl 5.10+ (but no reason why that should be).

i.e. no other alternatives jumped out and said "use me".

veryrusty:
Closing (as promised).

I'm going to take another look at Try::Catch. It may just be the sweet-spot for reliability vs performance.

PR #1120, opened 2016-01-29
switch from MooX::Types::MooseLike to Type::Tiny

SysPete:
Dumbbench perf script shows around 4% improvement in speed and I guess things will be even better for more complex routes.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -84,7 +84,7 @@ sub with_plugin {
 
 has _factory => (
     is      => 'ro',
-    isa     => Object['Dancer2::Core::Factory'],
+    isa     => InstanceOf['Dancer2::Core::Factory'],
     lazy    => 1,
     default => sub { Dancer2::Core::Factory->new },
 );
--- a/dist.ini
+++ b/dist.ini
@@ -53,7 +53,7 @@ match = dist.ini
 Moo = 2.000000
 Moo::Role = 0
 Role::Tiny = 2.000000
-MooX::Types::MooseLike = 0
+Type::Tiny = 0
 Carp = 0
 Digest::SHA = 0
 Exporter = 5.57
@@ -105,6 +105,7 @@ Test::Builder = 0
 ; Extra speed with Request
 URL::Encode::XS = 0
 CGI::Deurl::XS = 0
+Type::Tiny::XS = 0
 ; Strong session tokens
 Math::Random::ISAAC::XS = 0
 Crypt::URandom = 0
--- a/dist.ini
+++ b/dist.ini
@@ -93,6 +93,7 @@ Import::Into = 0
 Safe::Isa = 0
 Hash::Merge::Simple = 0
 App::Cmd::Setup = 0
+Sub::Quote = 0
 
 [Prereqs / Recommends]
 ; Serializers
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -163,15 +163,7 @@ sub full_message {
 
 has serializer => (
     is        => 'ro',
-    isa       => Sub::Quote::quote_sub(q{
-        use Safe::Isa;
-        # Perl5.8 not supported for $_DOES (yet) so use $_can + does
-        if ($_[0]) {
-            $_[0]->$_can('does')
-              && $_[0]->does('Dancer2::Core::Role::Serializer')
-              or die "does not have role Dancer2::Core::Role::Serializer";
-        }
-    }),
+    isa       => Maybe[ConsumerOf['Dancer2::Core::Role::Serializer']],
     builder   => '_build_serializer',
 );
 
--- a/lib/Dancer2/Core/Response.pm
+++ b/lib/Dancer2/Core/Response.pm
@@ -22,13 +22,7 @@ use overload
 
 has headers => (
     is     => 'ro',
-    isa    => Sub::Quote::quote_sub(q{
-        # HTTP::Headers::Fast reports that it isa 'HTTP::Headers',
-        # but there is no actual inheritance.
-        use Safe::Isa;
-        $_[0]->$_isa('HTTP::Headers')
-            or die "is not HTTP::Headers or HTTP::Headers::Fast"
-    }),
+    isa => InstanceOf['HTTP::Headers'],
     lazy   => 1,
     coerce => sub {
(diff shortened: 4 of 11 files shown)


SysPete:
At the moment all types are exported so would be good to restrict imports in consumers.

lib/Dancer2/Core/App.pm
```diff
@@ -36,7 +36,7 @@ sub supported_engines { [ qw<logger serializer session template> ] }
 
 has _factory => (
     is      => 'ro',
-    isa     => Object['Dancer2::Core::Factory'],
```
SysPete: Type::Standard's Object type throws an exception when used like this since Object doesn't check isa.

SysPete:
There is still room for improvement since some of the Dancer2 types are not inlined but I'm happy with this PR as it stands right now.

What I don't like: Dancer2::Core::Types exports all types which is needed for backwards compatibility especially for session/plugin/... code outside core. Perhaps this could be addressed later with a deprecation warning and eventual switch to exporting nothing by default.

Once switched to Type::Tiny we can start to take advantage of coercions defined within Core::Types rather than having coercion defined in attribute of consuming class. Something to consider for later.

SysPete:
perf.pl improvements I've seen are approx -1.8% for Type::Tiny (no XS) and -2.7% for Type::Tiny::XS

SysPete:
Today I ran 20 runs of perf tests against all three on a different system from previous tests and am still seeing improvements though not so large: -1.11% for Type::Tiny (no XS) and -2.3% for Type::Tiny::XS. So definitely looks like a worthy change on a performance basis.

SysPete:
Full details of most recent perf tests...

MooX::Types::MooseLike	| Type::Tiny PP	| Type::Tiny::XS |     _
--------|--------|---------|-----------
0.72398	| 0.7343	| 0.70167 |
0.7187	| 0.73416	| 0.69776 |
0.7199	| 0.70478	| 0.69788 |
0.72433	| 0.70282	| 0.70248 |
0.71206	| 0.71577	| 0.73264 |
0.73116	| 0.70075	| 0.7299 |
0.75646	| 0.71776	| 0.73168 |
0.71591	| 0.69227	| 0.695183 |
0.71188	| 0.73426	| 0.7086 |
0.71848	| 0.73423	| 0.69454 |
0.71926	| 0.7034	| 0.69914 |
0.7573	| 0.73233	| 0.69656 |
0.71776	| 0.69669	| 0.71317 |
0.70974	| 0.69479	| 0.69755 |
0.72264	| 0.72189	| 0.71073 |
0.72024	| 0.72551	| 0.72262 |
0.71847	| 0.69432	| 0.69107 |
0.71449	| 0.72344	| 0.69518 |
0.715289	| 0.7099	| 0.6933 |
0.70787	| 0.70228	| 0.69158 |
0.72179595	| 0.7137825	| 0.70516165 | average
	| 1.11%	| 2.30% | improvement

shumphrey:
This looks good. For my own sake, all my types are done with Type::Tiny family, so it would be good for D2 to be the same.

bigpresh:
All looks pretty sane to me!  I'm a little concerned that some more helpful exception messages, referring to the type that was required, are now more generic - for e.g. from the tests:

```
-    qr{\Q/.*/\E is not a HashRef},
+    qr{\Q/.*/\E.+did not pass type constraint},
```

However, it shouldn't be too difficult to find out what the constraint is. Some other stuff is clearer, though, and overall it looks cleaner and nicer, and your benchmarks show it to be faster too, which is always nice, so I'm in favour (especially if it also then allows the type-based captures discussed elsewhere!).

SysPete:
@bigpresh `Type::Tiny` exception messages are often overly informative - check this out:

```
perl -e 'use Types::Standard qw/Int/;Int->("qq")'
Value "qq" did not pass type constraint "Int" at -e line 1
    "Int" is a subtype of "Num"
    "Num" is a subtype of "LaxNum"
    Value "qq" did not pass type constraint "LaxNum"
    "LaxNum" is defined as: (defined($_) && !ref($_) && Scalar::Util::looks_like_number($_))
```
so compared to `MooX::Types::MooseLike` exceptions you're really not losing anything.

SysPete:
and there is more as referenced in an earlier comment...

`isa => Object['Dancer2::Core::Factory']` with `MooX::Types::MooseLike` silently actually gives you:

`isa => Object`

whereas `Type::Tiny` will throw an exception in this case. More wins.

racke:
@SysPete already convinced me about Type::Tiny in our Interchange projects, so :+1:

SysPete:
@bigpresh I've updated tests now so expected regexp includes the type name.

bigpresh:
Ah, nice one, thanks for the responses - great to see that the errors are as helpful, or even more so, than they were before!

lib/Dancer2/Core/App.pm
xsawyerx: What do you suggest instead?

xsawyerx:
@SysPete Great work! (as usual...) :+1: from me.

@bigpresh, the errors we will see for the type checks are for **us**, not for the user, and they should:

1. Be rare. They'll happen when **we** screw up, not the user. Our code should provide cleaner errors when a user does something wrong.
2. Be more meaningful to us.

lib/Dancer2/Core/App.pm
SysPete: This has been changed to `InstanceOf['Dancer2::Core::Factory']`

SysPete:
Looks like we have +1 * 2 - anyone have any objections? I'd really like to see this one merged (needed for #1127 for starters).

cromedome:
+1. Do it :)

xsawyerx:
Extra modules are mentioned in `Dancer2::Manual::Migration` and `Dancer2::Manual`. Could you add `Types::Tiny::XS` too?

SysPete:
On 03/04/16 11:41, Sawyer X wrote:
>
> Extra modules are mentioned in |Dancer2::Manual::Migration| and
> |Dancer2::Manual|. Could you add |Types::Tiny::XS| too?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/PerlDancer/Dancer2/pull/1120#issuecomment-204926022>
>
Done.

xsawyerx:
Bring it on!

xsawyerx:
Merged, thanks! :+1:

PR #1069, opened 2015-12-09
New documentation draft:

xsawyerx:
Docs are hard. We currently have the Introduction, the Tutorial,
the Manual, and the Cookbook. It's hard to tell what is documented
where.

The idea was that the Introduction explains things gently,
the Tutorial walks you through writing an application, the Manual
provides the complete coverage, and the Cookbook provides all the
tips, tricks, and hints collected by developers (based heavily
on contributor input).

I think the problem is mainly that we do not have a base-line for
what skills you need in order to understand Dancer and how to use
it. This means we have multiple places to start from, but they
are not clear. Each one tries to cater to multiple different
crowds.

What @mickeyn and I wanted to create was a simple guide that,
assuming you know basic HTTP and Perl, explains the structure of
the Dancer system: the Apps, the routing, the hooks, etc.

Starting from the ground up, this documentation is relatively
thin but is able to not only cover a large and significant portion
of the keywords, it also captures how Dancer works, and how its
structured internally (to an extent).

PR #1123, opened 2016-02-02
document exporting the utf8 pragma

### LINKED ISSUE #1122: Dancer2 messes my encoding
Running the following UTF-8 encoded source file:

    #!/usr/bin/perl -l
    print 'février';
    use Dancer2;
    print 'février';

Produces the following results:

    février
    f�vrier

hexdump reveals that “�” is actually “\xe9”, which is latin1 for “é”.

A grep through Dancer2’s source reveals the responsible line:

    $clean_import
        or $_->import::into($caller) for qw<strict warnings utf8>;

Please don’t do this. Silently messing with your caller’s encoding is never a good idea.

Of course, removing it now risks breaking some callers, so instead I suggest documenting it. Dancer2::Manual currently says:

    strict and warnings pragmas are also imported with Dancer2

It should instead say:

    strict, warnings *and utf8* pragmas are also imported with Dancer2

That would have saved me hours of debugging.

Grimy:
Fixes #1122

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -99,7 +99,8 @@ View the web application at:
 When Dancer2 is imported to a script, that script becomes a webapp, and at
 this point, all the script has to do is declare a list of B<routes>. A
 route handler is composed by an HTTP method, a path pattern and a code
-block. C<strict> and C<warnings> pragmas are also imported with Dancer2.
+block. C<strict>, C<warnings> and C<utf8> pragmas are also imported with
+Dancer2.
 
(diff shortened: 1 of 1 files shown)


veryrusty:
This looks good, matching the documentation for the `:nopragmas` import flag.  :+1:

SysPete:
:+1: Agree: doc fix essential.

veryrusty:
Checked the new docs draft in #1069. The import of the utf8 pragma has been noted there.

veryrusty:
Merged. @Grimy++ Thanks for the Pr!

PR #1126, opened 2016-02-05
Convert named parameters to named captures in routes

smcmurray:
replaces :name parameters in routes to (?\<name>[^/]+) named captures.

Also allows :name(regex) parameters, converting them to (?\<name>regex) named captures.

All named captures are returned as params.
For backwards compatability, all named captures are also returned as captures. But captures now becomes deprecated.

Caveats:

* :name(regex) only allows limited regex that don't contain ')'
* this PR is completely untested and would require close scrutiny before merging.
* this is really intended as a point of discussion more than an actual patch.

### DIFF
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -213,7 +213,7 @@ sub _build_regexp_from_string {
                 and warn q{Named placeholder 'captures' is deprecated};
 
             # Convert :name expressions to (?<name>[^/]+) named captures
-            $string =~ s!(:[^\/\.\?]+)!(?<$1>[^/]+)!g;
+            $string =~ s!:([^\/\.\?]+)!(?<$1>[^/]+)!g;
             $capture = 1;
         }
     }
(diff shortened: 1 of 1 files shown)


SysPete:
I like the idea. Just need to make sure docs (and code) make it clear this is a 5.10+ feature and that 5.8 still does it the old way.

SysPete:
My new typed named parameters (https://github.com/PerlDancer/Dancer2/issues/1127) allows the use of the `StrMatch` type so you can do things like:

```
get '/book/:date[StrMatch[qr{\d\d\d\d-\d\d-\d\d}]]' => sub {
```

This covers the same functionality as your `:name(regexp)` and in addition allows lots of other interesting type checks. I'll add details of the PR to this one as soon as I submit it.

SysPete:
@smcmurray you can see what the PR will mostly include here: https://github.com/SysPete/Dancer2/compare/feature/type_tiny...SysPete:feature/gh1127_route_params_with_types. Check out the tests for an example of regexp type.

xsawyerx:
I think I'm missing a big part here, the **why**. What's the value of this? Are you trying to make something else possible? Are you trying to fix a problem?

SysPete:
@xsawyerx once we have #1120 (switch to Type::Tiny) merged then I'll be pushing a new branch for #1127 (typed route params) which can do all that this PR does and lots more. I guess we can close this one if everyone agrees that #1127 is viable/sensible. @smcmurray are you good with that?

smcmurray:
@SysPete, looking forward to it.

PR #1128, opened 2016-02-07
No warning if optional megasplat not present.

bigpresh:
This is the same bug reported in D1 as PerlDancer/Dancer#1144 - the bug was
also present in D2.

The D1 fix was in commit PerlDancer/Dancer@a5e54dc

### DIFF
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -112,7 +112,7 @@ sub match {
 
             # megasplat values are split on '/'
             if ($token_or_splat[$i] eq 'megasplat') {
-                $values[$i] = [ split '/' => $values[$i] ];
+                $values[$i] = [ split '/' => $values[$i] || '' ];
             }
             push @splat, $values[$i];
         }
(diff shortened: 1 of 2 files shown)


veryrusty:
Looks good. Any chance for adding a test?  (`t/classes/Dancer2-Core-Route/match.t` is probably the best place)

bigpresh:
Right, there's tests for the behaviour of optional megasplats with & without a value - they pass before and after this change.

I was wondering if we should have all test scripts check for the absence of warnings too automatically, but it wouldn't have caught this, as we weren't testing an optional megasplat without a value, it seems.

veryrusty:
I'd normally use `Test::Warnings` to test for such warnings (or the lack there of).  We don't currently have it as a test dependency. Anyone got any thoughts on using `Test::Warnings` ?

xsawyerx:
@veryrusty Don't we use `Capture::Tiny`? Let me take a look at this today.

xsawyerx:
Okay, we *do* use `Capture::Tiny`, which means we can test warnings that way.

xsawyerx:
Oh, and :+1: for the PR. @bigpresh++

veryrusty:
:+1: I'll merge with the Capture::Tiny tests for warnings.

veryrusty:
Merged in 74afc7c. @bigpresh++

bigpresh:
Cheers!

Relequestual:
\o/

Issue #1125, opened 2016-02-03
Current use of Return::MultiLevel silently and unexpectedly breaks program flow


ribasushi:
While on its surface https://github.com/PerlDancer/Dancer2/pull/485 is a good idea, it comes with a huge footgun [as can be seen here](https://github.com/dbsrgits/dbix-class/blob/7cb358520/t/35exception_inaction.t#L74-L86)

The real-life implication of this is that anyone following [the advice in Dancer2::Plugin::LogReport](https://metacpan.org/pod/release/MARKOV/Log-Report-1.12/lib/Dancer2/Plugin/LogReport.pod#Logging-DBIC-database-queries-and-errors) completely breaks their DBIC retry logic and possibly other parts of their model, as the internals are not prepared for longjmp's from the middle of its exception stacks.

While [detecting the problem is relatively easy](https://github.com/dbsrgits/dbix-class/blob/118b2c36/lib/DBIx/Class/Schema.pm#L1067-L1123), protecting a large project from this [is extremely involved](https://github.com/dbsrgits/dbix-class/commit/ddcc02d1) (and expensive performance-wise).

I strongly recommend that the use of `Return::MultiLevel` is re-thought in light of this.

Highlighting @markov2 and @abeverley in case they decide to change `D2::P::LR` ahead of `D2` itself.

ribasushi:
Also highlighting @melo (who had the safer more conservative idea back then https://github.com/PerlDancer/Dancer2/issues/432#issuecomment-24824388) and @mauke (in case he has any thoughts on how `R::ML` itself can improve to aid with this)

xsawyerx:
@ribasushi You missed a later [comment](https://github.com/PerlDancer/Dancer2/issues/432#issuecomment-24824893).

ribasushi:
@xsawyerx Indeed I have. Thoughts on the rest of the ticket?

xsawyerx:
Honestly, I don't think I understand it.
Is it that "Return::MultiLevel causes a return from a situation that would have been handled by DBIx::Class"?

I'm not that fast on my feet today. :/

ribasushi:
*"Flow based on recoverable errors"* is a thing we all use a lot. Let's take an example [from Dancer2 itself](https://metacpan.org/source/XSAWYERX/Dancer2-0.166001/lib/Dancer2/Core/App.pm#L373-374). Now imagine that someone happened to do:
```
[...]

PR #1129, opened 2016-02-09
Call engine.logger.* hooks around logging a message

veryrusty:
The hook names were defined, but they were never called. 

Call them in an around modifier on the log method (similar to where session engine hooks are called). Either being lazy or wise; these hooks are passed the logging engine object, and all params to the log method.

Tests include an example of changing the log level before the message is passed to the logger.

### DIFF
--- a/lib/Dancer2/Core/Role/Logger.pm
+++ b/lib/Dancer2/Core/Role/Logger.pm
@@ -161,6 +161,14 @@ sub _serialize {
     ), @vars;
 }
 
+around 'log' => sub {
+    my ($orig, $self, @args) = @_;
+
+    $self->execute_hook( 'engine.logger.before', $self, @args );
+    $self->$orig( @args );
+    $self->execute_hook( 'engine.logger.after', $self, @args );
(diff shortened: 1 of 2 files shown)


SysPete:
:+1: I'm thinking of some really weird ways to use this - like it a lot.

xsawyerx:
:+1: Nice!

cromedome:
:+1: That's awesome. Like @SysPete, I am anxious to find ways to (ab)use this!

veryrusty:
Merged. Thanks everyone for taking a look!  My first plan for this in untangling D2::P::LR (per #1125).

xsawyerx:
@veryrusty Can you detail some of the plan?

veryrusty:
D2::P::LR does two things
  * puts logged messages into the session => look at doing this in `engine.logger.before` (but can stay where it is), and
  * in some cases redirects to `/` - move this to an `engine.logger.after` hook (which separates concerns between logging and doing something in dancer).

PR #1136, opened 2016-03-01
Fix finding DSL when loading plugin module

preaction:
If we load a plugin via Module::Runtime's use_module, the "caller" of our module isn't the app, but is instead Module::Runtime. To find the right DSL, we need to search up the call stack.

This manifested as a report against Import::Base: https://github.com/preaction/Import-Base/issues/27

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -19,7 +19,6 @@ my $_on_import = {};
 
 sub register {
     my $plugin = caller;
-    my $caller = caller(1);
     my ( $keyword, $code, $options ) = @_;
     $options ||= { is_global => 1 };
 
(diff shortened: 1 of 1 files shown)


xsawyerx:
:+1:

xsawyerx:
@veryrusty, objections?

If not, let's test, merge, and roll. :)

veryrusty:
@xsawyerx This Pr is against D2P1. We've since merged D2P2.

I'm outta tuits at the moment. @SysPete do you have time to see check the test case above against D2P2?

xsawyerx:
Eek I missed that part.

In that case, @preaction, could you please verify if this is still a problem on current **master** branch? The new plugin architecture is in place and that bug might not exist. Sorry for the mess and thanks. :)

preaction:
My test appears to pass now, thanks

xsawyerx:
@preaction Thank you for verifying!

Do you think it would be worthwhile to add a regression test?

SysPete:
Test case passes against D2P2 (current master) so do we add a regression test or just close this PR?

xsawyerx:
I think we can just close it.

PR #1139, opened 2016-03-08
Do not set wrong (character based) content-length in AutoPage handler

pepl:
Also added tests for correct content-length headers setting in case of non-ASCII output


veryrusty:
:+1:

veryrusty:
@pepl++  Thanks for the PR. Always happy to see new contributors to Dancer2!
Merged as 6bdc104. :dancers: 

ps. the PR was missing a change to `t/views/auto_page.tt` to include the `ö` character, and I also tweaked the test case to explicitly match the encoded bytes for a `ö` character.

PR #1140, opened 2016-03-11
Skip rendering of template by the autopage handler on HTTP methods other than GET

racke:
Tiny optimization for autopage handler - it doesn't make sense to render the template and throw it away.

### DIFF
--- a/lib/Dancer2/Handler/AutoPage.pm
+++ b/lib/Dancer2/Handler/AutoPage.pm
@@ -49,8 +49,16 @@ sub code {
             return;
         }
 
-        my $ct = $template->process( $page );
-        return ( $app->request->method eq 'GET' ) ? $ct : '';
+        my $ct;
+
+        if ( $app->request->method eq 'GET' ) {
+            $ct = $template->process( $page );
+        }
+        else {
+            $ct = '';
+        }
+
+        return $ct;
     };
 }
 


veryrusty:
[rfc 1945 - HTTP 1.0 section 8.2](https://tools.ietf.org/html/rfc1945#section-8.2) and (the obsolete) rfc 2616 - HTTP 1.1 state

> The metainformation contained in the HTTP headers in response to a HEAD request should be identical to the information sent in response to a GET request

i.e. A HEAD request _should_ include a Content-Length header.

Whereas [rfc 7231 - HTTP 1.1 section 4.3.2](https://tools.ietf.org/html/rfc7231#section-4.3.2) allows Content-Length header to be missing.


After #1139 was merged, the remaining logic in the AutoPage handler this Pr is optimising does need some cleaning up. However, I think we should always render the content and let our middleware stack remove the body for a HEAD request, so the Content-Length header is included in the response.

veryrusty:
and .. @racke++ for picking up that the logic needs cleaning up there!

xsawyerx:
I think we should benchmark not using the [Head](https://metacpan.org/pod/Plack::Middleware::Head) middleware. If we're already at the spot where we have the body and headers and we can just remove the body, that will save at least two subroutine calls per request for a single `if()` condition. It's much better.

(Some of the optimizations I did in D2 to make it faster than D1 was removing unnecessary middlewares - it made a big difference.)

xsawyerx:
So, so summarize:

* I agree with @veryrusty on the spec. `HEAD` should be `GET` minus the content, meaning we do need to run the `GET`, just not return the content.
* I will open a different ticket to remove the `HEAD` middleware. Middlewares are more expensive than we assume.

xsawyerx:
Done - #1170.

cromedome:
Doing some housekeeping... looking at this PR, the ensuing discussion, and #1170, is it ok to close this? I am inclined to. Thanks!

SysPete:
@cromedome agree: time to close this one.

cromedome:
Thanks!

PR #1141, opened 2016-03-15
Feature: Add `send_as` keyword (from Dancer2::Plugin::SendAs)

veryrusty:
Move functionality from Dancer2::Plugin::SendAs core.
```
  set serializer => 'YAML';
  # send json (not yaml)
  get '/api' => sub { send_as json => [ some => { data => 'structure' } ] };
```

There are often requests asking something similar to "how to disable a serializer for a route". Dancer2::Plugin::SendAs provided a clean solution to that. But its existence is tribal knowledge.
( @xsawyerx and @veryrusty kicked around when the plugin was written if it should be core, but left it as a plugin; we went the wrong way..)

This PR adds `send_as` to core, with some extra features:
  * serializer config will be loaded from the appropriate section of the apps' `engines->{serializer}` configuration.
  * extra options supported by `send_file` can be passed, eg. allowing custom `Content-Type` headers.
  * the apps `charset` setting is used for encoding html before returning to the client.

Enjoy!

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -833,6 +833,41 @@ sub log {
     $logger->$level(@_);
 }
 
+sub send_as {
+    my $self = shift;
+    my ( $type, $data, $options ) = @_;
+    $options ||= {};
+
+    # allow lower cased serializer names
+    my $prefix = 'Dancer2::Serializer::';
+    my $serializer_class;
+    for ( uc $type, $type ) {
+        if ( Class::Load::try_load_class( "$prefix$_" ) ) {
+            $serializer_class = "$prefix$_";
+            $type = $_;
+            last;
+        }
+    }
+
+    my $content;
+    if ( $serializer_class ) {
+        # load any serializer engine config
+        my $engine_options =
+            $self->_get_config_for_engine( serializer => $type, {} ) || {};
+        my $serializer = $serializer_class->new( config => $engine_options );
+        $content = $serializer->serialize( $data );
+        $options->{content_type} ||= $serializer->content_type;
+    }
+    else {
+        # send as HTML
+        my $charset = $self->config->{charset} || "UTF-8";
+        $content = Encode::encode( $charset, $data );
+        $options->{content_type} ||= 'text/html; charset=' . $charset;
+    }
+
+    $self->send_file( \$content, %$options );
+}
+
 sub send_error {
     my $self = shift;
     my ( $message, $status ) = @_;
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -46,7 +46,7 @@ sub _with_plugin {
 
     if ( ref $plugin ) {
         # passing the plugin as an already-created object
-        
+
         # already loaded?
         if( my ( $already ) = grep { ref($plugin) eq ref $_; } @{ $self->plugins } ) {
                 die "trying to load two different objects for plugin ". ref $plugin
@@ -64,10 +64,10 @@ sub _with_plugin {
 
     # check if it's already there
     if( my ( $already ) = grep { $plugin eq ref $_ } @{ $self->plugins } ) {
-        return $already;    
+        return $already;
     }
 
-    push @{ $self->plugins }, 
+    push @{ $self->plugins },
          $plugin = load_class($plugin)->new( app => $self );
 
     return $plugin;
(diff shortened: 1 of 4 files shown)


veryrusty:
Dammmm, force-pushed to s/tab/space in the test case :(

lib/Dancer2/Core/App.pm
```diff
@@ -833,6 +833,41 @@ sub log {
...
 
+sub send_as {
+    my $self = shift;
+    my ( $type, $data, $options ) = @_;
+    $options ||= {};
+
+    # allow lower cased serializer names
+    my $prefix = 'Dancer2::Serializer::';
+    my $serializer_class;
+    for ( uc $type, $type ) {
```
xsawyerx: Is this correct to assume the type is either `$type` or `uc $type`? Would work for YAML and JSON but not for other serializers which aren't uppercased. Maybe I'm not understanding this code.

lib/Dancer2/Core/App.pm
```diff
@@ -833,6 +833,41 @@ sub log {
...
+        if ( Class::Load::try_load_class( "$prefix$_" ) ) {
```
xsawyerx: Can we replace [Class::Load](https://metacpan.org/pod/Class::Load) with [Module::Runtime](https://metacpan.org/pod/Module::Runtime)?

Ooo, can we benchmark them against [Class::Load::XS](https://metacpan.org/pod/Class::Load::XS) to make sure [Module::Runtime](https://metacpan.org/pod/Module::Runtime) is also faster?

xsawyerx:
Otherwise, :+1: from me. About time. :)

lib/Dancer2/Core/App.pm
veryrusty: Class::Load _is_ a (thin) wrapper around Module::Runtime.

lib/Dancer2/Core/App.pm
veryrusty: Your assumption is correct. You can always specify the correct 'casing' for the serializer name.

I wanted to allow `json` or `JSON`. Adding `ucfirst` may be useful which would allow `sereal` or `Sereal` if Dancer2::Serializer::Sereal` existed. Not sure its worth adding further heuristics after that.

lib/Dancer2/Core/App.pm
xsawyerx: Is there a reason to use it? (Other than `Class::Load::XS` possibly making it faster.)

lib/Dancer2/Core/App.pm
xsawyerx: That wouldn't necessarily work either because our capitalization rules allow you to have "SerealButNotThePodcast" using underscores. I would suggest using the capitalization code we have[1](https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core.pm#L7), but that won't work for "JSON".

So now I have two problems:
* I don't want to make a special case for "JSON" and "YAML" which will block the other options. (Does it block other options? Where do you specify the correct casing?)
* I don't want to allow "json" because it's not the right one. "JSON" is the right one.

(Regarding the latter option, one can suggest we support everything in lowercase, but since we need to match it to package names, it becomes more of a mess because you *do* need to know the correct casing for the package to use it, so it will be a mess to handle.)

lib/Dancer2/Core/App.pm
SysPete: `Class::Load::XS` is not used in any significant way by `try_load_class` - most of the lifting in there is via `Module::Runtime`. I've been using `Module::Runtime::use_module` with new projects for a while (wrapped in `Try::Tiny`) mostly for reduced deps but mostly for this:

```
my $obj = use_module('MyClass')->new(%args);
```
Sweet.

SysPete:
:+1: from me too.

lib/Dancer2/Core/App.pm
veryrusty: The core uses Class::Load in several places already, e.g. Factory, MIME, Request, Role::SessionFactory.

xsawyerx: I'm convinced. Shall we open a ticket for deciding on replacing Class::Load with Module::Runtime? We could add a benchmark, have a quick convo, decide, and then either do it or close the ticket.

SysPete: New ticket is good.

xsawyerx: https://github.com/PerlDancer/Dancer2/issues/1145 - done.

lib/Dancer2/Core/App.pm
xsawyerx: Any chance to further address this?

veryrusty:
I'd forgotten this hadn't been merged!  Thanks for the prompt @xsawyerx.

veryrusty:
Reworked and opened new PR #1172 with changes.

PR #1143, opened 2016-03-24
fix utf8 in url

alexsorinpop:
fixed utf8 in urls and added a test for it

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -282,7 +282,7 @@ sub dispatch_path {
     $path =~ s|^/+|/|;
     # PSGI spec notes that '' should be considered '/'
     $path = '/' if $path eq '';
-    return $path;
+    return Encode::decode('utf8', $path);
 }
 
 sub uri_for {
--- a/t/utf8.t
+++ b/t/utf8.t
@@ -1,0 +1,33 @@
+use utf8;
+use strict;
+use warnings;
+
+use Test::More;
+use Plack::Test;
+use HTTP::Request::Common;
+use Encode;
+
+my $utf8 = 'ľščťžýáí';
+
+{
+    package MyApp;
+
+    use Dancer2;
+    use utf8;
+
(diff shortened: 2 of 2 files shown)


veryrusty:
Thanks @alexsorinpop ! Looks ok and matches rfc3986.

This also decodes route params to characters (which is desirable). We should add tests for this to make it explicit. I'm (slowly) working through some other corner cases (eg. when there are undecoded path elements, such as when behind a proxy).

xsawyerx:
:+1: 

This reminds me that we should attempt to use [Unicode::UTF8](https://metacpan.org/pod/Unicode::UTF8) if available. [Path::Tiny](https://metacpan.org/pod/Path::Tiny) does that[1](https://metacpan.org/source/DAGOLDEN/Path-Tiny-0.084/lib/Path/Tiny.pm#L43), [2](https://metacpan.org/source/DAGOLDEN/Path-Tiny-0.084/lib/Path/Tiny.pm#L502), [3](https://metacpan.org/changes/distribution/Path-Tiny#L675). Does anyone want to open a ticket for this enhancement?

alexsorinpop:
So after more digging around it seems that as with every utf8 issue you can't just add that and fix utf8. Go figure.

I found that the STDERR is not opened as utf8 so it might crash the whole app when a utf8 url errors with some syntax in template toolkit and what I found to be really bizarre is that the session is not always remembered... sometimes it will remember your session and the next url that you open will not work anymore. I tested this with ascii urls and I never get that, but I cannot say that I found a way to reproduce this behavior all the time with utf8 urls. I will still dig around

xsawyerx:
Any updates about this?

cromedome:
@alexsorinpop It's been a while :) Any updates to share? Thanks!

xsawyerx:
I suggest we close this and open a ticket that refers to this implementation and discussion to fix the matter.

cromedome:
@xsawyerx seconded. Are you volunteering? ;)

PR #1148, opened 2016-04-02
Rework body param passing to be independent of Plack version.

veryrusty:
The Plack 1.0040 TRIAL release breaks our body param passing; internally
Plack::Request altered the construction of $env->{plack.request.body}.

Calling the original `body_parameters` method from Plack::Request
(via SUPER) ensures this is always set up correctly, no matter what
version of Plack we are subclassing.


xsawyerx:
Why did i not :+1: thing earlier? Ugh.

veryrusty:
Merged as 07ed5dd. Thanks all!

Issue #1149, opened 2016-04-04
Dancer2 loads multiple configs (files and formats)


perlpilot:
If there is both a config.json and a config.yaml  (or any other of the formats that Config::Any understands), Dancer2 will load and merge them all into a single configuration.  While it makes sense to have alternate config files, it makes little sense to load multiple different formats into a single configuration.

See the `_build_config_files` routine at https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Role/ConfigReader.pm#L106

xsawyerx:
Agreed.

cromedome:
I think it would be easy enough to stop looping through possible extensions once we hit on a file/extension combination that works. The problem is, unless we specify an order for the extensions, the file we get might not be the one we expect. And while I don't expect that `Config::Any` will change what it supports anytime soon, we need to be mindful of the fact that if we specify the order of extensions to search, one might get added later that we miss.

I can easily patch things so we stop looking at extensions once we have a match, but what about the order of them? @perlpilot? @xsawyerx? Anyone else? :grinning:

xsawyerx:
Basic order from `Config::Any` is good enough. Having a warning when we have multiple files (so a user understands why it read the YAML file but not the JSON file) would also be useful.

@cromedome also mentioned having an environment variable available to control that, a la:

    DANCER_CONFIGFILE="config.yml"

or

    DANCER_CONFIGFILE_EXT="json"

In which case I think the warning for multiple files should not be printed, because you're explicitly narrowing the field.

veryrusty:
I view this as a poorly documented feature than a bug.

I've had one use-case where I wanted a hand written and commented config augmented with configuration generated from an orchestration tool. Having `config.yaml` and `config.json` made that easy.  TIMTOWTDI applies of course, but the dual config files was a simple solution.

xsawyerx:
The alternative would be to document this as intended behavior.

Honestly, you raise a good use case and I could swing either way.

perlpilot:
Problems I see with this "feature" as currently implemented:
* Future developers will be wondering why there's both a `config.json` and a `config.yaml` causing potential confusion.
* Not everyone has the luxury of using multiple config file formats
* For people not using multiple config file formats, they will have a `config.yaml` and a `config.yml` which, to my mind looks even more hacky (Why two YAML config files instead of just one?)
[...]

Issue #47, opened 2012-07-01
Add support for additional configuration files


dsheroh:
Applications often require local configuration settings (e.g., database connection info, support email address, etc.) which should neither propagate back into the master copy of the app nor preclude updating the local copy when the master adds new config settings.  To allow for this, Dancer should support a configuration loading order along the lines of:

config.yml, config.local,yml, [environment].yml, [environment].local.yml

In each case, the .local version would be merged into the overall configuration, being merged in essentially the same way that [environment].yml is currently overlaid onto config.yml.

Ideally, this could be generalized even further, allowing users to insert an arbitrary number of additional configuration files at any point in the config loading order, thus allowing configuration of a large number of plugins (both general Dancer plugins and app-specific plugins) without having to resort to putting everything into a single monolithic configuration.

rleir:
Can it be closed now?

cromedome:
Resolved by e9a633d0851f42fa5822151920180a40887a8dbc. Closing!

PR #1150, opened 2016-04-04
configuration bugfix + enhancement

perlpilot:
I've tried to keep the commits atomic so that they can be applied or not individually.  The first three commits are essentially for #1149.  To wit: 

* Once we find a config file in each of the main dir and in environments, don't use any others. (i.e. don't use both `config.ini` *and* `config.json` simultaneously if both files happen to be there.)
* Generate warnings if the user happens to have multiple config files in the main dir (e.g. both `config.json` and `config.yaml`) because in that situation, how is the user to know which was used?
* Let the user specify via an environment variable the configuration extension and only look for files with that extension.  However, there is no checking that this extension is a valid type that Config::Any understands.
* what started me looking at the config generation in the first place: at $work we have some developer configuration information that is common across all developers, but we also have some config info that's specific to each developer.  This commit allows for files such as `environments/my-development.yaml` for such cases.

### DIFF
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -136,6 +136,8 @@ sub _build_config_files {
     for my $conf (keys %multiples) {
         warn "Used config file $conf, but also found and did NOT use these config files: @{$multiples{$conf}}";
     }
+
+    push @files, $ENV{DANCER_ALT_CONFIG}  if exists $ENV{DANCER_ALT_CONFIG};
     return \@files;
 }
 
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -114,15 +114,24 @@ sub _build_config_files {
     my @exts = Config::Any->extensions;
 
     my @files;
+    my %multiples;
     for my $config ( [ $location, "config" ], 
                       [ $self->environments_location, $self->environment ] ) {
         my ($location, $filename) = @{$config};
+        my $found_config = 0;
         for my $ext (@exts) {
             my $path = path( $location, "$filename.$ext" );
             next if !-r $path;
-            push @files, $path;
-            last;
+            if ($found_config) {
+                push @{$multiples{$files[-1]}}, $path;
+            } else {
+                push @files, $path;
+                $found_config = 1;
+            }
         }
+    }
+    for my $conf (keys %multiples) {
+        warn "Used config file $conf, but also found and did NOT use these config files: @{$multiples{$conf}}";
     }
     return \@files;
 }
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -118,7 +118,8 @@ sub _build_config_files {
     my @files;
     my %multiples;
     for my $config ( [ $location, "config" ], 
-                      [ $self->environments_location, $self->environment ] ) {
+                      [ $self->environments_location, $self->environment ],
+                      [ $self->environments_location, "my-" . $self->environment ] ) {
         my ($location, $filename) = @{$config};
         my $found_config = 0;
         for my $ext (@exts) {
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -111,22 +111,20 @@ sub _build_config_files {
     # an undef location means no config files for the caller
     return [] unless defined $location;
 
-    my $running_env = $self->environment;
     my @exts = Config::Any->extensions;
+
     my @files;
-
-    foreach my $ext (@exts) {
-        foreach my $file ( [ $location, "config.$ext" ],
-            [ $self->environments_location, "$running_env.$ext" ] )
-        {
-            my $path = path( @{$file} );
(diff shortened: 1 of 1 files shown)


perlpilot:
Any love for this PR?

cromedome:
My bad. Started working with it, haven't had the time to finish. Sorry!

mrmaloof:
I love it. We could also really use developer specific configuration.

veryrusty:
As this PR is proposed, it will break existing deploys by limiting the configuration load to the first found extension.

There are some good ideas here that should be taken further:
  * Being more explicit about what config files are loaded. eg. logging each file as its read/merged.
  * Using an env var to limit the extensions, like `$ENV{DANCER_CONFIGFILE_EXT}`, and
  * A prefix/suffix for local overrides. My preference is `_local` as a suffix (see #47 too).
 
The `$ENV{DANCER_ALT_CONFIG}` is seems somewhat conflated. You could also point it at a file that Config::Any can't handle.  Another (hopefully simpler) approach is to have the prefix/suffix option on both the `config` and the `environment` file.

Once the above is in place, then there is scope look at only loading configuration from the first found extension, and deprecate existing functionality as necessary.

SysPete:
Rather than prefix/suffix for local overrides I'd much prefer to see a `config.d` directory as suggested in #47. This could even be implemented via a config key in config.yml (or whatever) such as:
```
config_dir: config.d
```
and also allow a list:
```
config_dir:
    - config.d
    - local.d
```
Otherwise agree 100% with @veryrusty and I don't want to see changes that would break existing projects.

racke:
Yes, I second @SysPete. Directory works better.

perlpilot:
@veryrusty Yes, this will break existing deployments that rely on this "feature" of how config files are loaded.  How many do you think that is?   I'm all for backwards compatibility, but not for a "feature" that seems more accidental that purposeful.

Below, I've quoted the stuff I blathered on IRC the other day for anyone who would care to comment.

> \<perlpilot> veryrusty: why exactly?  what's the use-case for having both config.yml and config.json?
\<perlpilot> veryrusty: also, you realize that D2 will currently load config.json first, then config.yml and any settings in config.yml over-ride those in config.json and there's no way for the user to modify this.
\<perlpilot> veryrusty: moreover, the order those files are loaded is entirely determined by the order that Config::Any->extensions provides the extensions (which is not documented).
\<veryrusty> I've a set of D2 apps deployed via as (docker) containers. the containers include a set of "static" config, as yaml files. secrets are written to json by orchestration tools and mounted into the containers.
\<veryrusty> TIMTOWTDI applies, but that "just works".
\<perlpilot> veryrusty: until someone accidentally changes the order that Config::Any->extensions returns things or someone accidentally (or purposefully) names a key in your YAML file the same as one of the keys in your JSON file (the YAML file will take precedence)
\<perlpilot> veryrusty: Would it be onerous for you to change the JSON files to by my-production.json ?
\<perlpilot> er, environments/my-production.json
\<veryrusty> only that I have to change something that currently works
\<veryrusty> ( I'd also prefer a '_local' suffix than the 'my-' prefix )
\<perlpilot> aye, we could bikeshed the name all day  :)   I don't care one way or the other. 
\<perlpilot> I'm just after some consistency and a little more control.
\<perlpilot> for instance, we use YAML files for our config, so currently the *only* way to have some creds in a config file is to use something other than YAML  (or, use config.yaml and config.yml)
\<perlpilot> which just seems wrong
\<perlpilot> (and again, we can't know for sure without looking under the hood a little bit which order they will be loaded)
\<perlpilot> I'm fine looking under the hood and understanding whatever trade-offs there are, but then I have to document that for the rest of my team so they don't accidentally mess something up (or if they do, at least have a clue about what went wrong)
\<perlpilot> but then, it's also not just *my* team that has to do this, but everyone who uses Dancer2
\<perlpilot> (that is, re-discover how to store local config info and that they might have to do mix-n-match two different config file formats)

I'd prefer that Dancer have some clean, documented way to achieve the 2 main things I'm after rather than have everyone rediscover how to do it:
* deployment-specific configuration
* developer-specific configuration

cromedome:
I tend to agree with @sukria on this point. We are adding too much complexity to the app by having the configuration dir, and I am not a fan of implementing it as such.

I agree with @perlpilot that the way it works now is an undocumented "feature", and if something changes in the way `Config::Any` works can potentially change the way configuration works in Dancer, causing apps to break due to no fault of our own. For that reason alone, we should tighten up the way we are processing config files.

I can see @veryrusty's point about not wanting to break existing apps, and @perlpilot's needs as well. I think there's a middle ground to be had between the two that doesn't involve moving to a configuration directory. I'll see what I can cook up tonight - a bit tied up for most of the day today.

Thanks for the patch, @perlpilot, and for getting the ball rolling on this.

perlpilot:
@veryrusty I was just thinking about your comment to log when each file is merged.  If it told you what config items were merged and which took precedence, that would be quite useful.  Something like "Replacing config.yaml 'service_A_info' with value from environments/development.yaml".  Especially if the group does decide to use some sort of config.d directory.  

BTW, I think having a config.d directory makes for too many moving parts when config files are merged as they are currently. And performance could be negatively impacted if there were many files in this directory.  I'm more in favor of some relatively small number of well-defined places for Dancer to look for configuration information.

cromedome:
Resolved by e9a633d0851f42fa5822151920180a40887a8dbc. Closing!

PR #1151, opened 2016-04-05
update pod noting that plugin's config is immutable after 1st read

SysPete:
Perhaps I've got the wrong end of the stick here but here is a simplified view of plugin2's config attribute from my view of the world. If this is correct then this needs adding since a number of plugin maintainers are used to altering config on the fly at runtime.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -854,6 +854,29 @@ For example:
         from_config => sub { 'See ya!' },
     );
 
+=head3 Config becomes immutable
+
+The plugin's C<config> attribute is loaded lazily on the first call to
+C<config>. After this first call C<config> becomes immutable so you cannot
+do the following in a test:
+
+    use Dancer2;
+    use Dancer2::Plugin::FooBar;
+
+    set plugins => {
+        FooBar => {
+            wibble => 1,  # this is OK
+        },
+    };
+
+    flibble(45);          # plugin keyword called which causes config read
+    
+    set plugins => {
+        FooBar => {
+            wibble => 0,  # this will NOT change plugin config
+        },
+    };
+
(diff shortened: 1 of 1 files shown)


xsawyerx:
I'm not sure that the best phrasing, since that's not exactly correct, and that's not the exact thing that is failing there.

What happens (in a technical way that is not the way to phrase it in the docs) is that all the engines are built once you call `to_app`. Engines that have triggers will get rebuilt by those when the settings are applied using `set`.

What I think we should stress here is that serializers (at the very least) should not be set after you called `to_app` (which is what happened in the original ticket) because they're part of how you perceive routes to begin with. It's what gets run **before** a `before` hook, not after.

SysPete:
I get the engines and serializers stuff - that is clear to me already: this is specific to plugin2 config.
so...
- plugin2 config attribute is immutable and lazy
- there is no writer or clearer
so `set plugins => { ... }` *after* config attr has been read does nothing unless something somewhere is doing something really nasty that I don't know about.

@xsawyerx: so my question is really...
Is it correct (and expected) that trying to set plugin config *after* something calls $plugin->config does nothing? The PR text is not very good - wrote it rather too quickly but I feel something is needed in Plugin pod about this since many plugin authors have got used to being able to change config at runtime whenever they like

xsawyerx:
Revisiting this again, I think I might have misunderstood what this documentation was trying to clarify, and that it is correct.

What I said is correct, but it was against engines, not plugins. Your change is against plugins. I think that setting plugins with `set` does (re)?instantiate them. We need to double check this.

cromedome:
:+1:

xsawyerx:
:+1:

cromedome:
Merged in 75a01c447d7c2e6824159b6b6cdee78298be5fc2. Closing!

PR #1157, opened 2016-04-13
Bugfix/inconsistent megasplat split

veryrusty:
The string that a megasplat matched is split on `/`. Without setting an LIMIT for the split, empty trailing values get stripped. That is; for the route `/foo/**`, the paths `/foo/bar` and `/foo/bar/` were both matching and gave the splat as `['bar']`.

Or, if there were further trailing slashes in the path:
`/foo/bar///a`  gave  `['bar','','','a']` but
`/foo/bar///`   gave  `['bar']`.   How's that for inconsistent.

Adding a (negative) limit to the split on megasplat values so empty trailing values are kept, which allows you to join the arrayref of values together and get the original path again.

The above examples become:
`/foo/bar/`     giving `['bar','']`

`/foo/bar///a`  giving `['bar','','','a']`  #as before

`/foo/bar///`   giving `['bar','','', '']`

This is a subtle change in behaviour, which _may_ cause some Dancers issues if they are not aware of the change.

### DIFF
--- a/Changes
+++ b/Changes
@@ -12,6 +12,13 @@
     * GH #1144: Change tt tags to span in skel (Jason Lewis)
     * GH #1046: "no_server_tokens" configuration option doesn't work.
       (Sawyer X)
+    # GH #1155, #1157: Fix megasplat value spliting when there are empty
+      trailing path segments. (Tatsuhiko Miyagawa, Russell Jenkins)
+      NOTE: Paths matching a megasplat that end with a '/' will now include
+      an empty string as the last value. For the route pattern '/foo/**',
+      the path '/foo/bar', the megasplat gives ['bar'], whereas '/foo/bar/'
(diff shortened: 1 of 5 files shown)


veryrusty:
(Updated commit comment - examples need to be correct!)

xsawyerx:
:+1: With a nice reflection of this in the Changes file.

racke:
:+1: I agree with @xsawyerx

veryrusty:
I'll merge this soon :) Thanks all (and especially to @miyagawa)

veryrusty:
@xsawyerx @racke I've added (a longish) changelog entry for this as a separate commit (edcbccc). Is that what you had in mind ?

racke:
Yeah, I'm just think it is "splitting" instead of "spliting".

veryrusty:
Thanks @racke. Fixed and merged.  :dancers:

PR #1158, opened 2016-04-14
Pod update: List apache modules with apachectl -M

Varadinsky:
Hi,

Just started tinkering with Dancer2 :-)

apache -l will only list modules that are compiled into the server. It will not list shared modules.

I believe it is better to use apachectl -M in this example since it lists both static and shared modules.

Thanks,

Zoran


xsawyerx:
Merged, thanks! :+1:

PR #1162, opened 2016-04-26
Pr/change skel template tag

jasonblewis:
The tags used in the D2 skeleton app are `[% %]`.

This is a problem if you want to make your own skeleton for generating applications that use the TT2 `[% %]` style tags as they incorrectly get consumed by the generating process.

This changes the skeleleton's template tags to `[d2% %2d]` which will be less
likely to clash with tags in templating systems.

Note: this is only the tags used for generating the application. ie `dancer2 gen -a MyWeb::App`

### DIFF
--- a/lib/Dancer2/CLI/Command/gen.pm
+++ b/lib/Dancer2/CLI/Command/gen.pm
@@ -216,8 +216,8 @@ sub _add_to_manifest_skip {
 sub _process_template {
     my ($template, $tokens) = @_;
     my $engine = Dancer2::Template::Simple->new;
-    $engine->{start_tag} = '[%';
-    $engine->{stop_tag} = '%]';
+    $engine->{start_tag} = '[d2%';
+    $engine->{stop_tag} = '%2d]';
     return $engine->render(\$template, $tokens);
 }
 
--- a/share/skel/Makefile.PL
+++ b/share/skel/Makefile.PL
@@ -8,9 +8,9 @@ my $eumm_version = $ExtUtils::MakeMaker::VERSION;
 $eumm_version =~ s/_//;
 
 WriteMakefile(
-    NAME                => '[% appname %]',
+    NAME                => '[d2% appname %2d]',
     AUTHOR              => q{YOUR NAME <youremail@example.com>},
-    VERSION_FROM        => '[% appfile %]',
+    VERSION_FROM        => '[d2% appfile %2d]',
     ABSTRACT            => 'YOUR APPLICATION ABSTRACT',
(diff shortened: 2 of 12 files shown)


jasonblewis:
Rake suggested `[💃 💃]` tags

jasonblewis:
further irc discussion:
22:17:13       veryrusty | rather than hard-coding some set of tags for the skel; how about the skel directory contain the tags to use in a config file ? (defaulting to [% %] if nothing is set)
22:18:45           k-man | so the config would be in the skel directory?
22:20:18       veryrusty | yep   _skel.json   (or some such)
22:20:33           racke | or regex right away :-)
22:22:42           racke | tag-regex: '\[%(.*?)%\]'

ambs:
I confess I do not like `[d2% %2d]`. It is big, easy to err (not sure mirroring d and 2 is good).
Why can't we go less creative, but still enough for most templating systems? `[$`, `[#`, `[!` etc...?

I know this is not to be used everyday, but ;-/
well, just my 2 cents.

cromedome:
I am with ambs on this. I am a fan of the PR in general... but not of the tags selected. [$ $] (or any of the other suggestions) are good with me.

jasonblewis:
I'm happy to change the tag to something else if that's the direction you want to go, but maybe a better approach is veryrusty's suggestion of making it a configuration option via a file in the /skel directory?

One problem I see with that approach is that it's another magic place to put things that alters behaviour. Maybe a command line option to the d2 app generator for a user customisable tag is better?

ambs:
agreed on not having yet another config file, at least, while we do not need it (realy need it).
I would go the simple way: define a tag different from the ones usually used in dancer apps ([% and <%).
Then, in the future, if by some reason that is needed, think on further customization

At least this would be my approach.
Cheers

cromedome:
This came up in conversation the other day, so I came back to this issue to re-review it. While I don't like the `d2` tags myself, it's really an issue with my personal preference. I don't see a reason not to merge this. It's not like people are making skeletons for Dancer apps every day, and this solves a problem that some of our users are having. I am for merging - so :+1: from me. Thanks, @jasonblewis!

racke:
Yeah let's go with that before it turns into a historical monument :-)
:+1: for merging.

cromedome:
Merged. Thanks so much!

jasonblewis:
Something that I didn't think of is that it breaks compatibility with everyone's skeletons they already wrote. for instance @szabgab's https://github.com/szabgab/Dancer2-Angular-Skeleton.

and now I have cold feet about this approach

Issue #1159, opened 2016-04-15
'template' keyword not available via `use Dancer2`


girlwithglasses:
From the Dancer2 advent calendar article ['Command line dancing'](http://advent.perldancer.org/2014/16)

> If you want to create command line applications that use the Dancer2 structures, you can simply `use Dancer2` and use the keywords it provides you.
> 
> What does this give you access to?
> 
> Dancer keywords
> `config` probably being the most helpful one, but `template` will also be available just like other engine-related keywords are - explained below under Engines.
> 

This is no longer true for Dancer2-0.166001, as `template` is only available from a route handler. Is there any chance of making it globally available? It would be very handy to be able to use it for writing files, etc. If not, could the advent calendar article be updated?

xsawyerx:
The only reason I can think of that the `template` keyword would only be part of a route handler is because it contains a reference to the request as a token. We can change that, obviously. If I change it, none of the tests fail, which is pretty good.

I agree it can be a useful feature.

Does anyone have an objection? (@PerlDancer)

veryrusty:
We'd need to wrap the parts of `Dancer2::Core::Role::Template->_prepare_tokens_options` that call the request attribute with a predicate check:  i.e.

```
if ($self->has_request) {
    $tokens->{request} = $self->request;
    $tokens->{params}  = $self->request->params;
    $tokens->{vars}    = $self->request->vars;
}
```

I agree that it seems like a reasonable feature.

xsawyerx:
@veryrusty Funny, that was my exact implementation, which I just pushed as #1163

racke:
Good idea :+1:

veryrusty:
@girlwithglasses++ Thanks for raising the issue. 

#1163 has been merged. Closing 😺

PR #1163, opened 2016-04-29
Make template keyword global (GH #1159):

xsawyerx:
This makes sure the `template` keyword is available outside the request process. The request object (including the parameters and variables) will not be available during that `template` call.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -774,10 +774,12 @@ sub template {
     my $template = $self->template_engine;
     $template->set_settings( $self->config );
 
+    # A session will not exist if there is no request (global keyword)
+    #
     # A session may exist but the route code may not have instantiated
     # the session object (sessions are lazy). If this is the case, do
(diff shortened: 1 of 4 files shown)


veryrusty:
Added test case and predicate checks for a request before calling `has_session` which looks for a cookie in the request.

( Session objects are lazy; when a request exists we inflate the session object before rendering a template if the route code didn't require the session be instantiated. )

xsawyerx:
@girlwithglasses what do you think?

girlwithglasses:
Looks great to me -- thank you!

xsawyerx:
Should we merge this too?

veryrusty:
Would be nice to get one more thumbup. Pinging @racke, @SysPete or @cromedome for review

SysPete:
:+1: looks good to me. Change is small and there are people out there who want this.

veryrusty:
Thanks everyone.  Merged as 5d3cdca.  👯

Issue #1145, opened 2016-03-29
Replace Class::Load in favor of Module::Runtime?


xsawyerx:
* Is Class::Load with Class::Load::XS faster than Module::Runtime?
* Is there any reason to use Class::Load instead of Module::Runtime?

xsawyerx:
So, uh... answer is yes.

```
my $bench = Dumbbench->new(
    target_rel_precision => 0.005,    # seek ~0.5%
    initial_runs         => 20,       # the higher the more reliable
);

$bench->add_instances(
    Dumbbench::Instance::PerlEval->new(
        name => 'Module::Runtime',
        code => q!
            use Module::Runtime 'use_module';
            for (1 .. 1e5) { use_module('Moose') }
        !
    ),
    Dumbbench::Instance::PerlEval->new(
        name => 'Class::Load (XS)',
        code => q!
            use Class::Load 'load_class';
            for (1 .. 1e5) { load_class('Moose') }
        !
    ),
);

$ perl bench.pl
Module::Runtime: Ran 25 iterations (5 outliers).
Module::Runtime: Rounded run time per iteration: 3.0413e-01 +/- 2.4e-04 (0.1%)
Class::Load (XS): Ran 23 iterations (3 outliers).
Class::Load (XS): Rounded run time per iteration: 3.3367e-01 +/- 5.4e-04 (0.2%)
```

xsawyerx:
PR submitted at #1164.

xsawyerx:
Done.

PR #1164, opened 2016-05-02
Replace Class::Load with Module::Runtime

xsawyerx:
This addresses #1145.

Module::Runtime is a faster module than Class::Load (even with
Class::Load::XS available) and is already available by our stack.

If and when App::Cmd removes its dependency on Class::Load, we
will not require this by our stack at all.

### DIFF
--- a/lib/Dancer2.pm
+++ b/lib/Dancer2.pm
@@ -4,7 +4,7 @@ package Dancer2;
 use strict;
 use warnings;
 use List::Util  'first';
-use Class::Load 'load_class';
+use Module::Runtime 'use_module';
 use Import::Into;
 use Dancer2::Core;
 use Dancer2::Core::App;
@@ -85,7 +85,7 @@ sub import {
     $final_args{dsl} ||= $config_dsl;
 
     # load the DSL, defaulting to Dancer2::Core::DSL
-    load_class( $final_args{dsl} );
+    use_module( $final_args{'dsl'} );
     my $dsl = $final_args{dsl}->new( app => $app );
     $dsl->export_symbols_to( $caller, \%final_args );
 }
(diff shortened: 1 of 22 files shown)


racke:
:+1:

lib/Dancer2.pm
```diff
@@ -85,7 +85,7 @@ sub import {
     $final_args{dsl} ||= $config_dsl;
 
     # load the DSL, defaulting to Dancer2::Core::DSL
-    load_class( $final_args{dsl} );
+    use_module( $final_args{'dsl'} );
     my $dsl = $final_args{dsl}->new( app => $app );
     $dsl->export_symbols_to( $caller, \%final_args );
```
veryrusty: This could be chained:

`my $dsl = use_module(  $final_args{'dsl'} )->new( app => $app );`

xsawyerx: D'oh! I intended to do that! @veryrusty++

xsawyerx: Done!

SysPete:
:+1:

veryrusty:
We _try_ to maintain compatibility back to perl 5.8.n. Module::Runtime works around a couple of core bugs before the 5.11 release. 

Should we be using `require_module` instead of `eval { require Some::Module; 1 }` to take advantage of that ?

xsawyerx:
@veryrusty added, but only for `lib`, not `t`.

veryrusty:
@xsawyerx++   :+1: with the `require_module` changes.

veryrusty:
Merged as 1e97ed3a. Thanks all!

xsawyerx:
Thank you. :)

PR #1167, opened 2016-05-05
Bugfix/forward proxy

### LINKED ISSUE #1165: forward doesn't set is_behind_proxy option
Dancer2 is new to me, so please bear with me.
I'm deploying an application behind an Apache reverse proxy. Apache terminates SSL. Setting the configuration option behind_proxy to true ensures that methods like request->base and uri_for return an URI with protocol https. 
However, when doing forward to another route the behind_proxy setting is lost in the forwarded request,
If not an intended design, perhaps I may suggest the attached fix. Keep in mind my inexperience and that I probably don't have a clue.

Thank you for an excellent piece of software.

[App.pm.zip](https://github.com/PerlDancer/Dancer2/files/249463/App.pm.zip)

veryrusty:
Copy `is_behind_proxy` (which comes from the `behind_proxy` setting) into new request on forward.

Resolves #1165

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1198,6 +1198,8 @@ sub make_forward_to {
     $new_request->{_route_params} = $request->{_route_params};
     $new_request->{body}          = $request->body;
     $new_request->{headers}       = $request->headers;
+    # Copy remaining settings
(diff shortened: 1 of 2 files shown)


SysPete:
:+1:

cromedome:
:+1:

veryrusty:
Merged as 1112f28.  Thanks everyone! 👯

PR #1171, opened 2016-05-27
Don't call private method for Request query params 

veryrusty:
Private methods are, well private. The may change.

The Plack 1.0040 (TRIAL) release removes the _parse_query method which
we were calling. Far safer to call the overridden public method via SUPER.

(Similar fix to #1148 for body params)


veryrusty:
I've verified the test suite passes with both Plack v1.0039 and v1.0040 with this patch.

SysPete:
:+1: for immediate pull

PR #1172, opened 2016-05-28
Rebased: Add `send_as` keyword (from Dancer2::Plugin::SendAs)

veryrusty:
This is #1141 rebased onto current master with the following changes:
  * Uses Module::Runtime instead of Class::Load
  * The "types" must match the casing of the serializer names; i.e. `send_as JSON => $data`
to clean up the last concerns from @xsawyerx.

@xsawyerx @SysPete any chance of a quick review in time for the next release as #1141 had thumbs up :)

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -2822,15 +2822,18 @@ Returns request header(s).
     get '/get/headers' => sub {
         my $xfoo = request_header 'X-Foo';
         ...
+    };
+
 =head2 send_as
 
 Allows the current route handler to return specific content types to the
 client using either a specified serializer or as html.
 
-Any Dancer2 serializer may be used, serialization classes will be loaded
-as required. Serializer configuration may be added to your apps
-C<engines> configuration. If no serializer can be found or C<HTML> is
-specified, the content will be returned assuming it is HTML with
+Any Dancer2 serializer may be used. The specified serializer class will
+be loaded if required, or an error generated if the class can not be found.
+Serializer configuration may be added to your apps C<engines> configuration.
+
+If C<HTML> is specified, the content will be returned assuming it is HTML with
 appropriate C<Content-Type> headers and encoded using the apps configured
 C<charset> (or UTF-8).
 
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -2822,6 +2822,50 @@ Returns request header(s).
     get '/get/headers' => sub {
         my $xfoo = request_header 'X-Foo';
         ...
+=head2 send_as
+
+Allows the current route handler to return specific content types to the
+client using either a specified serializer or as html.
+
+Any Dancer2 serializer may be used, serialization classes will be loaded
+as required. Serializer configuration may be added to your apps
+C<engines> configuration. If no serializer can be found or C<HTML> is
+specified, the content will be returned assuming it is HTML with
+appropriate C<Content-Type> headers and encoded using the apps configured
+C<charset> (or UTF-8).
+
+    set serializer => 'YAML';
+    set template   => 'TemplateToolkit';
+
+    # returns html (not YAML)
+    get '/' => sub { send_as HTML => template 'welcome.tt' };
+
+    # return json (not YAML)
+    get '/json' => sub {
+        send_as JSON => [ some => { data => 'structure' } ];
+    };
+
+C<send_as> uses L<send_file> to return the content immediately. You may
+pass any option C<send_file> supports as an extra option. For example:
+
+    # return json with a custom content_type header
+    get '/json' => sub {
+        send_as JSON => [ some => { data => 'structure' } ],
+                { content_type => 'application/json; charset=UTF-8' },
+    };
+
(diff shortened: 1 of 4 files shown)


SysPete:
:+1: all looks good.

xsawyerx:
Good stuff.

If someone doesn't provide a correct type (`jSON` instead of `JSON`), it silently ignores the error and sends it as HTML. Shouldn't this just fail instead?

*You asked to send it in a type we do not know. Did you misspell the output?*

I understand that the problem is that there is no `HTML` serializer. That's why you're counting on the failure to cause an HTML rendering. This means two things:

* If you provide it as `hTML`, it will still work, despite us saying you now have to use a proper name.
* More importantly, if someone uploads `Dancer2::Serializer::HTML` (which isn't a serialization format, really), this will break.

We can fix it by adding an `HTML` serializer (which would do what, exactly? I don't know...) or by either altering the keyword (`send_as HTML => {}` being a special case) or by adding a new keyword for HTML (`send_as_html`). I like altering the keyword to introduce a special case.

xsawyerx:
And to be clear: The "Good stuff" comment was not sarcastic. Thank you for doing this, rusty. :)

SysPete:
On 28/05/16 12:44, Sawyer X wrote:
>
> Good stuff.
>
> If someone doesn't provide a correct type (|jSON| instead of |JSON|),
> it silently ignores the error and sends it as HTML. Shouldn't this
> just fail instead?
>
> /You asked to send it in a type we do not know. Did you misspell the
> output?/
>
Good point.
>
> I understand that the problem is that there is no |HTML| serializer.
> That's why you're counting on the failure to cause an HTML rendering.
> This means two things:
>
>   * If you provide it as |hTML|, it will still work, despite us saying
>     you now have to use a proper name.
>   * More importantly, if someone uploads |Dancer2::Serializer::HTML|
>     (which isn't a serialization format, really), this will break.
>

This is something that didn't occur to me but perhaps someone *could* do
something this crazy.

> We can fix it by adding an |HTML| serializer (which would do what,
> exactly? I don't know...) or by either altering the keyword (|send_as
> HTML => {}| being a special case) or by adding a new keyword for HTML
> (|send_as_html|). I like altering the keyword to introduce a special case.
>

Don't like send_as_html since where do we stop if we go down that route?
send_as_json, ...?

veryrusty:
Thanks for all the feedback! Addes two more commits:
  * Special case `send_as HTML => $content`.
  * For any other "type", try and load `Dancer2::Serializer::$type`, croak if that fails.

lib/Dancer2/Manual.pod
```diff
@@ -2822,6 +2822,50 @@ Returns request header(s).
...
+
+Allows the current route handler to return specific content types to the
+client using either a specified serializer or as html.
+
+Any Dancer2 serializer may be used, serialization classes will be loaded
+as required. Serializer configuration may be added to your apps
+C<engines> configuration. If no serializer can be found or C<HTML> is
+specified, the content will be returned assuming it is HTML with
+appropriate C<Content-Type> headers and encoded using the apps configured
+C<charset> (or UTF-8).
```
SysPete: Croaks if no serializer is provided of if serializer class cannot be loaded.

veryrusty: Thanks @SysPete. Updated in bb5caf5

xsawyerx:
Awesome, thanks.

Last thing I see now is that if you call `send_as hTML => ...`, you will get an error that we couldn't load your specific serializer. Two ways out:

* call `lc($type) eq 'html' and ...` (either throwing an error/warning for anything *but* `HTML` or allow anything `lc($type) eq 'html'`)
* Introduce `D2::Serializer::HTML` that just renders.

I think the former is more practical, since having a message saying we can't load an HTML serializer is probably more confusing.

veryrusty:
I **do not** like the idea of D2::Serializer::HTML.

The string `html` was chosen to reflect the default content type applied to the response (`text/html`); I considered `send_as 'text/html' => ...` as being too unwieldily. Expecting html as the most likely content type, I chose `send_as html => ...` as it was a similar syntax to wanting serialized content. 

If we don't want the "can't load serializer" errors for `send_as hTML => ...`, my preference is to go for the simpler `lc($type) eq 'html'` logic.

xsawyerx:
I agree the `HTML` serializer is not a good idea. I noted that as a technical option, but I would also prefer we don't do it.

I agree with doing `lc()` on it. We can throw a warning saying "please use **html** next time". Doesn't matter as much, though.

SysPete:
I'd say just lc() and don't bother warning for bad case as I'm not sure the cost of a conditional is worth it for the little value it adds.

xsawyerx:
That would mean no need for deprecation cycle if we want to change anything in the guts of it again, but meh.. It's more than good enough for me as is.

xsawyerx:
Added both the `lc` and the check for a non-`lc` typo (`HTML`, `Html`, etc.). I just imagine people making such a common mistake and I would like to keep this part already "deprecated" so we can remove support for it in the future if we need it.

@veryrusty++
@SysPete++

PR #1173, opened 2016-05-30
Reword logged error if serializaiton fails

### LINKED ISSUE #1152: Confusing exception when the response is not serializable
```
Failed to serialize the request: encountered object <snip>, but neither allow_blessed, convert_blessed nor allow_tags settings are enabled (or TO_JSON/FREEZE method missing)
  at <snip>/lib/site_perl/5.18.2/JSON.pm line 154.
  in <snip>/lib/site_perl/5.18.2/Dancer2/Core/Response.pm l. 250
```

The error message says "Failed to serialize the **request**", but the exception is thrown from Dancer2::Core::Response.

veryrusty:
Serializers can be used in more places that request/response data.
Having serializers log serialization errors against the "request" is
likely to be confusing.

Instead, log the failure stating we failed to serialize the "content",
which is applicable for a response body, or some other data that failed
to serialize elsewhere.

Resolves #1152.

### DIFF
--- a/lib/Dancer2/Core/Role/Serializer.pm
+++ b/lib/Dancer2/Core/Role/Serializer.pm
@@ -50,7 +50,7 @@ around serialize => sub {
     } or do {
         my $error = $@ || 'Zombie Error';
         blessed $self
-            and $self->log_cb->( core => "Failed to serialize the request: $error" );
+            and $self->log_cb->( core => "Failed to serialize content: $error" );
     };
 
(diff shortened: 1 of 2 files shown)


SysPete:
Simple and should resolve OPs issue :+1:

racke:
:+1:

veryrusty:
Merged as 7a14d3a.  Thanks everyone.

PR #1174, opened 2016-05-30
Update plugin tests to stop deprecation warnings.

SysPete:
As pointed out by @veryrusty we have some plugin tests that are using deprecated methods.

- Stop using `plugin_args` and `hook`
- Call `request` as `app->request`.


veryrusty:
:+1: @SysPete++

Merged 👯

PR #1175, opened 2016-06-01
Plugins in all the namespaces

veryrusty:
Plugin2 requires plugins to be in the Dancer2 namespace. This is an attempt to liberate that..
Consider this "work in progress" - wanted to get this out there and get some discussions going.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -60,7 +60,11 @@ sub _with_plugin {
         return $plugin;
     }
 
-    $plugin =~ s/^/Dancer2::Plugin::/ unless /^Dancer2::Plugin::/;
+    # short plugin names get Dancer2::Plugin:: prefix
+    # plugin names starting with a '+' are full package names
+    if ( $plugin !~ s/^\+// ) {
+        $plugin =~ s/^/Dancer2::Plugin::/ unless /^Dancer2::Plugin::/;
+    }
 
     # check if it's already there
     if( my ( $already ) = grep { $plugin eq ref $_ } @{ $self->plugins } ) {
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -72,8 +72,9 @@ sub execute_plugin_hook {
     my ( $self, $name, @args ) = @_;
     my $plugin_class = ref $self;
 
-    $plugin_class =~ s/^Dancer2::Plugin:://
+    $self->isa('Dancer2::Plugin')
         or croak "Cannot call plugin hook ($name) from outside plugin";
+    $plugin_class =~ s/^Dancer2::Plugin:://; # short names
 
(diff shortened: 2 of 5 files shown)


lib/Dancer2/Plugin.pm
```diff
@@ -273,7 +270,7 @@ sub _exporter_app {
                 # if the plugin is a
                 # "candidate" for a hook
                 # See: App.pm "execute_hook" method, "around" modifier
-                if ( ref( $_[0] ) =~ /^Dancer2::Plugin::/ ) {
+                if ( $_[0]->isa('Dancer2::Plugin') ) {
```
xsawyerx: Much better.

lib/Dancer2/Core/App.pm
```diff
@@ -60,7 +60,11 @@ sub _with_plugin {
         return $plugin;
     }
 
-    $plugin =~ s/^/Dancer2::Plugin::/ unless /^Dancer2::Plugin::/;
+    # short plugin names get Dancer2::Plugin:: prefix
+    # plugin names starting with a '+' are full package names
+    if ( ! ( $plugin =~ s/^\+// ) ) {
```
xsawyerx: "If not regex match"? Why not "if regex does not match"?

lib/Dancer2/Plugin.pm
```diff
@@ -895,7 +892,7 @@ This is a (relatively) simple way for a plugin to use another plugin:
         default => sub {
             # if the app already has the 'Polite' plugin loaded, it'll return
             # it. If not, it'll load it in the app, and then return it.
-            scalar $_[0]->app->with_plugins( 'Polite' )
+            scalar $_[0]->app->with_plugin( 'Polite' )
```
SysPete: I agree with the singular method name in this example but I don't think `scalar` should be there.

SysPete: Let me phrase that better: could someone please explain to me why `scalar` is needed here? Otherwise it should go. I don't use it in the various plugins I've released in the last 24 hours.

veryrusty: Agreed. Asking for the scalar context of an object (blessed reference) gives you the object anyway.

yanick:
Looks good!

SysPete:
:+1: like yanick says

veryrusty:
Thanks for the feedback @xsawyerx and @SysPete.  Added commits to address those items.

Added tests for plugin hooks in other namespaces (and a commit to make that work). That cleans up any outstanding points I wanted to chase up.

xsawyerx:
If our plugin compat guru (@SysPete) approves, I approve. :+1: 

Thank you for fixing this!

veryrusty:
Force-pushed a rebased set of commits to get travis to test 'em.

veryrusty:
Merged! Thanks everyone :dancers:

PR #1176, opened 2016-06-01
Add Test::Deep dependency for testing

nunorc:
Some new test files require Test::Deep.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -133,6 +133,7 @@ Test::Fatal = 0
 HTTP::Body = 0
 HTTP::Cookies = 0
 HTTP::Headers = 0
+Test::Deep = 0
 
 ; for maintainers, see with mst how to avoid these
 ; strictures = 0


SysPete:
Thanks @nunorc - good catch.

I think I'd prefer not to add the dep since this is only in the new plugin2 tests. I'll cook up a new PR which drops Test::Deep then we see what other devs prefer.

SysPete:
So the question is do we add the build dep on Test::Deep or use modified tests as per #1177?

veryrusty:
@nunorc++ nice catch.

Since the changes to remove Test::Deep are fairly minimal, I'd prefer to see the test dependency removed.

xsawyerx:
@nunorc++

I agree in just removing `Test::Deep`.

nunorc:
Either option seems fine, as long as we can build an successfully test the package from a clean environment.

I'm closing this request if favor of https://github.com/PerlDancer/Dancer2/pull/1177. Thank you.

PR #1177, opened 2016-06-01
Stop using Test::Deep since we don't declare it as a build dep

SysPete:
Before plugin2 we didn't use Test::Deep and for the few tests for which it was used it is simple enough to rework these tests to use something else.

This is an alternative option to PR #1176 which adds a dep on Test::Deep.

### DIFF
--- a/t/plugin2/define-keywords.t
+++ b/t/plugin2/define-keywords.t
@@ -2,7 +2,6 @@ use strict;
 use warnings;
 
 use Test::More tests => 1;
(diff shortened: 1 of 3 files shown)


xsawyerx:
:+1:

racke:
:+1:

veryrusty:
@nunorc++ for reporting the Test::Deep dependency, @SysPete++ for removing it.

Rebased and merged as 334c164. Thanks all!

Issue #757, opened 2014-11-07
Default behavior for logger?


xsawyerx:
What should be the default behavior if no logger has been specified? Do we want the STDERR logger, do we want the capture logger?

What do you think?

yanick:
STDERR++  for two reasons. One, that's the default behaviour most would expect. Two: using the Capture log is a sure way to set a memory leak for most of the applications. Since there is no output, most peeps who wouldn't have read the docs closely will have no idea it's logging, and will have a heck of a nasty surprise as the memory usage grows and grows and grows. :-)

Btw, that's kinda related to perldancer/Dancer#1040

sakshee3:
As I can see in the Core::App::_build_default_config, 'console' is already the default valiue. Am I missing anything ?

cromedome:
Default logger is currently console, and from what I have mined from GitHub, since most people want the default to be STDERR, I am considering this resolved. Closing... thanks all!

PR #1179, opened 2016-06-02
Config take2

perlpilot:
@veryrusty @cromedome 

I've re-worked the configuration changes such that the configuration files are loaded the same as they are currently, but *_local.* files are looked for and loaded if present. i.e. for a development environment, the following config files will be loaded in this order if they exist: config.yml, config_local.yml, environments/development.yml environments/development_local.yml
It will still load config.yml and config.json if they both exist.

Also, you can restrict the extensions that Dancer2 will try to a single extension by setting the DANCER_CONFIG_EXT environment variable.

I added some tests, but more are probably needed.

And the things I'm less sure about ...

I also made it output some diagnostic information when loading the configuration files. You can turn it on by setting the DANCER_CONFIG_VERBOSE environment variable.

The diagnostic info is output using warn.   I tried $self->app->log but it didn't work for some reason and I'm too tired to investigate right now.  (maybe it didn't work because I'm too tired to do it correctly :)

### DIFF
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -29,20 +29,42 @@ those settings into a configuration file.
 
 =head2 Configuration file path and file names
 
-Dancer2 will first look for the file F<config.ext> (where .ext is the type
-of configuration file you are using) in the root directory of your
-application.  This is considered your global Dancer2 config file.  If you do
-not care to have separate settings for production and development
-environments (B<not> a recommended practice!), then this file is all you
-need.
+Dancer2 will first look for the file F<config.EXT> (where F<EXT> is the
+type of configuration file you are using; e.g. F<ini> or F<json> or
+F<yml>) in the root directory of your application. This is considered
+your global Dancer2 config file. If you do not care to have separate
+settings for production and development environments (B<not> a
+recommended practice!), then this file is all you need.
+
+Next, Dancer2 will look for a file called F<config_local.EXT>. This file
+is typically useful for deployment-specific configuration that should
+not be checked into source control. For instance, database credentials
+could be stored in this file.  Any settings in this file are merged into
+the existing configuration such that those with the same name in your 
+global configuration file will take precedence over those settings in 
+the global file. 
 
 Next, Dancer2 will look in the F<environments> directory for a configuration
-file specific to the platform you are deploying to (F<production.ext> and
-F<development.ext>, for example).  Any settings in these files that are
-named the same as settings in your global configuration file will take
-precedence over those settings in the global file. The rest of the settings
-are merged and Dancer2 uses the combination of settings from the two files
-as its operating configuration.
+file specific to the platform you are deploying to (F<production.EXT> or
+F<development.EXT>, for example).  Again, the configuration from the environment
+is merged with the existing configuration with the deployment config taking
+precedence.
+
+Finally, Dancer2 will look in the F<environments> directory for a
+local configuration for the specific platform you are deploying to
+(e.g. F<prodcution_local.EXT> or F<development_local.EXT>) 
+The configuration in this file is merged as before.
+
+Much like F<config_local.EXT>, this file would be useful for environment-
+specific configuration that would not be checked into source control.
+For instance, when developing an application that talks to multiple services, 
+each developer could have their own URLs to those services stored within
+their F<environments/development_local.yaml> file.
+
+Note, if there is no F<config.EXT>, Dancer2 will not look for a
+F<config_local.EXT>. The same is true for the local environment
+configuration.
+
 
 =head2 Supported configuration file formats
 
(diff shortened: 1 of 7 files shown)


veryrusty:
Finally got a chance to have a look over this Pr. @perlpilot thanks for doing this! Looks ok at a first pass.

Its likely the logger is being configured in the configuration files being loaded. There are some comments in #757 about what the "default" logger behaviour should be if none exists (hint: STDERR). I want to look to see if that approach can be used instead of the "warn"s here.  I also wondered if `DANCER_CONFIG_VERBOSE` might be better simplified to `DANCER_DEBUG` or similar; but thay may have other connotations.  Pinging @xsawyerx and @mickeyn for their thoughts too.

perlpilot:
@cromedome, @veryrusty, @xsawyerx  ... ping.   Anything else you want me to do with this?  Any comments or questions?

xsawyerx:
Let me take a look at it tomorrow. (12am here now.)

xsawyerx:
I'll preface by saying I am not against this feature. Our policy is that two core devs +1'ing this without objection means it's in. I do not object to it.

Having said this, I have reservations about this. My concern is that the way this is implemented (vs. the concept) is singling out this use-case (having local non-version-checked files) from a more abstract use-case (having automatically included files). If we supported being able to load *any* additional file, this `_local` wouldn't be required.

For example:

```yaml
# config.yml

# new key:
include_files:
    - production/local.yml
```

This would allow the configuration to (silently, I suppose) load additional include files. As many as you want. You could put it in `production.yml` or `config.yml`. Your pick.

I just feel like opting for a more general purpose solution would be better.

racke:
Yeah, I also like the "include" idea much more. That gives more flexibility and a lot of other servers do that too.

This would also allow something like:

```
include_files:
    - plugin.d/*-conf.yml
```

xsawyerx:
Glob syntax is a common feature. :+1: from me.

perlpilot:
Having an "include" feature would be nice, but for it to be supported for all configuration file formats,  Dancer would have to recognize at least one configuration setting and treat it specially when building the configuration or have some special syntax that is outside the purview of the config parser that is specifically for including files (sort of like a preprocessor before the config parser even gets to it).  This is currently unprecedented AFAIK.

And what would it be called?  If it's part of the normal config options, whatever it is would be a name that we're stealing from our users to use in their own configuration.  (Of course, something like "dancer_include_files" might be sufficiently unique and unwanted by our users for their own purposes)  If it's not a normal config option and we parse it specially before the config parse even sees it, we can call it anything we want, but it would do us well to probably make it stand out from whatever config format is being used.

Currently, we build a complete list of all config files, then those files are loaded and merged in order into the configuration.  If we're including files, then the process of loading and merging changes drastically.   I'm not saying this is bad, just that it must be.

Also, if included files can include other files do we set a limit on how deep they can do this?  If file A includes file B includes file C includes file D etc, is that okay?  Do we want to detect cycles for when they accidentally make an include loop? 

What happens if the magic string for including files appears more than once within the config?  If 
we're using a normal config option in JSON/YAML/XML/INI/whatever format, I think each config format has their own way of dealing with multiple keys, but I may be wrong about that.

Can file formats other than the current one be included?  I.e. can I include a JSON config in a XML config?

If I can have a config file structured like
> config stuff
> include file A
> more config stuff
> include file B
> even more config stuff
If any of the config is the same within file A or B as within the file that is including those files, does the included config override whatever is in the main config?   Does the "more config stuff" overrride whatever was in file A?  Does the config in file B override what came before it lexically in the file?  How would any of this work?  (If including is just straight text substitution into the file and then the one realized config is loaded, that would defeat the whole purpose of the original intent which was to load files in a certain order such that files loaded later over ride the settings in earlier files)  In short, what's the config merge order? 

Sorry, this is just a brain-dump of the things that occur to me for an "include" feature.

pdl:
The way we're doing local developer configs is the other way round - inheritance rather than inclusion, which seems to me to be a much more well-defined problem:

    application -e myconfig

loads `environments/myconfig.yml`.

- `myconfig.yml` has a key `inherits_from: "development"`, so we load `environments/development.yml` and merge into it;
- `development.yml` has a key `inherits_from: "base"`, so we load `environments/base.yml` and merge into it;
- `base` has no inherits_from so it's so we load `config.yml` and merge into it, and we're done.

In each case, we're performing straightforward load and merge. We also keep a record of the inheritance.

Currently, we've not implemented multiple inheritance but it wouldn't be too hard, and the semantics should be fairly obvious:

    inherits_from:
    - development
    - passwords

Last item in the list takes precedence.

This way you can effectively subclass the configs without making local changes to the parents to know about them.

Currently, building the config happens outside of Dancer.

perlpilot:
@cromedome @xsawyerx @veryrusty @racke 
Anyone have any response to my comments above?  Or better ... anyone want to merge my PR?  :-)

I still don't think that an include feature is quite right (yet) because I don't quite see how or if overriding settings is possible plus it would require a "large" semantic change in how config files are processed. (Note that @pdl's approach uses configuration manipulation outside the purview of  Dancer)

To reiterate, the main gist of my config patches are to take the existing "layered" configuration (where more specific configuration settings override more general settings) and allow the Dancer developer to exert a little more control on the process.

From most general to most specific, we'd have ...
* global Dancer app config  (`config.*`)
* deployment-specific global config (`config_local.*`)
* environment config (`environments/development.*`)
* deployment-specific environment config (`environments/development_local.*`)

Where settings in config files further down the list will replace same-named settings from config files above.

xsawyerx:
I'll review it in the following hours!

xsawyerx:
Alright. :+1: :)

SysPete:
There *are* other potential ways of extending config options but since @perlpilot is the only one who has written any code and it adds enough flexibility for what I need now then also :+1: .

cromedome:
Merged. Thanks!

PR #1183, opened 2016-06-07
Tutorial.pod: before_template_render

brxfork:
before_template becomes before_template_render


brxfork:
I don't understand why Tracis CI is unhappy whit this pod file and Perl 5.12...

PR #1184, opened 2016-06-07
hook before_template updated in two pod files

brxfork:
before_template becomes before_template_render
nota: I'm probably wrong to PR Manual.pod a second time (sorry for that)


veryrusty:
Merged in c0d9645. 

`Dancer2::Core::Hook` has a special case for aliasing a hook named `before_template` to `before_template_render`. Both (should) work. The docs should use the correct hook name :)

@brxfork++. Thanks again!  👯

Issue #1187, opened 2016-06-13
read_file_content should include $! in open error messages


exercism-1:
Dancer2::Template::Simple can produce errors like:

    Can't open '/home/vincent/mywebapp/views/test.cgi.tt' using mode '<' at /usr/local/share/perl/5.20.2/Dancer2/Template/Simple.pm line 31.

But it doesn't say why the open failed (file not found, permission denied, etc).

https://github.com/PerlDancer/Dancer2/blob/38c9dee87581ca7ef85b4447f14a316066b50048/lib/Dancer2/FileUtils.pm#L46 does

    croak "Can't open '$filename' using mode '$mode'";

That should be changed to

    croak "Can't open '$filename' using mode '$mode': $!";

PR #1188, opened 2016-06-13
include failure reason in error message from open_file (#1187)

exercism-1:


SysPete:
@exercism-1 thanks for the contribution: seems like a sensible change to me.

I guess a change to the test to check for `$!` is probably a good idea. Something like:
```
@@ -27,7 +27,7 @@ sub hexe {
 
 like(
     exception { Dancer2::FileUtils::open_file( '<', '/slfkjsdlkfjsdlf' ) },
-    qr{/slfkjsdlkfjsdlf' using mode '<'},
+    qr{/slfkjsdlkfjsdlf' using mode '<': \w},
     'Failure opening nonexistent file',
 );
```
perhaps?

racke:
Definitely makes sense to pass the error message to the hapless developer :-)

Issue #1185, opened 2016-06-08
t/classes/Dancer2-Core-Request/new.t fails with perl 5.25.1


eserte:
After patching Type::Tiny it's possible to build and test Dancer2 on perl 5.25.1 --- but one test fails:
```
    #   Failed test 'Dancer2::Core::Request->can(...)'
    #   at t/classes/Dancer2-Core-Request/new.t line 192.
    #     Dancer2::Core::Request->can('accept_charset') failed
    #     Dancer2::Core::Request->can('accept_encoding') failed
    #     Dancer2::Core::Request->can('accept_language') failed
    #     Dancer2::Core::Request->can('connection') failed
    #     Dancer2::Core::Request->can('keep_alive') failed
    #     Dancer2::Core::Request->can('x_requested_with') failed

    #   Failed test 'Request successful'
    #   at t/classes/Dancer2-Core-Request/new.t line 263.
    # Looks like you failed 2 tests of 6.

#   Failed test 'Defaults'
#   at t/classes/Dancer2-Core-Request/new.t line 267.
# Looks like you failed 1 test of 9.
t/classes/Dancer2-Core-Request/new.t ................... 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/9 subtests 
```
@andk: looks like a bisect candidate.

andk:
Hey, I got this in my inbox, interesting, so github does have a CC feature.

Bisect leads us to v5.25.0-71-g5d32d26, same as https://rt.perl.org/rt3/Ticket/Display.html?id=128179

andk:
Oops, I spoke too soon. There is randomness involved. I now see various successes with 71 and after

andk:
No, it was not randomness, my bisecting was not paying attention to the distroprefs needed. With manual help now we have a new bisect result arriving at v5.25.0-80-g2e6807b

A BBC is broadcasted as https://rt.perl.org/Ticket/Display.html?id=128352

Thanks!

xsawyerx:
Thank you, @eserte and @andk!

PR #1190, opened 2016-06-16
don't use barewords under use strict

tonycoz:
and report errors when dynamically creating subs.

Fixes https://github.com/PerlDancer/Dancer2/issues/1185


tonycoz:
Oops, fair warning, this was tested against the version my CPAN client downloaded and fixed the error there, but waiting for Dist::Zilla's deps to build and test on a blead perl build I was planning to throw away wasn't worth the time.

xsawyerx:
@tonycoz Thank you for the PR! That is perfectly fine. I will apply it immediately on blead, test, and push a release.

xsawyerx:
Merged, thanks! :+1:

Issue #1181, opened 2016-06-06
Make sure all Session engines work with current release


SysPete:
List does not include engines that ship with Dancer2 (Simple and YAML).

| Engine | CPAN | repo | Date checked | _change_id support ? | Status |
| --- | --- | --- | --- | --- | --- |
| CGISession | [CPAN](https://metacpan.org/pod/Dancer2::Session::CGISession) | [GitHub](https://github.com/pierre-vigier/Dancer2-Session-CGISession) | 2016-08-16 | :+1: | :+1: |
| Cookie | [CPAN](https://metacpan.org/pod/Dancer2::Session::Cookie) | [GitHub](https://github.com/PerlDancer/Dancer2-Session-Cookie) | 2016-08-16 | :+1: (actually a noop) | :+1: |
| DBIC | [CPAN](https://metacpan.org/pod/Dancer2::Session::DBIC) | [GitHub](https://github.com/interchange/Dancer2-Session-DBIC) | 2016-07-07 | :+1: | :+1: |
| JSON | [CPAN](https://metacpan.org/pod/Dancer2::Session::JSON) | [GitHub](https://github.com/PerlDancer/dancer2-session-json) | 2016-07-07 | :+1: (SessionFactory::File) | :+1: |
| Memcached | [CPAN](https://metacpan.org/pod/Dancer2::Session::Memcached) | [GitHub](https://github.com/PerlDancer/dancer2-session-memcached) | 2016-08-16 | :+1: | :+1: |
| MongoDB | [CPAN](https://metacpan.org/pod/Dancer2::Session::MongoDB) | [GitHub](https://github.com/PerlDancer/dancer2-session-mongodb) | 2016-08-18 | :+1: | :+1: |
| PSGI | [CPAN](https://metacpan.org/pod/Dancer2::Session::PSGI) | [GitHub](https://github.com/veryrusty/dancer2-session-psgi) | 2016-07-07 | :+1: | :+1: |
| Redis | [CPAN](https://metacpan.org/pod/Dancer2::Session::Redis) | [GitHub](https://github.com/burnersk/Dancer2-Session-Redis) | 2016-08-16 | :+1: | :+1: |
| Sereal | [CPAN](https://metacpan.org/pod/Dancer2::Session::Sereal) | [GitHub](https://github.com/PerlDancer/dancer2-session-sereal) | 2016-08-16 | :+1: (SessionFactory::File) | :+1: |

SysPete:
[...]

PR #1195, opened 2016-07-04
Add change_session_id method to App

SysPete:
It is considered good security practice to change session id on any change of privilege level (for example on login):

https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Renew_the_Session_ID_After_Any_Privilege_Level_Change

This is also a requirement of a number of established security standards.

This branch adds a new method `change_session_id` which will do one of the following:

1. A new method `_change_id` can be added to session engines which is then expected to do the "right thing" for that session engine.

1. Otherwise the session data is stashed, the session destroyed, and then the new session has the old data written back into it.

It is expected that the new session engine method will become a requirement for all session engines at some point in the future.

This PR is currently missing tests and documentation for the new hooks that have been added. These will be added soon.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -484,6 +484,37 @@ sub setup_session {
     }
 }
 
+sub change_session_id {
+    my $self = shift;
+
+    # Find the session engine
+    my $engine = $self->session_engine;
+
+    if ($engine->can('_change_id')) {
+        $engine->change_id;
+    }
+    else {
+
+        # grab data, destroy session and store data again
+        my $session = $self->session;
+        my %data = %{$session->data};
+
+        # destroy existing session
+        $self->destroy_session;
+
+        # get new session
+        $session = $self->session;
+
+        # write data from old session into new one
+        while (my ($key, $value) = each %data ) {
+            $session->write($key => $value);
+        }
+
+        # clear out destroyed session - no longer relevant
+        $self->clear_destroyed_session;
+    }
+}
+
 has prefix => (
     is        => 'rw',
     isa       => Maybe [Dancer2Prefix],
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -487,6 +487,8 @@ sub setup_session {
 sub change_session_id {
     my $self = shift;
 
+    my $session;
+
     # Find the session engine
     my $engine = $self->session_engine;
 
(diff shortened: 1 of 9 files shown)


SysPete:
It seems we have no existing documentation for session hooks so I haven't added anything for the new hooks. Otherwise documentation and tests are finished so time for review please...

In the meantime I'll test against all existing session engines.

SysPete:
This has now been tested against all engines and is working fine except for PSGI which needs patching (https://github.com/veryrusty/Dancer2-Session-PSGI/pull/4). All other engines work via the fallback (stash/destroy/create) and I've pushed changes to DBIC and Memcached for native support (pretty simple changes in both cases).

So any +1's anywhere?

cromedome:
Emphatic 👍  from me! Looks great!

racke:
Looks good to me as well :+1:

veryrusty:
:+1: Nice work!  :beers: for @SysPete. 

Dancer2::Session::PSGI has been patched and a new release pushed to CPAN.

lib/Dancer2/Core/App.pm
```diff
@@ -484,6 +484,62 @@ sub setup_session {
...
+        # get a different cookie value, no?
+        #
+        # TODO: think about this some more.
+
+        # grab data, destroy session and store data again
+        my %data = %{$session->data};
+
+        # destroy existing session
+        $self->destroy_session;
+
```
SysPete: @xsawyerx as an alternative to cooking up a new attribute I'm wondering whether it might be simpler to allow a flag to be passed to `destroy_session` that tells it not to stash the old session in the `destroyed_session` attribute. Then I don't have to worry about ordering since `clear_destroyed_session` will not need to be called.

Perhaps `destroy_session` could look something like:

```
sub destroy_session {
    my ( $self, $force ) = @_;
    ...
    $self->set_destroyed_session($session) unless $force;
    ...
}
```
Args can be hashref instead or whatever you prefer. Would this be acceptable?

SysPete:
Since we have very few session engines on CPAN and probably far fewer (if any) on darkpan I think the simplest way forward is to keep the code as is BUT add a deprecation notice for session engines that do not have support for `_change_id`. We could keep this in place for 6 months then simply require `_change_id` and remove the compat code.

@xsawyerx would this approach be acceptable to you?

SysPete:
All session engines now support this new feature (see #1181) so the compat code in this PR will only handle old versions of session engines so I am no longer concerned that that small piece of the PR is a little brittle. I suggest we add deprecation notice to the compat code in a future D2 release.

cromedome:
@SysPete I am for that. :+1:

PR #1199, opened 2016-07-06
Simplify YAML version range dep using != 1.16

SysPete:
Now newer version will be acceptable including 1.17 which does not break D2.


racke:
:+1: Looks good

PR #1206, opened 2016-07-08
Dancer2::Core::Error::_censor():  Improve detection of sensitive data.

jkeenan:
In addition to checking the keys of each hash at each level, we should also
check the keys of hashes blessed into classes.

We should also check for the values of sensitive keys in key-value pairs in
query clauses, e.g., '&pass=my_top_hush_pwrd'.

t/error.t: Add subtest block explicitly testing _censor().

### DIFF
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -9,6 +9,7 @@ use Data::Dumper;
 use Dancer2::FileUtils qw/path open_file/;
 use Sub::Quote;
 use Module::Runtime 'require_module';
+use Scalar::Util 'reftype';
 
 has app => (
     is        => 'ro',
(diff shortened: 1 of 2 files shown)


jkeenan:
While working my way through Dancer2::Manual, I encountered errors which dumped my config.yml and session data into the browser, thereby exposing my password.  Investigation suggested that Dancer2::Core::Error::_censor() was at fault.  Most parts of that subroutine were also unexercised by the test suite.  So I revised _censor() and added a test block with dummy data worked up from the error output I encountered earlier in the day.  The additional tests may not be in approved Dancer2 style, but they get the job done.  Test coverage of Dancer2::Core::Error has been improved as follows:
```
----------------------------------- ------ ------ ------ ------ ------ ------
File                                  stmt   bran   cond    sub   time  total
----------------------------------- ------ ------ ------ ------ ------ ------
before: lib/Dancer2/Core/Error.pm     96.8   82.3   72.2  100.0  100.0   89.5
after:  lib/Dancer2/Core/Error.pm     99.3   85.2   79.6  100.0  100.0   92.8
```

ambs:
It might be extra zealous in some situations, but I shouldn't hurt. For me, :+1:.

jkeenan:
On 07/08/2016 01:21 PM, Alberto Simões wrote:
> It might be extra zealous in some situations, but I shouldn't hurt. For me, :+1:.
>
>

Thanks for the feedback.

After creating the p.r., I got to thinking:  Isn't this a candidate for 
a stand-alone CPAN module?  It's buried deep inside Dancer not really 
Dancer-specific.

ambs:
Yes, that's true. I think that if you want to create one, nobody would oppose to start using it. But that is me speaking (not the core group).

jkeenan:
Okay, I might attempt that in a few weeks.  In the mean time, I'd like to see if this is an acceptable patch for Dancer2.

Thank you very much.
Jim Keenan

veryrusty:
@jkeenan Read through #530 before tackling that.

jkeenan:
Since there already exists https://github.com/PerlDancer/Dancer2/issues/530 with the same feature request, I will close this request.  I hope that some improved detection of sensitive data can be implemented soon.

ambs:
Let it open,until something is done. :-)

cromedome:
This is resolved in #1729. Closing.

PR #1209, opened 2016-07-08
Provide correct arguments for 'set views'.

jkeenan:
The documentation for 'Views' in Dancer2::Manual stated:

    You can change this location [for storing views] by changing the setting
    'views'.  For instance if your templates are located in the 'templates'
    directory, do the following:

        set views => path(dirname(__FILE__), 'templates');

However when I created a 'templates' directory, copied everything from the
existing 'views' directory thereto, entered the 'set views' line above into
'lib/myapp.pm' and attempted to login, I got a 500 runtime error, which boiled
down to:

    Can't open '.../mywebapp/lib/templates/login.tt'

In this example the file 'myapp.pm' is one directory beneath the top-level
directory.  To get from 'lib/myapp.pm' we have to go up to 'mywebapp' and then
down into 'templates'.  Hence, we should change 'templates' to '../templates'.
That worked for me.

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -1193,7 +1193,7 @@ You can change this location by changing the setting 'views'. For instance
 if your templates are located in the 'templates' directory, do the
 following:
 
-    set views => path(dirname(__FILE__), 'templates');
+    set views => path( app->location , 'templates' );
 
 By default, the internal template engine L<Dancer2::Template::Simple> is
 used, but you may want to upgrade to L<Template
@@ -1313,7 +1313,7 @@ template toolkit, your config file will look like this:
 Static files are served from the F<./public> directory. You can specify a
 different location by setting the C<public_dir> option:
 
-    set public_dir => path(dirname(__FILE__), 'static');
+    set public_dir => path( app->location , 'static' );
 
 When you modify default public_dir you have to set C<static_handler> option.
 


veryrusty:
Thanks for the PR @jkeenan.  If we are going to tweak this example, it should be to something that "always" works, no matter how far down the directory structure the source file was. 

eg  `set views => path( app->location , 'templates' );`

jkeenan:
@veryrusty , Thanks; I have updated the pull request with your suggested code.

jkeenan:
@veryrusty, Should we also change this line of documentation in the same way?
```perl
set public_dir => path(dirname(__FILE__), 'static');
```
See, e.g., https://metacpan.org/source/XSAWYERX/Dancer2-0.200002/lib/Dancer2/Manual.pod#L1323

Thank you very much.
Jim Keenan

racke:
On 07/09/2016 08:42 PM, James E Keenan wrote:
> @veryrusty <https://github.com/veryrusty>, Should we also change this
> line of documentation in the same way?
> 
> set public_dir => path(dirname(__FILE__), 'static');
> 
> See, e.g.,
> https://metacpan.org/source/XSAWYERX/Dancer2-0.200002/lib/Dancer2/Manual.pod#L1323
> 

Yes that makes sense. Thanks for your useful contributions!

Regards
        Racke

jkeenan:
Revised p.r. to handle that second case.  I think the p.r. should be ready to go now.  Thanks.

racke:
Looks good to me :+1:

veryrusty:
:+1: @jkeenan++ Thanks! Merged as 4e7a1e9.

PR #1210, opened 2016-07-11
Fix logging shows wrong linenumber

masakyst:
hello!  dancer2 is very cool!!

Change line numbers in stack trace

package Simple;
use Dancer2;
get '/' => sub {
    info "hello!!"; # <- line number
    return 'index';
};

version 0.200002
Dancer2/Core/App.pm l. 840

version 0.166001
/Simple/bin/../lib/Simple.pm l. 7

thank you!

### DIFF
--- a/lib/Dancer2/Core/Role/Logger.pm
+++ b/lib/Dancer2/Core/Role/Logger.pm
@@ -74,7 +74,7 @@ sub format_message {
     $message = Encode::encode( $self->auto_encoding_charset, $message )
       if $self->auto_encoding_charset;
 
-    my @stack = caller(5);
+    my @stack = caller(8);
     my $request = $self->request;
     my $config = $self->config;
 


SysPete:
This certainly fixes things for some debug logs from my app...

was:
```
[IIP:31263] debug @2016-07-20 18:59:23> random debug in map route in /home/apm/perl5/perlbrew/perls/perl-5.16.2/lib/site_perl/5.16.2/Dancer2/Core/App.pm l. 840
```
now:
```
[IIP:31036] debug @2016-07-20 18:55:15> random debug in map route in /home/apm/camp31/IIP/bin/../lib/IIP/Routes/Places.pm l. 106
```
but still plenty of useless stuff:
```
[IIP:31035] core @2016-07-20 18:55:16> Entering hook core.app.after_file_render in (eval 218) l. 1
```
For that last one going back 6 stack frames instead of 8 gets me:
```
[IIP:31578] core @2016-07-20 19:02:50> Entering hook core.app.before_file_render in /home/apm/Dancer2/lib/Dancer2/Core/Role/Hookable.pm l. 111
```
which is an improvement in that case.

So does anyone know where the original magic number came from? Maybe we need to find stack depth and work backwards or perform some other voodoo?

SysPete:
@masakyst I agree this PR is a definite improvement for logging data emitted from a route which is the most common place for simpler app so I'd say :+1: for merging this. I also think we should consider opening a new issue about this since setting the stack frame in this way does not always catch the correct file/line which is very bad.

cromedome:
@SysPete and I just had a chat about this. This PR improves the current situation, but is an indication that some more work is needed.

I think, given that, I am :+1: for the PR, and let's open a new issue to cover the rest.

Thanks for the PR!

racke:
Seems to be difficult to determine the stack level, it would be good if we add a comment  and this place.
But :+1: for now.

cromedome:
Merged. Thanks for the contribution!

masakyst:
Thank you so much!!

Dancer2-0.200004 :+1:

Issue #900, opened 2015-05-04
Are we sure we want to_json vs encode_json?


ambs:
From the documentation of JSON module, `to_json` keyword:

````````
If you want to write a modern perl code which communicates to outer world, you should use encode_json (supposed that JSON data are encoded in UTF-8).
````````
and I can't foresee a non UTF-8 world out there.

logust79:
I don’t know if this is relevant, but there are many Chinese websites with non UTF-8 encoding.

ambs:
Also, after digging it up a little, I found the utf8 flag on to_json 
disturbing. Needed to turn it off to output UTF-8 data.

Something is not clear in the way these modules work. At least for me.

cromedome:
Closed by PR #1212. Thanks @nunorc!

Issue #1196, opened 2016-07-04
Should we move to `JSON::MaybeXS`?


xsawyerx:
We currently use [JSON](https://metacpan.org/pod/JSON), but it both has issues and is in pure-Perl. [JSON::XS](https://metacpan.org/pod/JSON::XS) is useful but it's only XS and the author is... well, yes. [JSON::MaybeXS](https://metacpan.org/pod/JSON::MaybeXS) is both optionally XS (which means users with XS will have it installed automatically) and first attempts to load the cPanel JSON::XS implementation, which has a better maintainer, IMHO.

This means also recommending the cPanel variation of JSON::XS, which I would honestly prefer.

cromedome:
I’ve heard nothing but good things about Cpanel::JSON::XS. I’d have no problem making these changes. +1.

SysPete:
:+1: JSON::MaybeXS does seem to be the way to go. I've had to switch some modules over already due to the known JSON::XS issues.

racke:
:+1:  Please go ahead.

ambs:
Changed labels. If anyone start working on this, please warn here ;-)

nunorc:
PR https://github.com/PerlDancer/Dancer2/pull/1212 is a tentative pull request to move serializer to JSON::MaybeXS. Comments and feedback are welcome. Thank you.

cromedome:
Closed by PR #1212. Thanks @nunorc!

PR #1212, opened 2016-07-11
Move serializer from JSON to JSON::MaybeXS

nunorc:
Tentative pull request to address issue https://github.com/PerlDancer/Dancer2/issues/1196

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -89,7 +89,7 @@ Plack::Middleware::RemoveRedundantBody = 0
 ; JSON is just a serialiser, so in theory should be optional.
 ; But it is used in the DSL (send_error) and in tests.
 ; And also so much used in apps...
-JSON = 0
+JSON::MaybeXS = 0
 ; Used in DSL (send_error)
 URI::Escape = 0
 parent = 0
@@ -109,7 +109,7 @@ YAML = >= 0.86, != 1.16
 
 [Prereqs / Recommends]
 ; Serializers
-JSON::XS = 0
+Cpanel::JSON::XS = 0
 ; Developers are expected to write tests and so use Dancer2::Test
 ; that depends on Test::More. So Test::More is needed for more than
 ; just tests of the Dancer distribution
--- a/lib/Dancer2/Serializer/JSON.pm
+++ b/lib/Dancer2/Serializer/JSON.pm
@@ -2,7 +2,7 @@ package Dancer2::Serializer::JSON;
 # ABSTRACT: Serializer for handling JSON data
 
 use Moo;
-use JSON ();
+use JSON::MaybeXS ();
 use Scalar::Util 'blessed';
 
 with 'Dancer2::Core::Role::Serializer';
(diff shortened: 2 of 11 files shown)


xsawyerx:
:+1:

veryrusty:
Looks good! @nunorc++

This also address #900; The only places left using `to_json` are the tests. @nunorc could you modify those to use `encode_json` as well ?

cromedome:
Good work, @nunorc! 

I would like to point out that one test is failing:
```
t/plugin_syntax.t ......................................
    #   Failed test 'plugin_setting returned the expected config'
    #   at t/plugin_syntax.t line 46.
    #          got: '{"plugin":"42"}'
    #     expected: '{"plugin":42}'
    # Looks like you failed 1 test of 5.
t/plugin_syntax.t ...................................... 1/?
#   Failed test 'global and route keywords'
#   at t/plugin_syntax.t line 58.
# Looks like you failed 1 test of 4.
t/plugin_syntax.t ...................................... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/4 subtests
```
If you can fix the test, and take @veryrusty's suggestion, I will get this merged ASAP for you.

Thanks so much!

nunorc:
@cromedome thank you for your comment, the test from `t/plugin_syntax.t` should be passing after commit https://github.com/PerlDancer/Dancer2/pull/1212/commits/10bc227f32c235bd8ad4c6bd91ff0c6e72e98c95 (actually the problem was in the test itself) does it still fails for you?

veryrusty:
@cromedome @nunorc that is a somewhat "tricky" test. The source is loaded from `t/config.yml`. YAML only defines "scalars" and leaves it up to the application for "type" handling. JSON is much more picky about types. Perl is, well, perl. JSON implementations often looks at last context a scalar was used to determine is that 42 should be a string or an integer.  Its messy.

As it stands, its a regression that someone _may_ be relying upon. I haven't had a chance to track down what's going on. We may want to update one of the routes involved to ensure we explicitly numify that 42.

nunorc:
@veryrusty commit https://github.com/PerlDancer/Dancer2/pull/1212/commits/7486abbd869c876696cee5b9b8e24ba30778b53d should address issue https://github.com/PerlDancer/Dancer2/issues/900, all the `to_json` calls to `JSON::MaybeXS` where updated to use `encode` or `encode_json`.

nunorc:
@veryrusty @cromedome yes, I agree that the test in `t/plugin_syntax.t` is tricky, the configuration is loaded using `Config::Any` that uses `YAML::XS` or `YAML::Syck` to actually read the YAML file, and even these two modules have different behaviours. Given that later in the stack, Perl with probably deal ok with either way 42 (or any other config value) is stored as a string or an int, I think we should improve the test itself. Commit https://github.com/PerlDancer/Dancer2/pull/1212/commits/6217ac212f7260145046890e8c3b6df3548e209f uses a _normalize_ function to standardize _number_ values. This makes sure that if the configuration was read correctly (whatever YAML parser was used) the test succeeds. Comments are welcome.

cromedome:
Merged. Closes #900 and #1196. Thanks so much!

Issue #1216, opened 2016-07-18
[Dancer2 0.200003] Weird behaviour with Plugin Inclusion


pierre-vigier:
Hi,

i got an app that was working on Dancer 0.166001 and is breaking with 0.200003 .
Here is a trimmed down version of the problem. Once including Dancer2::Plugin::Database , i get a `Can't locate object method "dsl" via package "DemoApp"` error message

* bin/app.psgi
```perl
#!/usr/bin/env perl
use strict;
use warnings;
use lib 'lib';
use DemoApp;

return DemoApp->to_app;
```

*lib/DemoApp.pm
```perl
package DemoApp;
use DemoApp::Extension;
use Dancer2;

get '/' => sub { return 'root' };

true;
```

*lib/DemoApp/Extension.pm
```perl
package DemoApp::Extension;
use Dancer2 appname => 'DemoApp';
use Dancer2::Plugin::Database;

get '/extension' => sub { 'extension'; };
```

That code runs on Dancer2 0.166001, but not on 0.200003.

Fix is pretty simple, in my code, if i reverse in `lib/DemoApp.pm` the inclusion of Dancer2 and DemoApp::Extension, it is then working as expected:
```perl
package DemoApp;
use Dancer2;
use DemoApp::Extension;
```
instead of
```perl
package DemoApp;
use DemoApp::Extension;
use Dancer2;
```

It might be worth a hint in documentation for people upgrading from a version below 0.200001 to 0.200001 and above.

The code is working properly in 0.200001 and above if in my case the DemoApp::Extension package do not include plugin Dancer2::Plugin::Database .
I can't tell if it's a bug of Dancer2 or of my code, but it's a breaking change. That's happening with other plugin as well, like Dancer2::Plugin::Email.

xsawyerx:
What version of the plugins are you using?

pierre-vigier:
Should be the last ones,
Dancer2::Plugin::Database => 2.16
Dancer2::Plugin::Email => 0.003
```
> perl -I local/lib/perl5 -MDancer2::Plugin::Database -E 'say $Dancer2::Plugin::Database::VERSION;'
2.16
>perl -I local/lib/perl5 -MDancer2::Plugin::Email -E 'say $Dancer2::Plugin::Email::VERSION;'
0.0003
```

xsawyerx:
Preliminary debugging shows that the problem is due to the order here:

* You load `DemoApp`
[...]

PR #1217, opened 2016-07-18
Call DSL from plugin on the consumer class

xsawyerx:
The problem in GH #1216 is when a user loads a plugin which tries to access a DSL, while only loading "Dancer2" (and thus instating the DSL) comes after.

In that case, the DSL in the app class does not exist yet, and thus fails.

Instead of calling the app name as a class, use the first class in the caller tree that implements 'dsl', which is more likely to be the consumer of the plugin, even if under a different class.


veryrusty:
:+1:

cromedome:
:+1:

xsawyerx:
Merged.

Issue #1192, opened 2016-06-20
Encoding of body_params vs param/params


patzim:
- Dancer2 v0.200001
- _config.yml_ has `charset: "UTF-8"` set.
- When doing

        my @params = (Empfänger => '毛泽东');
        $user_agent->post(
            $url,
            'Content-type' => "application/x-www-form-urlencoded; charset='utf8'",
            Content      => [ map { Encode::encode_utf8( $_ ) } @params ],
        );

- `params` returns

        { Empfänger => '毛泽东' }

- and `body_parameters` returns

        { EmpfÃ¤nger => 'æ¯æ³½ä¸' }

- When doing a `map { Encode::decode( "utf-8", $_) }` on that I get the decoded version.

I tracked `params` down to Request.pm:312: `$self->_params` which is initialized in line 460 with `$self->_body_params` wich does a `_decode` in line 125.

And `body_parameters` in _Request.pm:368_: `return $self->env->{'plack.request.body'}`.

I am rather unskilled with providing good information. So if I should provide more (which?) information, please say so!

SysPete:
Looking at code and tests (t/dsl/parameters.t) this actually looks intentional though I have not (yet) found anything in the docs that mentions this difference.

@PerlDancer (especially @xsawyerx) I've been hunting through the various issues and PR #988 but I can't find any discussion or explanation for this particular difference between params and parameters. Can anyone help enlighten us?

SysPete:
@patzim many thanks for spending the time to submit this detailed issue. I've created PR #1219 which addresses it.

cromedome:
Fixed in 1d62e519017faed3ae0e22ab2578592f46795c30. Closing.

PR #1219, opened 2016-07-22
body|query|request_parameters should be decoded

SysPete:
See issue #1192.

@veryrusty said: I suspect the param encoding change was a regression.  When the request object became a subclass of Plack::Request, we used the existing body_params methods from Plack::Request.  They do not decode params. Whereas Dancer(2) always has decoded params.  Decoding *is* the correct thing to be doing (IMHO).


cromedome:
:+1:

racke:
:+1:

xsawyerx:
:+1:

cromedome:
Merged! 1d62e519017faed3ae0e22ab2578592f46795c30

Issue #1180, opened 2016-06-04
Type::Tiny::XS crash on Strawberry Perl 5.22


hurricup:
Just FYI. Bug reported to developer: https://github.com/tobyink/p5-type-tiny-xs/issues/4

xsawyerx:
@PerlDancer, should we revert Type::Tiny changes for now?

cromedome:
As there seems to be some ongoing issues with it, I'd say we should.

SysPete:
Does anyone have any news about tobyink or a possible fork & new maintainer? If it is likely that something might happen soon I'd prefer to avoid the revert since it adds work elsewhere. If there is nothing on the horizon then I agree that a revert is the sane option.

veryrusty:
We've reverted the change to use Type::Tiny in #1220.  (Thanks again @SysPete).  Dancer2::Types uses MooX::Type::MooseLike again until the current Type::Tiny maintenance issues are resolved.

SysPete:
Closing this issue now that Dancer v0.202000 has been released which includes #1220 which reverts back to using MooX::Types::MooseLike.

PR #1220, opened 2016-07-28
Revert type tiny

SysPete:
Revert feature/type_tiny so we use `MooX::Types::MooseLike` (see #1180)
    
Rather than using git revert which can do some terrible things to history this patch started as:
```    
gitt diff 88f9ef3d063d2cceb318403fecb424bd95544eca \
      5d3cdca3e53fb776767f8d9f507456fa333653ea
```
which was then cut back to something slightly saner (along with fixes where it did not apply).
    
I did not put back the inlined type checks that existed prior to the move to `Type::Tiny` since there is likely little (if any) performance gain from using hand-built inlined checks rather than `MooX::Types::MooseLike` type checks.

Extra tests added by the feature/type_tiny branch have been kept.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -59,7 +59,7 @@ match = dist.ini
 Moo = 2.000000
 Moo::Role = 0
 Role::Tiny = 2.000000
-Type::Tiny = 0.008
+MooX::Types::MooseLike = 0
 Carp = 0
 Digest::SHA = 0
 Exporter = 5.57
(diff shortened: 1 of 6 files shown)


SysPete:
I'd still prefer to see a fixed Type::Tiny but unless that happens soon...

cromedome:
:+1: Looks good, tests fine.

veryrusty:
Looks good. :+1:  I'll rebase and merge.

veryrusty:
@SysPete++ Thanks!

Merged as ba746256.

PR #1223, opened 2016-08-02
Add YAML::XS to Recommends.

SysPete:
If you are unfortunate enough to have YAML::Syck installed but not
YAML::XS then Config::Any will use YAML::Syck to load your configs. This
is all well and good as long as you have no UTF-8 config files which
YAML::Suck (stet) will merrily trash leaving you scratching your head
for hours as to why some UTF-8 glyph from the config has been double
decoded before reaching your webpage. Just don't do YAML::Syck. Please.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -117,6 +117,9 @@ Test::Builder = 0
 ; Extra speed with Request
 URL::Encode::XS = 0
 CGI::Deurl::XS = 0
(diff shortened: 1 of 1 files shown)


cromedome:
:+1:

cromedome:
Merged in 48c6a7db736455db5e84f6aa040302454210b593. Closing.

PR #1227, opened 2016-08-08
Allow plugins to access the DSL keywords

### LINKED ISSUE #1266: Add 'is_forward' key to request on forward
Perhaps 'is_internal_forward'? Thoughts appreciated. Will sort a PR after #1246 is merged.

xsawyerx:
This fixes #1266.

You will now be able to do this:

```perl
package Dancer2::Plugin::MyPlugin;
use Dancer2::Plugin;

sub my_thing {
    my $self = shift;

    # define a new GET route with the DSL methods:
    $self->dsl->get( '/' => sub {...} );
}
```

Please read the commit messages before approving. We will also need to document this, since it's a big (and important) change. It doesn't break, but it makes things much easier for plugin authors.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -447,6 +447,20 @@ END
                qw/ plugin_keywords plugin_hooks /;
 }
 
+sub _find_consumer {
+    my $class;
+
+    ## no critic qw(ControlStructures::ProhibitCStyleForLoops)
+    for ( my $i = 1; my $caller = caller($i); $i++ ) {
+        $class = $caller->can('dsl')
+            and last;
+    }
+
+    $class
+        or croak('Could not find Dancer2 app');
+
+    return $class;
+};
 
 # This has to be called for now at the end of every plugin package, in order to
 # map the keywords of the associated app to the plugin, so that these keywords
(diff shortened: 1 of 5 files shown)


xsawyerx:
This now produces the following warning:

```t/plugin_multiple_apps.t ............................... Subroutine Dancer2::Plugin::DancerPlugin::dsl redefined at /home/sawyer/code/personal/Dancer2/lib/Dancer2/Plugin.pm line 486.```

**Comment by xsawyerx** [comment]
Fixed.

File: lib/Dancer2/Plugin.pm [hunk 1]
```diff
@@ -481,6 +481,10 @@ sub register_plugin {
                 my $app_dsl_cb = _find_consumer();
                 my $dsl        = $app_dsl_cb->();
 
+                {
+                    no strict 'refs';
+                    *{"${plugin_module}::dsl"} = sub {$dsl};
```
veryrusty: We close over the DSL object fetched in line 482 above. There may be potential for a memory cycle.
(dsl => app => plugin => dsl arrangement).  I couldn't tickle that with the included tests in this Pr, but wondered if we should be defensive and weaken `$dsl` where its generated in line 482; viz
`Scalar::Util::weaken( my $dsl = $app_dsl_cb->() );`

veryrusty:
Looks ok. As noted in the inline comment; should we weaken the fetched `$dsl` object before closing over it ?

SysPete:
Yeah! @xsawyerx++. Calling dsl directly instead of hunting through code to find methods is a big help for plugin authors. A really nice addition.

xsawyerx:
The original fix here only allowed older plugins to have this feature. It was now fixed that everyone can have it. I'll take a look at the deref, but I wouldn't dereference it if there's no failure because it means we haven't really understood it.

xsawyerx:
This still does not fix another topic I realized is what we need at $work, which is plugins being able to call other plugins' defined DSL. I have a similar test for that and I'll work on adding that too, hopefully successfully. :)

xsawyerx:
@veryrusty, Okay I see this. For some reason the cyclic detection didn't pick up on it. I'm wondering if to debug this or not. Maybe it's because we're not closing over an object, but a callback.

xsawyerx:
Okay. Weakening breaks it. So, I guess it doesn't cycle there. :)

xsawyerx:
@PerlDancer, Can anyone double check this and give a fresh new :+1: to it?

Two to approve. :)

SysPete:
:+1: nice work @xsawyerx.

cromedome:
:+1:

cromedome:
Merged!

PR #1228, opened 2016-08-08
Allow register_plugin() to pass @_ properly:

xsawyerx:
This is inside an eval block, so I think it needs to be escaped
so it would be used as a literal.

Can someone approve this for me?


veryrusty:
:+1:

SysPete:
:+1:

cromedome:
Merged!

PR #1229, opened 2016-08-09
Bring the dealing with sub attributes in the BEGIN stage

### LINKED ISSUE #1224: :PluginKeyword sub attribute does not export
Using the plugin:

```
package Dancer2::Plugin::Example;

use 5.10.0;

use Dancer2::Plugin;

plugin_keywords foo => sub {  say "foo" };

sub bar :PluginKeyword {  say 'bar' }

1;

```

and the app:

```
use Dancer2;

use Dancer2::Plugin::Example;

get '/' => sub {
    foo;
    bar;
};

dance();
```

`foo` is correctly recognized, but `bar` is not.

yanick:
without that, the subroutines tagged by :PluginKeyword
are not being caught in time to be exported with the rest

Fixes #1224

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -139,7 +139,7 @@ sub _p2_has_keyword {
 
 # :PluginKeyword shenanigans
 
-sub PluginKeyword :ATTR(CODE) {
+sub PluginKeyword :ATTR(CODE,BEGIN) {
     my( $class, $sym_ref, $code, undef, $args ) = @_;
     my $func_name = *{$sym_ref}{NAME};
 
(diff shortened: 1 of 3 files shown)


SysPete:
5.10 breakage :unamused: bummer. The only ticket I've found so far that seems even vaguely related is this one:

https://rt.perl.org/Public/Bug/Display.html?id=40695

I've tested 5.10 with latest Attribute::Handlers from CPAN (0.99) but same result. Adding `no strict "refs"` stops 5.10 compilation failure but then the subroutine doesn't get exported (as per current release) so we end up with code that behaves differently depending on Perl version.

xsawyerx:
Can't we check for 5.10 and then not export that part? We already have two separate exporting statements there. (Both with `eval`.)

yanick:
I think I can fudge that. Patch coming shortly...

xsawyerx:
Bam! The Yanick does it again!

yanick:
In this case, "does it" consisted mostly of quietly pushing the 5.10 hairball under the carpet while nobody's looking. ;-)

SysPete:
Works for me hairball or no hairball and tests pass at least as far back as 5.8.9 so :+1: from me.

cromedome:
:+1: Nice work!

cromedome:
Merged, thank you!

Issue #1230, opened 2016-08-10
Plugins cannot access DSL from other plugins


xsawyerx:
This is an issue related to GH #1226, but in this case it's plugins accessing DSL by other plugins.

```perl
package MyApp {
    use Dancer2;
    use Dancer2::Plugin::Foo;
}

package Dancer2::Plugin::Foo {
    use Dancer2::Plugin;
    use Dancer2::Plugin::Bar;
    plugin_keywords('outer');
    sub outer { inner() }
}

package Dancer2::Plugin::Bar {
    use Dancer2::Plugin;
    plugin_keywords('inner');
    sub inner {1}
}
```

We have a Dancer App which loads a plugin which then loads another plugin. Both plugins have DSL keywords. Currently the second plugin (`Bar`) has a keyword (`inner`) which will never be exposed anywhere. It is not available to the App or to the first plugin (`Foo`).

cromedome:
Merged with 9775d0031248fd41999e3dcff2094e0d228638d3. Closing!

PR #1231, opened 2016-08-10
Allow import of plugins' DSL into other plugins

xsawyerx:
The reason plugins calling DSL on other plugins doesn't work is because we explicitly only allow importing plugins DSL from applications.

The way we verify this is by trying to call "app" attribute on the consuming class. Plugins don't have this. The change I made here is to check if "->dsl->app" is also available, which is available in plugins.

This fixed GH #1230, but assumes that GH #1227 was already merged.


cromedome:
:+1:

veryrusty:
:+1:

cromedome:
Merged with 9775d0031248fd41999e3dcff2094e0d228638d3. Closing!

Issue #1232, opened 2016-08-22
request->body_parameters empty when body contains serialized (JSON) data


veryrusty:
If a request contains serialized JSON, *and* some middleware wrapping the app created a Plack::Request object and called a method on that that causes the body to be read, then `$env->{'plack.request.body'}` is set to an empty Hash::MultiValue object (empty implying it has no keys).

Then within Dancer2, the logic to trigger deserializing of body params (https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Request.pm#L368) fails, as that env key already exists.

I'm working on a test case. ~~Not sure what the best fix is at the moment.~~

veryrusty:
Also; we spec'd the deserialization to return decoded (character) data. There is no need to decode that again (see https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Request.pm#L372 )

cromedome:
Resolved in 2f3579408e5a79ae25dc71dfb09b27ba96f123e9. Closing!

PR #1233, opened 2016-08-22
Bugfix/body data parse

veryrusty:
A logic error (described in #1232) made it possible for a request to
have serialized data that wouldn't be deserialized into body_parameters
if any middleware had previously created a Plack::Request object and
parsed the request body.

Instead, when a Dancer2::Core::Request object is instanciated, explicitly
call $self->data to force deserialization and (directly) set
env->{plack.request.body} to a HMV of that data (if its a hash). Setting
env->{plack.request.body} in this way is a bit messy; but we are
subclass Plack::Request so digging in internals is somewhat excused (for
now).

Also; deserialized data has already been decoded to characters (that how
we define our deserialization actions). A change in the 0.202000 release
was double decoding them.

Includes tests for both the deserialization (into body_parameters) and
checking for double decoding of utf8.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -79,7 +79,8 @@ sub new {
     $opts{'body_params'}
         and $self->{'_body_params'} = $opts{'body_params'};
 
-    # parsing body for HMV first (also deserializes body)
+    # Deserialize/parse body for HMV
+    $self->data;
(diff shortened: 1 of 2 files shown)


xsawyerx:
:+1:

SysPete:
:+1:  calling $self->data in new is much less messy than anything I can think of to sort this.

PR #1234, opened 2016-08-24
add code_mapping, status_mapping and all_mappings

yanick:
Add a few functions to allow code to access (copies) of our HTTP_CODES table. Mostly useful for plugins like Dancer2::Plugin::REST that won't have to copy the table for its own internal use.

### DIFF
--- a/lib/Dancer2/Core/HTTP.pm
+++ b/lib/Dancer2/Core/HTTP.pm
@@ -5,6 +5,8 @@ package Dancer2::Core::HTTP;
 use strict;
 use warnings;
 
+use List::Util qw/ pairmap pairgrep /;
+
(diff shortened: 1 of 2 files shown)


cromedome:
:+1::+1::+1: Does that count as three? ;) This is great!

veryrusty:
:+1:

cromedome:
Merged in as f7057e32e0c5ddf585d68d55f0ecda3e42803fd6. Closing!

xsawyerx:
I seem to recall us finding this in another module and wanting to use that one instead of our own.

I think it was [HTTP::Status](https://metacpan.org/pod/HTTP::Status).

PR #1236, opened 2016-08-26
retrieved sessions can have blessed hashrefs too

JaderDias:
There's no technical reason why a session couldn't have a
blessed hashref. So this commit modifies the check to accept
them too.

TODO: a session that has something other than a hashref
should not be simply ignored, but it should die or warn.

It shouldn't take a look at Dancer2 internals to figure
out why the data your Dancer2::Session engine is providing
is not made available to the Dancer2 app that is using it.

### DIFF
--- a/lib/Dancer2/Core/Role/SessionFactory.pm
+++ b/lib/Dancer2/Core/Role/SessionFactory.pm
@@ -11,6 +11,7 @@ use Digest::SHA 'sha1';
 use List::Util 'shuffle';
 use MIME::Base64 'encode_base64url';
 use Module::Runtime 'require_module';
+use Scalar::Util 'reftype';
 
 sub hook_aliases { +{} }
 sub supported_hooks {
@@ -163,7 +164,7 @@ sub retrieve {
     my %args = ( id => $id, );
 
     $args{data} = $data
-      if $data and ref $data eq 'HASH';
+      if ref $data and reftype $data eq 'HASH';
 
     $args{expires} = $self->cookie_duration
       if $self->has_cookie_duration;


yanick:
It would be nice to have a test to go with the patch, where an object cookie is (successfully) passed

lib/Dancer2/Core/Role/SessionFactory.pm
```diff
@@ -163,7 +164,7 @@ sub retrieve {
     my %args = ( id => $id, );
 
     $args{data} = $data
-      if $data and ref $data eq 'HASH';
+      if $data and ref $data and reftype $data eq 'HASH';
```
yanick: Could simply be 

```
if $data and reftype $data eq 'HASH'
```

yanick: I should look how/if the session data is serialized. Passing objects instead of pure hashes could make things break.

JaderDias: @yanick Is there any chance of $data being a scalar? My version avoids the following warning:
```
$ perl -MScalar::Util=reftype -we'my $data = 1; print "OK" if reftype $data eq "HASH"'
Use of uninitialized value in string eq at -e line 1.

```

xsawyerx:
Technically there's no reason why this can't be a blessed hashref, unless it's also created somewhere else, which it might be. I need to double check.

Meanwhile, the correct line would be: `ref $data && reftype($data) eq 'HASH'`. `ref` will already check if it's true *and* a ref. No need for `if $data`.

lib/Dancer2/Core/Role/SessionFactory.pm
yanick: True, that. I'm usually the kind of savage who does

```
no warnings 'uninitialized';
... if $foo eq 'bar';
```

instead of 
```
if $foo and $foo eq 'bar';
```

But, yeah, not everybody's cup of tea. :-)

JaderDias:
I updated the commit (push -f) to implement Sawyer's suggestion of not keeping the old check `if $data` since it was already unnecessary since it was always followed by `&& ref $data`

xsawyerx:
So here's what bothers me about this: The returned value should be the representation of the data. Returning an object is likely to be to have some tie magic on it in order in order to trigger fetching/saving when accessed/used.

The problem with this approach is that:

1. You cannot be certain it will be used inside the core the way you assume it will be.
2. This might be cached, might be refetched, etc. There are no certainties about that either.
3. This mixes actions (retrieve/save) with representation of data.

This is why I think it should remain a simple hashref.

Can you explain your use-case to me? Maybe I got this wrong.

JaderDias:
Actually this is a feature that prevented me from doing the wrong thing. I
wanted to return an object that would update the server side cookies values
when the request ended, and it turned out that other session engines don't
update the values unless you modify the hashref via the proper write
methods.

Now that I know this, I would like to retract this pull request.

Issue #1240, opened 2016-09-02
Some JSON requests are modified by Hash::MultiValue


bjornax:
I noticed that for some JSON body parameter schemas, arrays are sometimes simply removed when deserialized and stored in request->body_request()
For example, this JSON body parameter:
```javascript
{
  "foo": [ { "bar": 1 } ],
  "baz": { "foobar": [ { "blah": 2 } ] }
}
```
will be parsed into this in a route:
```perl
$VAR1 = {
          'foo' => {
                     'bar' => 1
                   }
          'baz' => {
                     'foobar' => [
                                   {
                                     'blah' => 2
                                   }
                                 ]
                   },
        };
```

Notice that the "foo" field is no longer an array after the deserialization, but the "foobar" field is as expected. I can reproduce this in both 0.202000 and 0.203000.

I tracked it down to the use of Hash::MultiValue->from_mixed(), and with this naive patch it works as expected for me:
```patch
--- Dancer2/Core/Request.pm.orig	2016-09-02 11:05:27.042773871 +0200
+++ Dancer2/Core/Request.pm	2016-09-02 11:05:35.714634529 +0200
@@ -213,8 +213,8 @@
 
     # Set body parameters (HMV)
     # Not happy with fiddling with Plack::Request internals -- veryrusty Aug 2016.
     $self->env->{'plack.request.body'} =
-        Hash::MultiValue->from_mixed( ref $data eq 'HASH' ? %$data : () );
+        Hash::MultiValue->new( ref $data eq 'HASH' ? %$data : () );
 
     return $data;
 }
```

I expect my "solution" to break more than it fixes, but I hope that you experts can come up with a better solution.

SysPete:
@bjornax many thanks for creating the issue. PR on its way.

SysPete:
Converting JSON (and other serialized data) to HMV seems like a bad move:
https://github.com/miyagawa/Hash-MultiValue/issues/11

@bjornax I'd suggest as a workaround until we've got this nailed down is to NOT use `body_parameters` but grab your data using the good old `params` keyword:
```
my $data = from_json(params('body'));
```

bjornax:
@SysPete ok. Thank you for the impressively quick initial response. A temporary workaround is no problem for me as I'm still prototyping. Looking forward to a permanent solution when ready.

PR #1241, opened 2016-09-02
Dont use HMV->from_mixed when deserializing JSON body data

SysPete:
This addresses #1240.

It appears that HMV->from_mixed trashes many first-level values that are array references but any deeper structures are left alone. Seems I misunderstood its workings when I first used it.

### DIFF
--- a/Changes
+++ b/Changes
@@ -4,6 +4,9 @@
     [ BUG FIXES ]
     * GH #1237: Specify minimum version of List::Util required for pair*
       functionals. (Russell @veryrusty Jenkins)
+    * GH #1240: JSON body deserialization should use HMV->new since
+      HMV->from_mixed does horrible things to first-level hash reference
+      values. (Peter Mottram - SysPete)
 
 0.203000  2016-08-24 22:09:56-05:00 America/Chicago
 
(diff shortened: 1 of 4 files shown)


veryrusty:
:+1:

SysPete:
So as it is this breaks things elsewhere :unamused: ...

https://travis-ci.org/PerlDancer/Dancer2/jobs/157060815#L1438-L1453

SysPete:
Perhaps some relevant discussion here:

https://github.com/miyagawa/Hash-MultiValue/issues/11

SysPete:
Looking further it seems that using HMV via `body_parameters` is just plain wrong for JSON and instead JSON should be extracted using something like `from_json(params('body')`. Problem is now we have those shiny new "preferred" methods.

SysPete:
I'm closing this for now since it breaks stuff elsewhere. We should move discussion back to #1240

bjornax:
I'd expect my naive patch to break with multi-valued form data. I strikes me as an ugly solution, but would it be possible to use HMV->new() for JSON data and HMV->from_mixed() for form data?

PR #1244, opened 2016-09-09
RFC: allow the code of a route to be a chain of subs

yanick:
It would make it a wee bit easier to segment the logic of routes in several parts. E.g.,

    use Dancer2;

    sub check_param {
        my $next = shift;

        return "nope" unless param 'secret';

        $next->();
    }

    get '/' => [ \&check_param, sub {
            'is good';
    } ];

### DIFF
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -4,7 +4,7 @@ package Dancer2::Core::Route;
 use Moo;
 use Dancer2::Core::Types;
 use Carp 'croak';
-use List::Util 'first';
+use List::Util 'first', 'reduce';
 use Scalar::Util 'blessed';
 
 our ( $REQUEST, $RESPONSE, $RESPONDER, $WRITER, $ERROR_HANDLER );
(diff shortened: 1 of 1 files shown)


pdl:
This takes up two 'slots' in the DSL: 

- The pattern http_method ( CODE , ARRAYREF )
- Arguably, the first value passed in

This seems of questionable value when you could trivially write a function (outside of Dancer) that implemented the behavior by returning a coderef that does the same thing from an array.

    get '/' => nest ( \&check_param, sub { 'is good' } );

If this becomes part of Dancer, it would seem to me more idiomatic to say:

    get '/' => [ sub { pass if param 'secret'; "nope" }, sub { 'is good' } ];

However, this means the first coderef does not regain control after the second has executed so might not work for other cases you might have in mind.

yanick:
Hmmm, you bring a good point with the `pass` keyword, I forgot about that.

yanick:
And 

 get '/' => [ sub { pass if param 'secret'; "nope" }, sub { 'is good' } ];

can also be written as 

 get '/' => $_ for sub { pass if param 'secret'; "nope" }, sub { 'is good' };

if you're super-evil. 

But yeah, using `pass` instead of `$next->()` seems to be much more idiosyncratic. I approve. 

In that case, I think it's probably better to stay with what we have. Thanks for being an awesome sounding board / voice of reason. :-)

Issue #1245, opened 2016-09-11
GET params not passed on properly with forward


b10m:
Considering an app like this:
```
get '/foo' => sub {
    return "Got: ".query_parameters->get('bar');
};

get '/foo/bar' => sub {
    forward '/foo', { bar => 'baz' };
};
```
You'd expect to see "Got: baz" when the /foo/bar URL was triggered, yet the `bar` parameter doesn't get added to the query_parameters (although the docs do suggest this is the preferred way of both adding params and reading params).

Using param() rather than query_parameters() works for now, but the additional param should simply be added to the query_parameters multivalue'd hash, IMHO.

xsawyerx:
I remember seeing something like that. I don't remember if we opened a ticket, fixed it in a branch, or what.

xsawyerx:
Ah, here we go. Related issue: #1116.

veryrusty:
If we consider `forward '/foo', { bar => 'baz' };` as being the same as `forward '/foo?bar=baz';`, that is treat those params as query params;  then we can make `request->query_parameters` and `request->params('query')` be consistent and include those new params.

b10m:
... and of course the same for `forward '/foo', { bar => 'baz' }, { method => 'POST' }` and `body_parameters`.

b10m:
FWIW, the [documentation](https://metacpan.org/pod/Dancer2::Manual#forward) suggests it does this (emphasis added):

> More formally; when forward is executed, the current dispatch of the route is aborted, the request is modified (**altering query params** or request method), and the modified request following a new route is dispatched again.

veryrusty:
I've started on resolving this (and #1116).  Commits are in the branch  `bugfix/hmv_forward`.

Any new params supplied to the `forward` get added to `params`, `query_params` and `query_parameters`, which should resolve this issue. (Yes, even if the method is a POST, the new params are *query* params).

For 1116; the HMV for plack body params from the current request is cloned into the new request.

No PR; as I haven't added test cases. Feel free to pick up that work and add tests :)

cromedome:
Resolved by 6dc4a48d5ef4a11cc0ff6a3defbc3f32acb23c73. Closing. Thanks @veryrusty!

PR #1246, opened 2016-09-19
Bugfix Hash::MultiValue params on forward 

### LINKED ISSUE #1116: /forward does not get body_parameters of original 'POST' request.
post '/' => sub {
    forward '/foo' if body_parameters->get('bar') eq 'foo';
...

post '/foo' => sub {
body_parameters() at this point is an empty Hash::MultiValue

The contents can still be reached, however, by using param():
my $bar = param('bar') ... though this is not the recommended way in the docs.

veryrusty:
To resolve #1116 and #1245, the code in `App->make_forward_to` needs to ensure the "cached" HMV params in the current request are cloned into the new request.

However it doesn't make sense for our App class to know about the Request class internals. Instead, this PR refactors a large chunk of `App->make_forward_to` into a private method on Request. Within this;
  * any `body_parameters` in the current request are copied into the new request.
  * any extra params supplied to forward get added to `params`, `query_params` and `query_parameters`;

Note that extra params to forward are always considered to be **query** parameters, even if the method is POST.  The test cases included check this and that #1116 and #1245 are resolved.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1288,46 +1288,18 @@ sub forward {
 
 # Create a new request which is a clone of the current one, apart
 # from the path location, which points instead to the new location
-# TODO this could be written in a more clean manner with a clone mechanism
 sub make_forward_to {
     my $self    = shift;
     my $url     = shift;
     my $params  = shift;
     my $options = shift;
 
-    my $request = $self->request;
-
-    # we clone the env to make sure we don't alter the existing one in $self
-    my $env = { %{ $request->env } };
-
-    $env->{PATH_INFO} = $url;
-
-    # request body fh has been read till end
-    # delete CONTENT_LENGTH in new request (no need to parse body again)
-    # and merge existing params
-    delete $env->{CONTENT_LENGTH};
-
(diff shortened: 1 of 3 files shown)


SysPete:
This all looks good to me :+1:

veryrusty:
There is some inter-dependency between this PR and #1269. We need to use our our stash for *decoded* body params. This PR needs updating to clone that stash once #1269 is merged.

veryrusty:
I've rebased this onto the current master; plus
  * tests now include utf8 characters
  * an extra commit to ensure the (shallow) clone only works with the *decoded* query and body parameters (which we now stash separately, see #1269)
 * also stash our merged, decoded HMV parameters in our request class, rather than the psgi env key Plack::Request used.  Mixing Plack::Request's (encoded) HMV with Dancer::Core::Request's (decoded) HMV is going to cause doubt, frustration, confusion and other emotions that lead to the dark side.

@SysPete you should have another look over this (since you previously have it a +1)

SysPete:
Looks good :+1:

cromedome:
Nice work. Really like how clean `make_forward_to()` has become as a result. :+1:

PR #1247, opened 2016-09-19
 Fix another edge case in plugin compat:

xsawyerx:
[If you can figure this out... `$diety` help you.]

Here's the thing. I don't know how to describe this problem. I
never seem to be in the zone enough to fully understand it or
explain it. However, I have these bits of gleaned information
which I quote below.

```
So, I think it's like this: If you have a plugin that uses
DSL directly (which are only available on D2P1, in D2P2 you
have to use `$self`⁠⁠), anyou didn't call `register_plugin`,
those will not be available yet, and when you try to call them,
it will fail that they do not exist.

(Which makes sense. We create them for you, with a warning,
if you call `⁠register_plug⁠`, which you would do for D2P1
plugin anyway.)

However, and here's the kicker, if you do this while defining
your `appname` to a class that does not exist, you get a
different error. The error you get is that the specified
`appname` class cannot call the keyword you wanted.

This is because there was one more place in which I didn't
resolve GH #1226.
```

Here, then, is the final fix. It should be to just use the found
"app_cb_dsl" class, but that broke a test. Ugh. Anyway, crazy,
but works. How? Not sure.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -523,7 +523,10 @@ sub register_plugin {
                     $plugin_module->can($keyword)
                       or *{"${plugin_module}::$keyword"} = sub {
                         $_[0]
-                          ? do { my $cc = shift()->app->name->can($keyword); $cc->(@_) }
+                          ? do {
+                            my $cb = shift()->app->name->can($keyword);
+                            $cb->(@_);
+                          }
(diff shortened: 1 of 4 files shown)


cromedome:
:+1: Can merge after rebase.

cromedome:
Merged, thanks!

PR #1250, opened 2016-09-21
Consistency for TemplateToolkit config

klaus:
Update skel config.yml to use the original Template::Toolkit
configuration keys and disable the ANYCASE config.

Use of the lower-case config tags will log a warning in debug mode.

This fixes: gh-1249


xsawyerx:
This was merged. I apologize for the delay.
Thanks! :)

PR #1251, opened 2016-09-24
added cpanfile

mickeyn:
Added a cpanfile to allow installing local dependencies with carton.
The list of modules is out of dist.ini + some modules used in tests ('skipped') added as author_requires.


SysPete:
Although I have no objection per se to having a cpanfile I'm concerned that it will end up going out of sync with dist.ini over time. Is there a way to have dzil use cpanfile sanely instead of what it in dist.ini or have cpanfile generated by dzil?

mickeyn:
https://v1.metacpan.org/pod/Dist::Zilla::Plugin::Prereqs::FromCPANfile

mickeyn:
patched. I avoided touching dist.ini as it needs to be tested to ensure this doesn't break the distribution process, but I totally agree with you.

SysPete:
@mickeyn I like this a lot. I especially like being able to install deps quickly in the repo without waiting eons for dzil.

xsawyerx:
:+1: Yes, please.

racke:
Nice work :+1: . Any idea why travis check failed?

mickeyn:
it reported this line to fail: `The command "dzil listdeps --author --missing | cpanm --no-skip-satisfied || { cat ~/.cpanm/build.log ; false ; }" failed and exited with 1 during .`
but i can't recreate it, it works on my machine.

racke:
On 10/08/2016 01:25 PM, Mickey wrote:
> it reported this line to fail: |The command "dzil listdeps --author --missing | cpanm --no-skip-satisfied || { cat
> ~/.cpanm/build.log ; false ; }" failed and exited with 1 during .|
> but i can't recreate it, it works on my machine.
> 

I'm not surprised about random failures of dzil :-;, so I'm fine with merging that PR.

Regards
         Racke


-- 
Ecommerce and Linux consulting + Perl and web application programming.
Debian and Sympa administration.

mickeyn:
ok, just fixed a small bug that broke parsing... let's wait till the next travis run ends

SysPete:
Looks good :+1:

PR #1254, opened 2016-09-27
One should use get_all() to get multiple values.

bigpresh:
Nod to tushar on #dancer who was having trouble and lead me to spot this.

Hash::MultiValue's docs for get() say: "Note that this always returns the single
element as a scalar, regardless of its context".


xsawyerx:
Whoa! Big "whoops" here.

:+1: Please merge. :)

xsawyerx:
Or I can merge.

PR #1256, opened 2016-10-03
Several Additions to Migration Manual

pdl:
Mostly things I've run into while migrating to Dancer2.

### DIFF
--- a/lib/Dancer2/Manual/Migration.pod
+++ b/lib/Dancer2/Manual/Migration.pod
@@ -508,7 +508,8 @@ Then in other packages:
 
 =head3 appdir
 
-This keyword does not exist in Dancer2.
+This keyword does not exist in Dancer2. However, the same information can be
+found in C<< config->{'appdir'} >>.
 
 =head3 load
 
--- a/lib/Dancer2/Manual/Migration.pod
+++ b/lib/Dancer2/Manual/Migration.pod
@@ -434,6 +434,13 @@ for these options, or the particular session engine for other supported options.
 	    is_secure: 1
         is_http_only: 1
 
+=head3 Plack Middleware
+
+In Dancer1 you could set up Plack Middleware using a C<plack_middlewares> key
+in your C<config.yml> file. Dancer2 will not respect this. Instead, invoke
+middleware in your C<app.psgi> as demonstrated in
+L<Dancer2::Manual::Deployment>.
+
 =head2 Keywords
 
 =head3 load
--- a/lib/Dancer2/Manual/Migration.pod
+++ b/lib/Dancer2/Manual/Migration.pod
@@ -477,7 +477,7 @@ In Dancer1, keywords could be imported individually into a package:
 
     get '/foo' { ... };
 
-Keywords could be called explicitly:
+Any keywords you did't export could be called explicitly:
 
     package MyApp;
     use Dancer qw< get post params session >;
@@ -485,9 +485,11 @@ Keywords could be called explicitly:
 
     Dancer::any sub { ... };
 
-Dancer2's DSL is implemented differently, and neither of these are possible
-directly. If you need this ability, you can write a package which provides
-this as follows:
+Dancer2's DSL is implemented differently. Keywords only exist in the namespace
+of the package which C<use>s Dancer2, i.e. there is no C<Dancer2::any>, only
+e.g. C<MyApp::any>.
+
+If you only want individual keywords, you can write a shim as follows:
 
     package MyApp::DSL;
     use Dancer2 appname => 'MyApp';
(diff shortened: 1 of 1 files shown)


lib/Dancer2/Manual/Migration.pod
```diff
@@ -500,4 +569,15 @@ delete:
...
+
+    session->destroy
+
+In Dancer2, we write:
+
+    app->destroy_session if app->has_session
+
+If you make changes to the session in an C<after> hook, your changes will
+not be written to storage, because writing sessions to storage also takes
+place in an (earlier) C<after> hook.
```
xsawyerx: I'm wondering if this is not a bug.

lib/Dancer2/Manual/Migration.pod
```diff
@@ -298,6 +300,26 @@ function with decoded input.
 
 =back
 
+=head3 Templating engine changes
+
+Whereas in Dancer1 it was acceptable to write:
+
+    template '/foo/bar'
+
+In Dancer2, the leading slash is not permitted:
```
pdl: @xsawyerx - looks like it's both.

D1 simple: `/index.tt` treated as `APPDIR/views/index.tt`.
D1 template_toolkit: `/index.tt` treated as `APPDIR/views/index.tt`.
D2 simple: `/index.tt` treated as `APPDIR/views/index.tt`.
D2 template_toolkit: `/index.tt` behaves differently - looks like it's looking for `/index.tt`, i.e. at the filesystem root.

I guess 'not permitted' is not the right phrase. But I think we should flag this change (assuming it's deliberate), even if it's strictly not Dancer Core but the TT2 plugin.

lib/Dancer2/Manual/Migration.pod
```diff
@@ -432,8 +454,55 @@ for these options, or the particular session engine for other supported options.
 	    is_secure: 1
         is_http_only: 1
 
+=head3 Plack Middleware
+
+In Dancer1 you could set up Plack Middleware using a C<plack_middlewares> key
+in your C<config.yml> file. Dancer2 will not respect this. Instead, invoke
+middleware in your C<app.psgi> as demonstrated in
```
pdl: @xsawyerx Yes, that's clearer, thanks.

lib/Dancer2/Manual/Migration.pod
```diff
@@ -432,8 +454,55 @@ for these options, or the particular session engine for other supported options.
...
+Dancer2's DSL is implemented differently, and neither of these are possible
+directly. If you need this ability, you can write a package which provides
+this as follows:
+
+    package MyApp::DSL;
+    use Dancer2 appname => 'MyApp';
+
+    use Exporter qw< import >;
+
+    our @EXPORT = qw< get post ... >
```
pdl: Works for me.

lib/Dancer2/Manual/Migration.pod
pdl: I had wondered that, too!

lib/Dancer2/Manual/Migration.pod
```diff
@@ -298,6 +300,26 @@ function with decoded input.
...
+
+Whereas in Dancer1, template engines have the methods:
+
+    $template_engine->view('foo.tt')
+    $template_engine->view_exists('foo.tt')
+
+In Dancer2, you should instead write:
+
+    $template_engine->view_pathname('foo.tt')
+    $template_engine->pathname_exists($full_path)
```
pdl: I think it's probably fine to put it like that although strictly code where I found this is not writing a templating engine, it's just using TT2. Admittedly, it's wrapping `template` - but all it's really doing is checking to see if the view exists and if not, trying to find another view which does.

pdl:
I've made most of the changes suggested. You write:

> I can see the session factory _does_ provides a `destroy` method.

I'm not sure if you mean that a) it's not an issue, b) a fix is possible, or c) something else. For me, a bare app with `session->destroy` fails.

SysPete:
@pdl you are correct: `session->destroy` is not possible and `app->destroy_session` is the correct way. The session object cannot destroy itself since destruction is actually performed by the app's `session_engine` which the individual `session` object has no knowledge of.

For me this looks good so :+1:

cromedome:
Thanks @pdl, this is great! :+1:

cromedome:
Merged and I forgot to close this. Oops. Thanks again!

pdl:
@cromedome this was closed but not merged. Perhaps you were confused by a reference from a PR in the advent-calendar repo?

cromedome:
That could be. Sorry for the mix up!

cromedome:
Merged (for real this time!). Thanks!

PR #1257, opened 2016-10-04
Add support for shimming eval and better error handling

demerphq:
At $work we use $SIG{__DIE__} handlers to ensure all
of our errors have backtraces from the correct context, and
we perform other forms of stack backtrace munging for security
reasons as well.

For this to be effective and efficient we need to know how many
"wrapper evals" that are being used by the infrastructure layer
of our webservers.

Dancer imposes a number of these evals, and possibly at varying
depths of the call stack, and therefor we need a way to inject
scoped code into each eval. The simplest way to do that is to
have a coderef shim which is used to execute the code inside
of the eval BLOCK's that dancer uses.

Additionally, Dancer is supposed to run on Perl's that have the
"clobbered $@ bug", so this patch also corrects said logic to not
depend on $@ being true after a failed eval (which it might not
on older perls, or in some bizarre cases in more modern ones.)

The basic pattern is that code like:

    eval { require_module( $serializer_class ); 1; } or
        croak "Unable to load serializer class for $type";

turns into

     our $EVAL_SHIM = sub {
        my $code = shift;
        return $code->(@_);
     };

     eval {
         $EVAL_SHIM->(sub {
             require_module( $serializer_class );
         });
         1;
     } or do {
         my $err = $@ || "Zombie Error";
         croak "Unable to load serializer class for $type: $err";
     };

so that $EVAL_SHIM can be overriden in situ to provide more complex
logic. For instance at $work we would do this:

     $Dancer2::Core::App::EVAL_SHIM = sub {
        my $code = shift;
        local $EVAL_COUNTER = $EVAL_COUNTER + 1;
        return $code->(@_);
     };

So that our $SIG{__DIE__} handler can check to see how many evals
we are expecting from the infrastructure layer so that we can tell
if we are inside of a *true* try/catch or not (if we are we don't
munge the stack).

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -27,6 +27,12 @@ use Dancer2::Core::Factory;
 
 use Dancer2::Handler::File;
 
+our $EVAL_SHIM = sub {
+    my $code = shift;
+    $code->(@_);
+};
+
+
 # we have hooks here
 with qw<
     Dancer2::Core::Role::Hookable
@@ -424,9 +430,20 @@ sub _build_session {
 
         # if we have a session cookie, try to retrieve the session
         if ( defined $session_id ) {
-            eval  { $session = $engine->retrieve( id => $session_id ); 1; }
-            or do { $@ and $@ !~ /Unable to retrieve session/
-                        and croak "Fail to retrieve session: $@" };
+            eval  {
+                $EVAL_SHIM->(sub {
+                    $session = $engine->retrieve( id => $session_id );
+                });
+                1;
+            }
+            or do {
+                my $err = $@ || "Zombie Error";
+                if ( $err !~ /Unable to retrieve session/ ) {
+                    croak "Failed to retrieve session: $err"
+                } else {
+                    # XXX we throw away the error entirely? Why?
+                }
+            };
         }
     }
 
@@ -917,8 +934,15 @@ sub send_as {
 
     # Try and load the serializer class
     my $serializer_class = "Dancer2::Serializer::$type";
-    eval { require_module( $serializer_class ); 1; } or
-        croak "Unable to load serializer class for $type";
+    eval {
+        $EVAL_SHIM->(sub {
+            require_module( $serializer_class );
+        });
+        1;
+    } or do {
+        my $err = $@ || "Zombie Error";
+        croak "Unable to load serializer class for $type: $err";
+    };
 
     # load any serializer engine config
     my $engine_options =
@@ -1116,11 +1140,12 @@ sub compile_hooks {
                 $Dancer2::Core::Route::RESPONSE->is_halted
                     and return;
 
-                eval  { $hook->(@_); 1; }
+                eval  { $EVAL_SHIM->($hook,@_); 1; }
                 or do {
+                    my $err = $@ || "Zombie Error";
                     $app->cleanup;
-                    $app->log('error', "Exception caught in '$position' filter: $@");
-                    croak "Exception caught in '$position' filter: $@";
+                    $app->log('error', "Exception caught in '$position' filter: $err");
+                    croak "Exception caught in '$position' filter: $err";
                 };
             };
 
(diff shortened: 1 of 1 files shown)


xsawyerx:
:+1:

demerphq:
No problem. Is there a reason you haven't merged it? 

(And the other 25 PR's waiting? I have to admit I was thinking of filing a bug "PR's not being applied". :-) Need any help integrating them? )

Yves

veryrusty:
@demerphq "policy" is for two core devs to give approval before merge.  I'll be looking at it later in the day.

demerphq:
Thanks for explaining.

xsawyerx:
I commented before going to bed. We needed another :+1: to fully accept it.

xsawyerx:
I double checked tests, added an entry to `Changes`, and merged.
Thanks, @demerphq.

veryrusty:
Hmmm, I actually had some questions about this :(

@demerphq The PR as is was fine. Thanks! Benchmarks showed no negative impact either.

What I didn't grok was *why* its only the `evals` in the App class that need shimming. If the handful of other evals in the code need the same treatment, should the shim live in the `Dancer2` namespace rather than in the Dancer2::Core::App namespace as its a GLOBAL setting.

I also wondered about ongoing maintenance. We'll need to ensure any evals added in the future get wrapped by the shim as contributors may not be aware of the construct. If a try/catch approach eliminates the need for the shim, we may want to revisit options there.

xsawyerx:
Whoops. I misread and rushed. I reverted it meanwhile. It can always be rereverted as soon as we're satisfied or fixed up and merged after that.

demerphq:
On 5 October 2016 at 11:59, Russell Jenkins <notifications@github.com>
wrote:

> Hmmm, I actually had some questions about this :(
>
> @demerphq <https://github.com/demerphq> The PR as is was fine. Thanks!
> Benchmarks showed no negative impact either.
>
> What I didn't grok was *why* its only the evals in the App class that
> need shimming. If the handful of other evals in the code need the same
> treatment, should the shim live in the Dancer2 namespace rather than in
> the Dancer2::Core::App namespace as its a GLOBAL setting.
>
>
It depends what the eval is doing.

AFAIUI Dancer2::Core::App does dispatch, and is the place where Dancer2
gracefully converts handler errors to a human readable error page.

So the eval's that App does are not try/catch in the formal sense that we
intend to catch a particular error and "handle it" somehow appropriately.
Instead they are using eval to prevent the webserver being broken by a
broken handler. So errors that are thrown inside of those evals need to be
countable.

On the other hand a try/catch that is intended to truly catch an error and
gracefully recover without showing the error to a developer do not need the
shim.


> I also wondered about ongoing maintenance. We'll need to ensure any evals
> added in the future get wrapped by the shim as contributors may not be
> aware of the construct.
>
I think the shim is only needed in code that is related to dispatch or at
the top of the dancer2 stack. Stuff lower down probably does not need it.
However I may be correct.

> If a try/catch approach eliminates the need for the shim, we may want to
> revisit options there.
>
I guess you could implement you own try/catch layer that implements the
shim in a shared way so that all the dancer2 code would use your try/catch
layer.

However I would like to point out that there are *very* subtle effects
involved in this patch and sequence, and IMO you are unlikely to be able to
achieve the same results using a try/catch layer that does not ALSO do the
shim.

So one of the points is that we need to use local here to correctly manage
unwinding the counter at exception time. A simple try/catch implementation
is not going to facilitate this. Similarly you cannot do what is needed in
the catch logic of a try/catch pattern. It must happen *inside* the eval,
which is where the shim comes in. Essentially we need to inject code into
the beginning of an eval BLOCK.

Another point here is how would a user like $work override the behavior of
a try/catch sub? Most of the standard patterns just wont give the proper
scoping semantics for the local. So basically if you do a dancer2 specific
try/catch module then IMO *it* will still need the shim.

Personally I dont like try/catch wrappers around eval or do. I find they
distract from understanding the code, and lead to subtle bugs and
misunderstandings. (I loathe Try::Tiny for instance.)

But that is not a call for me to make.

Cheers,
Yves




-- 
perl -Mre=debug -e "/just|another|perl|hacker/"

veryrusty:
Thanks for the explanation @demerphq 

I'm happy with this being merged (again ;) ).  :+1:

xsawyerx:
Remerged!

Issue #1191, opened 2016-06-20
Nammed capture regex messing with the routing engine


pierre-vigier:
Hi,

in a test script, i was exposed to some strange behavior when parsing a response with a nammed capture regex. After using the regex, all the requests were defaulted to the same route. Scoping the regex solved the issue.
I wrote a small piece of code to show case the issue.

```perl
package app;
use Dancer2;
get '/1' => sub {
    return '1';
};
get '/2' => sub {
    return '2';
};
package main;
use Plack::Test;
use HTTP::Request;

my $test = Plack::Test->create( app->to_app );
my $request  = HTTP::Request->new( GET => 'http://localhost/1' );
my $response = $test->request( $request );
print $response->content, "\n";
"12345" =~ m#(?<capture>23)#;
my $c = $+{capture};
$request  = HTTP::Request->new( GET => 'http://localhost/2' );
$response = $test->request( $request );
print $response->content, "\n";
```

That piece of code is display
```
1
1
```

If i use a regex without nammed capture, it is then working correctly
```perl
"12345" =~ m#(23)#;
my $c = $1
```

If i scope the regex part, it's working as well
```perl
{
    "12345" =~ m#(?<capture>23)#;
    my $c = $+{capture};
}
```

Both of those version display 1 and then 2 as expected.

I can reproduce on Dancer2 0.166001 and Dancer2 0.200001 , the problem is present on perl 5.16.3 and 5.24.0 , i did not try other version of perl.

Pierre

demerphq:
Fixed in:

https://github.com/PerlDancer/Dancer2/pull/1258

Simple explanation, use of regex vars without properly checking that the pattern matched. Since pattern vars are dynamically scoped this means the Dancer2 internals read the state of an external pattern.

xsawyerx:
Fixed by @demerphq in PR. Merged.
Thanks! :+1:

PR #1258, opened 2016-10-04
Fixup issue #1191 - use of named captures prior to dispatch breaks di…

demerphq:
…spatch

The code to handle named captures was being executed
even when the pattern did not match. If a named capture
matches then @values must have at least one value so
we can just move the empty @values guard clause up one
block.

I do think its a bit odd that the presence of named captures
causes such a change in code path. That seems dangerous
and problematic, but I leave that to a future bug report

### DIFF
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -83,16 +83,17 @@ sub match {
 
     my @values = $request->dispatch_path =~ $self->regexp;
 
+    return unless @values;
+
     # if some named captures are found, return captures
     # no warnings is for perl < 5.10
+    # - Note no @values implies no named captures
     if (my %captures =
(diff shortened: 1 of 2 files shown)


demerphq:
Hey guys, not be a nag, but of the three pr's i pushed the other day IMO this one the most "no brainer" of the changes. Basically the existing code is plain and simply wrong to use a regex meta-variable without checking for a successful match. Lets not forget this one huh?

SysPete:
:+1:

racke:
Looks reasonable to me and comes with a proper test, so :+1:

xsawyerx:
:+1: 
(I was too busy coming back from Seattle...)

xsawyerx:
Merged, thanks! :+1:

PR #1259, opened 2016-10-04
fix #1255 - add_hook overriding is not operational

### LINKED ISSUE #1255: add_hook overriding is not operational
There is a typo in line 575 when overriding app->add_hook

$hook_code is set to $hook->code although $hook is a hash reference, not a blessed hash reference.

Therefore the code, when used, is raising an exception.

The correct syntax should be :

$hook_code = $hook->{code}.
```

# "There's a good reason for this, I swear!"
#    -- Sawyer X
# basically, if someone adds a hook to the app directly
# that needs to access a DSL that needs the current object
# (such as "plugin_setting"),
# that object needs to be available
# So:
# we override App's "add_hook" to provide a register a
# different hook callback, that closes over the plugin when
# it's available, relocalizes it when the callback runs and
# after localizing it, calls the original hook callback
{
    ## no critic;
    no strict 'refs';
    no warnings 'redefine';
    my $orig_cb = Dancer2::Core::App->can('add_hook');
    $orig_cb and *{'Dancer2::Core::App::add_hook'} = sub {
        my ( $app, $hook ) = @_;

        my $hook_code = $hook->code; #### <== HERE ####
        my $plugin    = $CUR_PLUGIN;

        $hook->{'code'} = sub {
            local $CUR_PLUGIN = $plugin;
            $hook_code->(@_);
        };

        $orig_cb->(@_);
    };
}
```

demerphq:
Better fix than suggested in the bug report, we check if $hook is blessed, if it is we call the method, otherwise we access the $hook slot.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -572,7 +572,7 @@ sub _exporter_expand_sub {
     $orig_cb and *{'Dancer2::Core::App::add_hook'} = sub {
         my ( $app, $hook ) = @_;
 
-        my $hook_code = $hook->code;
+        my $hook_code = Scalar::Util::blessed($hook) ? $hook->code : $hook->{code};
         my $plugin    = $CUR_PLUGIN;
 
         $hook->{'code'} = sub {


veryrusty:
Looks good. :+1: Thanks @demerphq!

xsawyerx:
:+1:

SysPete:
:+1:

xsawyerx:
Merged. Thanks! :+1:

PR #1267, opened 2016-10-07
Feature/send file charset

### LINKED ISSUE #1260: Allow charset as an arg to send_file
Scenario:

App has default charset set to UTF8 but for one particular text file delivered using send_file iso-8859-1 is needed instead.

veryrusty:
Specify optional `charset` to `send_file` and `send_as`:

```
send_file( $filehandle, content_type => 'text/plain', charset => 'utf-8' );
```

Resolves #1260.

### DIFF
--- a/t/dsl/send_file.t
+++ b/t/dsl/send_file.t
@@ -36,7 +36,7 @@ use File::Spec;
 
     get '/filehandle' => sub {
         open my $fh, "<:raw", __FILE__;
-        send_file( $fh, content_type => 'text/plain' );
+        send_file( $fh, content_type => 'text/plain', charset => 'utf-8' );
     };
 
     get '/check_content_type' => sub {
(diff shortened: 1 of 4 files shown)


SysPete:
Looks good to me :+1:

t/dsl/send_file.t
```diff
@@ -36,7 +36,7 @@ use File::Spec;
 
     get '/filehandle' => sub {
         open my $fh, "<:raw", __FILE__;
-        send_file( $fh, content_type => 'text/plain' );
+        send_file( $fh, content_type => 'text/plain', charset => 'utf-8' );
```
xsawyerx: There's a big difference between `utf-8` and `UTF-8`. This might be traced back to some of our Unicode problems.

veryrusty: Except.. not in this instance. There's no encoding going on here; the `charset` gets appends to the response `Content-Type` header. Happy to make that clearer in the Manual..

racke:
:+1:

cromedome:
:+1:

cromedome:
Merged in with 608b458f84c399f2c4083c1910fd324666584004. Thanks!

Issue #1253, opened 2016-09-25
304 (Not Modified) for static files or send_file(...)


vanHoesel:
A server can respond with a `Last-Modified` header. This enables (intermediate) caches to revalidate stale responses with a requests that includes a `If-Modified-Since`.

The expected response on such a revalidation request should (in general) either be a `200 (OK)` response or hopefully a `304 (Not Modified)`. The latter one, as might be clear from the description, indicaties that the stored response in the cache is still valid and that the cache can use it to respond to it's client. If the resource or webpage or file has changed since that timestamp, it must return a `200 (OK)` response, and the full payload or body.

When the Dancer2 Framework (out of the box) serves a static file from the `public` directory or when `send_file` has been called, it does send a `Last-Modified` HTTP date, based on the file-system. All good and well, however, when a client or cache would then later try to do a revalidation request, the framework does not respond with a `304` ever, even though the file has not been changed. It will always (assuming no errors occurred) a `200 (OK)` status.

I'd love to see that my favorite web-framework would do the right thing. The current solution is not disastrous, it only slows down the internet tremendously. Wouldn't it be nice that even novice and medium experienced users of Dancer2 would get the right responses, without  the need of installing another proxy server in front of Dancer2 ?

Please make it return a `304 (Not Modified)` when applicable.

Cheers, and happy dancing

veryrusty:
Thanks for raising the issue @vanHoesel.

This is something we'll need to work with upstream authors to remedy. Work is in process to do so.

Until then; depending on exact requirements one can implement the bulk of the requirements for `Not Modified` responses by using the ConditionalGET middleware in the app.psgi file:

```
use strict;
use warnings;

use Plack::Builder;
use Plack::Middleware::ConditionalGET;
use Plack::App::File;

use MyApp;

my $app = MyApp->to_app;
my $builder = Plack::Builder->new();

# static content paths
my $public_dir = MyApp->config->{public_dir}
for my $path (qw/css fonts images js/) {
    $builder->mount( "/$path" => builder {
          enable "Plack::Middleware::ConditionalGET";
          Plack::App::File->new( root => "$public_dir/$path" )->to_app;
    });
}

# mount application itself at '/'
$builder->mount('/' => $app);

# return the PSGI coderef
$builder->to_app;
```

veryrusty:
Upstream PR:  https://github.com/plack/Plack/pull/575

SysPete:
@veryrusty :hugs: :beers:

[...]

PR #1268, opened 2016-10-09
Feature/static 304

veryrusty:
The previous use of Middleware::Static was roughly

  If a file exists in the public directory that matched the request
  PATH_INFO, then use App::File to generate the response for the
  request.  Otherwise we pass-through to the Dancer2 app.

Then #1253 requested (quite reasonably) that the static handler support
Conditional GET's. However its not possible to do that with
Middleware::Static. A patch was submitted upstream and rejected; as
there *is* a simplier way of implementing the above logic using existing
middleware.

Instead, the above logic is implemented using Middleware::Conditional to
deal with the "if/then/else" part. If the file exists, use an App::File
instance wrapped in ConditionalGET to server the content, resolving #1253.
Otherwise the request is handled by out app.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -11,9 +11,11 @@ use Sub::Quote;
 use File::Spec;
 use Module::Runtime    'use_module';
 
+use Plack::App::File;
 use Plack::Middleware::FixMissingBodyInRedirect;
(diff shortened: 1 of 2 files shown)


cromedome:
:+1: Nice fix!

SysPete:
Excellent! :+1:

Issue #1243, opened 2016-09-05
body_parameters not always being decoded


SysPete:
I have not been able to reproduce this in a test. See 52e6d529a9d147aabde2e466f5dceb5950ac2e29 for a hack which fixes my app but breaks tests added in #1232

veryrusty:
I suspect the same applies to query parameters. See these two lines:
 *  https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Request.pm#L356
 *  https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/Request.pm#L375

Both `$self->SUPER::query_parameters` and `$self->SUPER::body_parameters` return Hash::MultiValue objects. Calling `_decode_ on those isn't going to do what we want..

veryrusty:
(deleted my previous incorrect comment)

@SysPete your "hack" is correct.  We need to stash the *decoded* body parameters separately from the plack request's stash of the (encoded) body parameters. Mixing the two (as we were) causes headaches..

veryrusty:
I've rebased and amended @SysPete's commit so that deserialized data is stashed in the same way, so the tests added in 1232 pass. See 96e662a.

cromedome:
Per discussion with @SysPete this can be closed. Thanks!

PR #1269, opened 2016-10-09
Stash our decoded body_parameters separately to Plack::Requests encoded stash.

veryrusty:
It is no good using $self->env->{'plack.request.body'} to stash the
*decoded* body_parameters. Middleware may have interacted with the
Plack::Request stash of the *encoded* body parameters, and certainly
would not be expecting decoded parameters

Instead stash in a local attribute that we have control over, resolving #1243.


SysPete:
I'm :+1: here but I'm not really sure my vote counts :grinning:

cromedome:
:+1:

cromedome:
Merged. Thanks!

PR #1270, opened 2016-10-10
Named routes

xsawyerx:
[This related to GH #33.]

This PR allows a user to name their routes and then use a new keyword `uri_for_route()` to generate a path to that route with a LOT of control over how it's constructed.

Giving names:

```
# Assuming this is run on localhost:5000

get 'test1' => '/:foo' => sub {1};
get 'test2' => '/*/**' => sub {1};
get 'test3' => '/:foo/:bar/*' => sub {1};
get 'test4' => '/:foo[Str]' => sub {1};
```

Using it in code:

```
# Assuming this is run on localhost:5000

# Route parameters
# $path = http://localhost:5000/bar
$path = uri_for_route( 'test1', { 'foo' => 'bar' } ); 

# Route parameters with query parameters
# $path = http://localhost:5000/bar?id=4
$path = uri_for_route( 'test1', { 'foo' => 'bar' }, { 'id' => 4 } ); 

# Splat and Megasplat args
# $path = http://localhost:5000/foo/bar/baz
$path = uri_for_route( 'test2', [ 'foo', [ 'bar', 'baz' ] ] );

# Mixed arguments
# $path = http://localhost:5000/1/2/hello
$path = uri_for_route( 'test3', { 'foo' => 1, 'bar' => 2, 'splat' => ['hello'] } ); 

# Escaping content
# $path = http://localhost:5000/bar?id=he%2Flo
$path = uri_for_route( 'test1', { 'foo' => 'bar' }, { 'id' => 'he/lo' } );

# Types route parameters
# $path = http://localhost:5000/bar
$path = uri_for_route( 'test1', { 'foo' => 'bar' } ); 
get NAME, PATH, sub {...};
```

This also works in templates:

```
<!-- some_template.tt -->

[% request.uri_for_route( 'my_route_name', { 'foo' => 'bar' }, { 'id' => 4 } ) %]
```

* Works with route parameters
* Works with typed route parameters
* Works with splat
* Works with megasplat
* Works on GET, POST, PUT, PATCH, DELETE
* Does not work with regexp routes
* Does not allow multiple routes of the same name
* Not supported on HEAD routes (which are created automatically with GET routes)


xsawyerx:
(rebased from *master*)

xsawyerx:
Any comments from the @PerlDancer team?

SysPete:
So does this mean if the user does: 
```
uri_for( '1', { id => 4 } );
```
and expects something like `http://example.com/1?id=4` they are not going to get what they expect?

xsawyerx:
Wouldn't you have to provide `uri_for( '/1' )` if you wanted to get `http://example.com/1`?

xsawyerx:
We can also try to either revise the syntax, like:

    uri_for( { '1' => { 'id' => 4 } } );

Or add a new keyword:

    named_uri_for( '1', { 'id' => 4 } );

xsawyerx:
We can also detect that if the second parameter is a hashref, the first parameter is a name, not a path.

SysPete:
Second param can already be a hashref:
```
sub uri_for {
    my ( $self, $part, $params, $dont_escape ) = @_;
    ...
    $uri->query_form($params) if $params;
```

SysPete:
I think I'd prefer a new `named_uri_for` keyword just to be sure weird stuff doesn't happen by accident.

racke:
Yeah a new keyword would be less confusing and prevents nasty surprises.

xsawyerx:
Alright, so update:

In introduced a new keyword for the `uri_for`. It's called `uri_for_route`.

However: (and following is a summary of 047d858129f9d4dfa60b91fa2da5214e2794b213)

* You cannot send it the query parameters as you could for `uri_for`.
* The only parameters currently used are for route parameters. (So, splat/megaplat isn't supported.)
* The "don't escape" option is not supported.
* I haven't exposed this to the template yet, because I'm not sure how. Suggestions welcomed.

cxw42:
This looks very handy!  Any news?  Has anyone been running on this branch and noticed any gotchas?

lib/Dancer2/Core/App.pm
```diff
@@ -1619,6 +1620,33 @@ sub response_not_found {
     return $response;
 }
 
+sub uri_for_route {
+    my ( $self, $route_name, $named_params, $unnamed_params, $dont_escape ) = @_;
```
cxw42: It looks like `$unnamed_params` and `$dont_escape` are not yet implemented.  To avoid footguns, should we croak if either is defined?  Or, better yet, just remove them from the arg list for now?  

Separately, is `$unnamed_params` actually intended to be the future query-parameters argument, similar to [`uri_for`](https://metacpan.org/pod/distribution/Dancer2/lib/Dancer2/Manual.pod#uri_for)?

lib/Dancer2/Core/DSL.pm
```diff
@@ -232,11 +233,22 @@ sub _normalize_route {
...
+        # @_ = ( REGEXP, OPTIONS, CODE )
+        # @_ = ( NAME, REGEXP, CODE )
+        if (ref $_[1] eq 'HASH') {
+            @args{qw<regexp options code>} = @_;
+        } else {
+            @args{qw<name regexp code>} = @_;
+        }
+    }
 
     return map $app->add_route( %args, method => $_ ), @{$methods};
```
cxw42: Would it be worth having add_route store a hashref per method from name to route for lookup speed and for detection of attempts to accidentally add the same route twice?

lib/Dancer2/Core/App.pm
```diff
@@ -1619,6 +1620,33 @@ sub response_not_found {
...
+
+    my $route =
+      List::Util::first { $_->name eq $route_name } @{$self->routes->{'get'}};
```
cxw42: Taking inspiration from `any`, would it be worth accepting an arrayref of `[$method, $name]` for `$route_name`?  Then we could support POST with no extra effort.

lib/Dancer2/Core/App.pm
```diff
@@ -1619,6 +1620,33 @@ sub response_not_found {
...
+    $string =~ /[:@]/
+        or return $string;
+
+    # XXX This regex here are taken from
+    #     Dancer2::Core::Route::_build_regexp_from_string.
+    #     Would anyone like to refactor? -- SX.
+    my @params = $string =~ m{:([^/\.\?]+)}g;
+    foreach my $param (@params) {
+        my $value = $named_params->{$param}
+            or Carp::croak("$param is not a part of route $route_name");
```
cxw42: This error message confuses me --- we know `$param` is in the route, since it came from `$route->spec_route`.  I think this should be

    Route $route_name uses parameter ${param}, which was not provided

or some such.  Am I missing something?

lib/Dancer2/Core/Route.pm
```diff
@@ -9,6 +9,14 @@ use Scalar::Util 'blessed';
 
 our ( $REQUEST, $RESPONSE, $RESPONDER, $WRITER, $ERROR_HANDLER );
 
+my $count = 0;
+
+has name => (
+    is      => 'ro',
+    isa     => Str,
+    default => sub { $count++ },
```
cxw42: I suggest `++$count` so that the autogenerated names are always truthy (until ints wrap around :) ).  (I'm not sure when I would test `if($route->name)`, but somebody might, and truthy default names would prevent the first route declared from behaving differently from the others.)

lib/Dancer2/Core/Route.pm
```diff
@@ -9,6 +9,14 @@ use Scalar::Util 'blessed';
 
 our ( $REQUEST, $RESPONSE, $RESPONDER, $WRITER, $ERROR_HANDLER );
 
+my $count = 0;
+
+has name => (
+    is      => 'ro',
+    isa     => Str,
```
cxw42: I suggest requiring route names to be truthy for consistency with my next comment.

lib/Dancer2/Core/DSL.pm
cromedome: By design, to make sure a route isn't added twice (and to locate it) we'd agree we need a hashref here. Thanks :+1:

lib/Dancer2/Core/Route.pm
cromedome: Ok. :+1:

lib/Dancer2/Core/App.pm
cromedome: We'll change this. Thanks for pointing it out.

lib/Dancer2/Core/Route.pm
cromedome: This can be verified when you create a new one. We will test when a named route is created.

lib/Dancer2/Core/App.pm
xsawyerx: Took me time digging from the commit message and my memory to remember the issue.

I separated named parameters from unnamed parameters because if `splat` or megasplat are used, they need expanding unnamed parameters.

I think it's possible to support unnamed parameters (I just didn't because I wanted to throw a quick PR out, little did I know...) so I'll try to add that.

The escaping isn't supported but *should* be supported.

lib/Dancer2/Core/DSL.pm
xsawyerx: Dancer allows, by design, adding any amount of routes that conflict. There are two types of conflicts: Unintentional and intentional.

Unintentional would be `get '/foo' => sub {...}` and `get '/:word' => sub {...}`. A request to `/foo` can be served by both.

Intentional would be `get '/foo' => sub {...}` and `get '/foo' => sub {...}`. While the proposal for eliminating this is appealing, Dancer does allow it. (Changing this would introduce breakage of the Dancer behavior as part of this and I'm not big on doing it.)

For both cases, Dancer provides tools for picking what will happen. The first matching route will be called, then `pass` allows ending it and trying to match the next one.

```
get '/' => sub {
    # Main router
    session('user')
        or pass;
    }

    ... # handle logged in users
};

# ... other routes ...

get '/' => sub {
    template 'default';
};
```

This seems counter-intuitive, but it is the way Dancer2 is designed.

Having said this, there might be room for a strict mode that tries to find this and surface this, or even prevents it from happening, but I suggest we separate this from the current discussion so we can support what Dancer currently provides.

lib/Dancer2/Core/App.pm
xsawyerx: I think this is reasonable. Let me think of how I want to support other methods before I commit to the semantics of `any`.

lib/Dancer2/Core/App.pm
xsawyerx: 👍

lib/Dancer2/Core/Route.pm
xsawyerx: 👍

lib/Dancer2/Core/Route.pm
xsawyerx: `Types::Standard` doesn't have `True()` as a type. There is `Defined()` which I believe all strings are anyway, as well as `0` as a value.

Is there are particular type value you're suggesting or just replacing it with `sub { defined && length && !!$_ }` or something?

lib/Dancer2/Core/App.pm
xsawyerx: If we add a check for the route by name, it won't matter if the route is a GET, PUT, POST, or anything else.

We also don't produce anything other than a URI, so whether it will be used in a GET or POST method is up to the user.

xsawyerx:
Addendum: Routes defined with regexp directly cannot have their arguments expanded. This would require `uri_for_route()` to crash if it was defined as a regexp.

```
get 'entry' => qr{/view/[0-9]+} => sub {...};

uri_for_route( 'entry', ... ); # Boom!
```

lib/Dancer2/Core/Route.pm
cxw42: @xsawyerx That `sub` indeed covers what I was thinking of.  I think `sub { !!$_ }` would be sufficient:
- `!!$_` true implies `defined`.  
- As for `length`, I'd say a user who goes to the trouble of making an empty string truthy should not expect any help with issues :laughing:

xsawyerx:
Okay, rebased and rewrote everything.

Here is what we have now:

```
# Assuming this is run on localhost:5000

get 'test1' => '/:foo' => sub {1};
get 'test2' => '/*/**' => sub {1};
get 'test3' => '/:foo/:bar/*' => sub {1};
get 'test4' => '/:foo[Str]' => sub {1};

# ...

# Route parameters
# $path = http://localhost:5000/bar
$path = uri_for_route( 'test1', { 'foo' => 'bar' } ); 

# Route parameters with query parameters
# $path = http://localhost:5000/bar?id=4
$path = uri_for_route( 'test1', { 'foo' => 'bar' }, { 'id' => 4 } ); 

# Splat and Megasplat args
# $path = http://localhost:5000/foo/bar/baz
$path = uri_for_route( 'test2', [ 'foo', [ 'bar', 'baz' ] ] );

# Mixed arguments
# $path = http://localhost:5000/1/2/hello
$path = uri_for_route( 'test3', { 'foo' => 1, 'bar' => 2, 'splat' => ['hello'] } ); 

# Escaping content
# $path = http://localhost:5000/bar?id=he%2Flo
$path = uri_for_route( 'test1', { 'foo' => 'bar' }, { 'id' => 'he/lo' }, 1 );

# Types route parameters
# $path = http://localhost:5000/bar
$path = uri_for_route( 'test1', { 'foo' => 'bar' } ); 
```

Lots of tests, updated documentation.

I'm also testing for various failures:
* You tried to create multiple routes with the same name
* You failed to include parameter values for all the route parameters
* You tried calling `uri_for_route()` to a nonexistent route
* You provided an incorrect splat/megasplat arguments compared to how many were necessary

Last things to check before fully signing off:
* CI tests failures
* Support in template (because that's when it would be really useful)

lib/Dancer2/Core/Route.pm
xsawyerx: I eventually replaced it with `has_name()` predicate because I needed to know when a route name can be used or not.

Providing a default name might conflict with a name a user provides. Imagine we give them numbers and a user does the same - there's a conflict. I'm not sure we should be automatically naming them. This can also lead to a possible accidentally call to `uri_for_route()` in which the name is typo'ed but since we named it, it hits a different route.

xsawyerx:
Fixed all the tests.

xsawyerx:
Using it in templates is now supported as well:

```
<!-- some_template.tt -->

[% request.uri_for_route( 'my_route_name', { 'foo' => 'bar' }, { 'id' => 4 }, 1 ) %]
```

(This is an example with all possible arguments.)

lib/Dancer2/Core/DSL.pm
```diff
@@ -241,11 +242,26 @@ sub _normalize_route {
...
-    @args{qw/regexp options code/} = @_ == 3 ? @_ : ( $_[0], {}, $_[1] );
+    # Options are optional, try to deduce their presence from arg length.
+    if ( @_ == 4 ) {
+        # @_ = ( NAME, REGEXP, OPTIONS, CODE )
+        # get 'foo', '/foo', { 'user_agent' => '...' }, sub {...}
+        @args{qw<name regexp options code>} = @_;
+    } elsif ( @_ == 2 ) {
+        # @_ = ( REGEXP, CODE )
+        # get '/foo', sub {...}
+        @args{qw<regexp options code>} = ( $_[0], {}, $_[1] );
```
cxw42, line 253: When I first read this, I tripped over `REGEXP, CODE` vs. `regexp options code`.  Suggestion to make this self-documenting and consistent with with the other cases:
```perl
@args{qw<regexp code>} = ( $_[0], $_[1] );
$args{options} = {};
```

lib/Dancer2/Core/DSL.pm
```diff
@@ -241,11 +242,26 @@ sub _normalize_route {
...
+    } elsif ( @_ == 3 ) {
+        # @_ = ( REGEXP, OPTIONS, CODE )
+        # get '/foo', { 'user_agent' => '...', sub {...}
+        # @_ = ( NAME, REGEXP, CODE )
+        # get 'foo', '/foo',sub {...}
+        if (ref $_[1] eq 'HASH') {
+            @args{qw<regexp options code>} = @_;
+        } else {
+            @args{qw<name regexp code>} = @_;
```
cxw42, line 262: It looks like `options` isn't set on this path --- should it be?  I not, then it can also be removed from the `@_ == 2` case.

lib/Dancer2/Core/DSL.pm
```diff
@@ -241,11 +242,26 @@ sub _normalize_route {
...
+        # @_ = ( NAME, REGEXP, OPTIONS, CODE )
+        # get 'foo', '/foo', { 'user_agent' => '...' }, sub {...}
+        @args{qw<name regexp options code>} = @_;
+    } elsif ( @_ == 2 ) {
+        # @_ = ( REGEXP, CODE )
+        # get '/foo', sub {...}
+        @args{qw<regexp options code>} = ( $_[0], {}, $_[1] );
+    } elsif ( @_ == 3 ) {
+        # @_ = ( REGEXP, OPTIONS, CODE )
+        # get '/foo', { 'user_agent' => '...', sub {...}
```
cxw42, line 256: typo?
```suggestion
        # get '/foo', { 'user_agent' => '...' }, sub {...}
```

lib/Dancer2/Core/Route.pm
```diff
@@ -12,6 +12,14 @@ use Type::Registry;
 
 our ( $REQUEST, $RESPONSE, $RESPONDER, $WRITER, $ERROR_HANDLER );
 
+my $count = 0;
```
cxw42, line 15: I see this is added in b4707f05, but it appears unused as of 2ff2d530.  Should it be removed from b4707f05?

lib/Dancer2/Manual.pod
```diff
@@ -213,6 +213,13 @@ coderef to execute, which returns the response.
 The above route specifies that, for GET requests to C</hello/...>, the code
 block provided should be executed.
 
+
+You can also provide routes with a name:
+
+    get 'hi_to' => '/hello/:name' => sub {...};
+
+See C<uri_for_route> on how this can be used.
```
cxw42, line 221: Suggestion:
```suggestion
See L<Dancer2::Manual::Keywords/uri_for_route> on how this can be used.
```
(and likewise in other pod outside D2::M::K)

lib/Dancer2/Core/App.pm
```diff
@@ -4,13 +4,14 @@ package Dancer2::Core::App;
 use Moo;
 use Carp               qw<croak carp>;
 use Scalar::Util       'blessed';
+use List::Util         ();
```
cxw42, line 7: This duplicates line 13

lib/Dancer2/Manual/Keywords.pod
```diff
@@ -1146,6 +1180,108 @@ URL encoding via a third parameter:
     uri_for('/path', { foo => 'qux%3Dquo' }, 1);
     # would return http://localhost:5000/path?foo=qux%3Dquo
 
+=head2 uri_for_route
+
+An enhanced version of C<uri_for> that utilizes their names.
```
cxw42, line 1185:
```suggestion
An enhanced version of C<uri_for> that utilizes routes' names.
```

cxw42 (approved):
Thanks very much!

Some points re. b4707f05's metadata:
* the commit summary ends with a trailing colon --- is something missing from the summary line?
* The author date is in 2016!
* the commit message still says "This should lend in the future to allow `uri_for` (or a different DSL keyword)".  Now that 2ff2d530 implements `uri_for_route`, I think this piece of the message should be removed or modified.
* the commit messages says "globally incrementing number".  However, it looks like 2ff2d530 changes that to leave routes anonymous unless a name is given.  Therefore, I suggest removing the `$count` logic from b4707f05 and removing the relevant text from the commit message.

lib/Dancer2/Core/App.pm
```diff
@@ -1694,6 +1710,63 @@ sub response_not_found {
...
+            if ( is_arrayref($splat_params->[$i]) ){
+                my $megasplat = join '/', @{ $splat_params->[$i] };
+                $string =~ s{\Q(?#megasplat)\E}{$megasplat};
+            } else {
+                $string =~ s{\Q(?#splat)\E}{$splat_params->[$i]};
+            }
+        }
+    }
+
+    return $self->request->uri_for( $string, $query_params );
```
cxw42, line 1767: As of 2ff2d530, it looks like `$dont_escape` is still unused.  However, an escape-control parameter is documented.  Should that parameter be passed to this `uri_for()` call?

lib/Dancer2/Manual/Keywords.pod
```diff
@@ -1146,6 +1180,108 @@ URL encoding via a third parameter:
...
+(Technically, only C<GET> requests should include query parameters, but
+C<uri_for_route> does not enforce this.)
+
+=item * Escaping
+
+The final parameter determines whether the URI will be URI-escaped:
+
+    get 'show_entry' => '/view/:str_id' => sub {1};
+    # ...
+    $path = uri_for_route( 'show_entry' => { 'str_id' => '!£%^@' }, {}, 1 );
```
cxw42, line 1276: The `1` is the `$dont_escape` parameter, but this is showing the result as URI-escaped.  I think the documentation of this parameter needs to be flipped around --- default == escape, `1` == don't escape.  But I could be missing something!

lib/Dancer2/Core/App.pm
cxw42: Resolved (but GH doesn't give me a "Resolve conversation" button)

lib/Dancer2/Core/DSL.pm
cxw42: Resolved

lib/Dancer2/Core/App.pm
cxw42: Resolved

lib/Dancer2/Core/App.pm
cxw42: Resolved

lib/Dancer2/Core/Route.pm
cxw42: Resolved --- no longer relevant since routes are no longer given automatic names.

lib/Dancer2/Core/Route.pm
cxw42: Resolved

lib/Dancer2/Core/DSL.pm
xsawyerx: I prefer not to. Notice that `@args{...}` is always done with an order and I can see the value for each one. Sorry, but this is more readable for me this way.

lib/Dancer2/Core/DSL.pm
xsawyerx: I'm sorry. I'm not sure I understand this comment.

lib/Dancer2/Core/DSL.pm
xsawyerx: Fixed

lib/Dancer2/Core/Route.pm
xsawyerx: Fixed!

lib/Dancer2/Manual.pod
xsawyerx: Good call! Fixed.

lib/Dancer2/Core/App.pm
xsawyerx: Nice catch! Fixed.

lib/Dancer2/Manual/Keywords.pod
xsawyerx: Fixed.

lib/Dancer2/Core/App.pm
xsawyerx: Just noticed this myself. :) Fixed!

lib/Dancer2/Manual/Keywords.pod
xsawyerx: Yeah, this confused me a bit too. I decided to keep this with `uri_for()` so it's alas, a disabling flag that - when set to true - will disable.

I fixed the docs though!

xsawyerx:
I appreciate your thorough review!

> Thanks very much!
> 
> Some points re. [b4707f0](https://github.com/PerlDancer/Dancer2/commit/b4707f05f7a023c68037be142c0437415aa3cb94)'s metadata:
> 
>     * the commit summary ends with a trailing colon --- is something missing from the summary line?

This is a common practice of mine. When a commit summary line has a colon, it means I have commit text. When I don't add a commit content, I don't end it with a colon.

So, on purpose.

> 
>     * The author date is in 2016!

Yup. I amended, which kept this data.

>     * the commit message still says "This should lend in the future to allow `uri_for` (or a different DSL keyword)".  Now that [2ff2d53](https://github.com/PerlDancer/Dancer2/commit/2ff2d530efb677c1e237532e42ecb22a3c10d0a0) implements `uri_for_route`, I think this piece of the message should be removed or modified.

This is, to me, a piece of history. I intended to put it in `uri_for()` but couldn't. I think it's fine.

> 
>     * the commit messages says "globally incrementing number".  However, it looks like [2ff2d53](https://github.com/PerlDancer/Dancer2/commit/2ff2d530efb677c1e237532e42ecb22a3c10d0a0) changes that to leave routes anonymous unless a name is given.  Therefore, I suggest removing the `$count` logic from [b4707f0](https://github.com/PerlDancer/Dancer2/commit/b4707f05f7a023c68037be142c0437415aa3cb94) and removing the relevant text from the commit message.

Same.

lib/Dancer2/Manual/Keywords.pod
```diff
@@ -1146,6 +1180,124 @@ URL encoding via a third parameter:
...
+    # ...
+    $path = uri_for_route(
+        'test',
+        [ 'hello', 'world', [ 'myhello', 'myworld' ],
+    );
+    # $path = http://localhost:5000/hello/world/myhello/myworld
+
+=item * Mixed route parameters
+
+If you have a route that includes both, the plat and megasplat
```
cromedome, line 1232: `plat` should be `splat`.

lib/Dancer2/Core/DSL.pm
cxw42: Sorry for not being sufficiently clear!  For context, I'm using substantially the same code-review standards I do at $work :) .

If `@_ == 3 && !is_hashref($_[1])`, `$args{options}` winds up unset.  This is different from all the other cases in this function.  I don't know why there is a difference, but the fact that there's a difference stood out to me.  

Does `$args{options}` need to be set when `$app->add_route` is called?  I don't think it does, because it gets plumbed through to `Route`, where `options` is not marked as required ([ref.](https://github.com/PerlDancer/Dancer2/pull/1270/files#diff-e2c044e62f9fd3169102d5e7583838633cef4548832abad3c56a714ae28b8440L40-L45)).  However, I could be wrong.

- If we do need `$args{options}` in all cases, this line could be something like
  ```suggestion
              @args{qw<name regexp code options>} = (@_, +{});
  ```
- If we do not need `$args{options}` in all cases, then line 253 can be simplified to
  ```diff
  -        @args{qw<regexp options code>} = ( $_[0], {}, $_[1] );
  +        @args{qw<regexp code>} = ( $_[0], $_[1] );
  ```

lib/Dancer2/Core/DSL.pm
cxw42: ACK --- understood!

lib/Dancer2/Core/DSL.pm
xsawyerx: Ah! Now I understand! That's a great catch. I fixed it to remove the added setting of an empty `options` key. That attribute has a predicate to determine if it's available. And even if it's available, it will only be used to iterate and set values.

A change that we can also make is to always set it to empty hashref if it's not provided and then remove the predicate and the check for the predicate. Since it will only trigger a loop on existing keys and there would be no keys, it wouldn't be an issue if it's set by default to empty hashref.

Thanks for the catch and for explaining it till I got it. :)

lib/Dancer2/Manual/Keywords.pod
xsawyerx: Fixed. Thanks!

cromedome:
Make sure to document usage from a template:
```
<!-- some_template.tt -->

[% request.uri_for_route( 'my_route_name', { 'foo' => 'bar' }, { 'id' => 4 } ) %]
```

xsawyerx:
> `<!-- some_template.tt --> [% request.uri_for_route( 'my_route_name', { 'foo' => 'bar' }, { 'id' => 4 } ) %]`

Done.

Also updated the article. I think it's ready to merge.

cromedome:
👏 This looks great. Thanks for walking me through this. An emphatic 👍 from me.

xsawyerx:
@racke What do you think?

racke (approved):
Looks really good :+1:

cromedome:
Excellent! I'll work on getting a release put together over the next couple of days. Thanks everyone!

cromedome:
Ready for release. Thanks everyone!

PR #1271, opened 2016-10-11
fix wrong regex check against  $_

mickeyn:
pretty sure the original line shouldn't have checked against $_ and should have been:
`... unless $plugin =~ /^Dancer2::Plugin::/;`

while at it, we could have just add the check into a negative look-ahead on the substitution regex.


veryrusty:
:+1:

SysPete:
:+1:

xsawyerx:
:+1:

xsawyerx:
Merge away.

PR #1272, opened 2016-10-11
eval named capture regex so tests pass on perl 5.8

veryrusty:
Named captures are only supported for perl >= 5.10; with perl 5.8.9
throwing an error about the syntax of the named capture group
(only version tested). eval the regex as a string of code to "hide"
the named capture on perl < 5.10.


SysPete:
:+1:

xsawyerx:
:+1:

xsawyerx:
Merged, thanks!

Issue #1273, opened 2016-10-12
Should lack of Test::Perl::Critic prevent install?


nanis:
I am asking this question, because I just ran into this on my test rig. `cpanm Dancer2` failed because `Test::Perl::Critic` depends on `MCE`  and it looks like some MCE tests are crashing with the particular build of `perl` (never had problems before, and I haven't diagnosed it yet).

But, this made me think about why there is a `test_requires 'Test::Perl::Critic';` in `cpanfile` and
```
"TEST_REQUIRES" => {
    "Test::Perl::Critic" => 0
  },
```
in `Makefile.PL`.

It seems to me that critic tests are for authors only, and the failure to run or pass those tests should not prevent an **automated** test/install of `Dancer2`? Of course, if I just unpack the distribution and do `nmake test`, I get the warning:

```
# *** WARNING WARNING WARNING WARNING WARNING WARNING WARNING WARNING ***
#
# The following REQUIRED prerequisites were not satisfied:
#
# Test::Perl::Critic is not installed (any version required)
```

but all tests pass.

xsawyerx:
Oh dear! No. This shouldn't happen!

xsawyerx:
Fixed by @episodeiv and merged!

Thanks!

nanis:
Thank you.

PR #1279, opened 2016-10-16
As per issue #1273, Test::Perl::Critic should only be an author test

episodeiv:
This should be it, shouldn't it?


xsawyerx:
Merged, thanks! :+1:

Issue #1214, opened 2016-07-16
Migration documentation updates


racke:
- [ ] Switch from param/params to new keywords: route_parameters, query_parameters, and body_parameters
- [ ] `$config->{'public_dir'}`instead of `setting('public')`

xsawyerx:
Just pushed a commit that attends to these.

racke:
Thanks @xsawyerx !

PR #1280, opened 2016-10-17
Update documentation to use specific parameter keywords

mcmillhj:
Summary of changes:
1. Change usage of `param`/`params` to use the more specific keywords `route_paramters`, `query_parameters`, and `body_parameters` where appropriate

patch for #1214


xsawyerx:
Wow! Excellent work.

Is there still some documentation of `param` that a person could find if she opens an application with the DSL in it without knowing what it is? Should we add there that it's an older (and unrecommended) version of the new DSL keywords?

racke:
Excellent :+1:

lib/Dancer2/Manual.pod
```diff
@@ -3090,7 +3075,7 @@ Or to request a specific layout, of course:
     };
 
 Some tokens are automatically added to your template (C<perl_version>,
-C<dancer_version>, C<settings>, C<request>, C<params>, C<vars> and, if you
+C<dancer_version>, C<settings>, C<request>, C<query_parameters>, C<body_parameters>, C<route_parameters>, C<vars> and, if you
 have sessions enabled, C<session>). Check L<Dancer2::Core::Role::Template>
```
SysPete: I haven't checked the code but I can see via Dancer2::Debugger that I'm getting 'params' in my tokens but none of query_parameters, body_parameters or route_parameters.

mcmillhj: Hmm, I thought I checked this. I will verify and then send another commit as needed.

mcmillhj:
@xsawyerx I left the top level documentation for param/params in `lib/Dancer2/Manual.pod`, but as you said all the examples were removed. 

What if both versions of the examples existed, but one was labelled as deprecated? That way user's that have older Dancer apps would still have access to examples.

xsawyerx:
@mcmillhj This might actually be a better idea... We could do that too. You don't think it would be too much?

We could also have a note explaining those and saying it's unrecommended, instead of at every example. What do you think?

mcmillhj:
@xsawyerx Having both may be a bit much to take in, especially for a new user. I can add the section explaining `param` and `params` and a note about not using them for new apps and we can see how it looks?

xsawyerx:
I think that sounds like a good idea! (Sorry for the late response.)

mcmillhj:
@xsawyerx No worries :) I pushed a commit that adds top-level examples for `params`, please let me know if you think any information is missing.

lib/Dancer2/Manual.pod
mcmillhj: @SysPete You were correct, within a template `route_parameters`, `body_parameters`, and `query_parameters` are not exposed. You still must reference them from `params('route')`, etc... I pushed a commit that reverts this change. 

It seems like `route_parameters`, `body_parameters`, and `query_parameters` are now the recommended way to access this data, should I file an issue to add expose them within the template engine similar to `params`?

lib/Dancer2/Manual.pod
```diff
@@ -178,16 +178,12 @@ requests.
 =head2 Route Handlers
 
 The route action is the code reference declared. It can access parameters
-through the C<params> keyword, which returns a hashref. This hashref is a
+through the specific C<route_parameters>, C<query_parameters>, and C<body_parameters> keywords, which return an L<Hash::MultiValue> object. This hashref is a
```
xsawyerx: `an L<Hash...>` -> `a L<hash...>`. I missed that before. Sorry.

lib/Dancer2/Manual.pod
```diff
@@ -1219,9 +1204,8 @@ L<Template> module.
 In order to render a view, just call the
 L<template|Dancer2::Manual/template> keyword at the end of the action by
 giving the view name and the HASHREF of tokens to interpolate in the view
-(note that for convenience, the request, session, params, and vars are
-automatically accessible in the view, named C<request>, C<session>,
-C<params> and C<vars>) - for example:
+(note that for convenience, the request, session, route_parameters, body_parameters, query_parameters, and vars are
+automatically accessible in the view, named C<request>, C<session>, C<params>, and C<vars>) - for example:
```
xsawyerx: Is this the same text just stretched beyond 80 cols?

Also, `params` appears there too. Another miss.

lib/Dancer2/Manual.pod
```diff
@@ -2551,6 +2535,19 @@ list context) or a hash reference (in scalar context) to all defined
 parameters. Check C<param> below to access quickly to a single parameter
```
xsawyerx: Maybe remove this last sentence?

lib/Dancer2/Manual.pod
xsawyerx: We can expose them as objects, so you could still call methods on them. We already do it with the `request`, `response`, etc.

If you could open an issue for that, it would be great. We could also do it and merge it before we merge this document, if you want to amend it, or we could merge a different PR that just adds that extra bit of documentation.

lib/Dancer2/Manual.pod
mcmillhj: No worries, will push a commit to fix. I originally had them as `L<hash>` but for some reason got in my head the `L<Hash::MultiValue>` was correct.

lib/Dancer2/Manual.pod
mcmillhj: I thought we wanted to leave that here for now? Since the other keywords aren't exposed yet in the template engine.

lib/Dancer2/Manual.pod
mcmillhj: This is part of the top level documentation for `param`/`params`, we wanted to keep this I thought for user's who have older Dancer2 apps.

lib/Dancer2/Manual.pod
xsawyerx: Oh I see what you mean. You're right! My bad. :)

lib/Dancer2/Manual.pod
xsawyerx: My bad again! :)

xsawyerx:
Good to merge, I believe. :)
:+1: Excellent work, @mcmillhj.

SysPete:
Don't we need #1284 to be addressed before merging this?

xsawyerx:
From what I understand from the last commit @mcmillhj introduced, we don't.

SysPete:
Bah. Missed that one. Then :+1: from me too.

racke:
Yeah, great work :+1:

mcmillhj:
Oops, one final thing. I need to update the return value listing for `route_parameters`,etc... from `L<Hash::MultiValue>` to `L<Hash>` I forgot to do that in my latest commit. After that I will rebase so everything is nice and tidy.

xsawyerx:
You don't need to change it to "Hash". I just wanted to change the typo from "an" to "a".
`a L<Hash::MultiValue> ...`, not `an L<Hash::MultiValue> ...`

mcmillhj:
Oh my mistake, I thought those keywords returned Hashes.

cromedome:
:+1:

cromedome:
Merged, thank you!

PR #1281, opened 2016-10-17
use Ref::Util in Core

mickeyn:
any reason not to use Ref::Util in all ref checks in Core?

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -31,6 +31,7 @@ requires 'Plack', '1.0035';
 requires 'Plack::Middleware::FixMissingBodyInRedirect';
 requires 'Plack::Middleware::RemoveRedundantBody';
 requires 'POSIX';
+requires 'Ref::Util';
 requires 'Return::MultiLevel';
 requires 'Role::Tiny', '2.000000';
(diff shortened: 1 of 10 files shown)


cromedome:
None that I can see. :+1: I like this - looks a lot cleaner to me.

xsawyerx:
`Ref::Util` is only XS at the moment. I need to rerelease the PP implementation of it and have `Ref::Util::XS` be the XS version. Then `Ref::Util` will try to load the XS version automatically.

This PR would stay the same, though. :)

mickeyn:
OK, so on-hold until that is done.

SysPete:
@xsawyerx any reason not to merge this one now? I guess we'd need an appropriate prereq for Ref::Util::PP.

xsawyerx:
Nope. We can merge it. The requirement could say `Ref::Util` since that will try to also install `Ref::Util::XS` if you support XS.

SysPete:
:+1: also removed 'on hold' from title.

mickeyn:
fixed conflicts

cromedome:
Thank you!

Issue #1115, opened 2016-01-25
Profiles for command line scaffolding


xsawyerx:
The `dancer2` command line utility could make use of different profiles (one using a JS client-side framework, one built on Bootstrap and FontAwesome, etc.).

cromedome:
I have an idea I would be willing to pursue:

Give `dancer2 gen` a `-t` / `--template` option to specify a template to be used for generation of the application. That template can be a URL (to a Github or Bitbucket repo, for example), a file URL (for templates you may have created locally), or the name of a template to use.

Template names could be specified in a `.dancer2templaterc` in your home dir. A simple key/value list, like ```bootstrap = https://your.url.com/repo/name
reactjs = https://your.other.url.com/repo/other```

Each template needs a .dancer2template file that says what template engine is needed, plus whatever other information we decide we need to put there.

Like `dancer2 gen` does now, a new app is built from the template. `dancer2 gen` will go to the specified URL, grab the files, and use them as the template for the new app being created.

This gives us the ability to have pre-defined templates that use different frameworks and template engines. It also allows people to build and share templates with other Dancer developers.

Thoughts? Feedback welcome. I came up with this on my second glass of bourbon tonight, so it might be complete bunk ;)

jasonblewis:
you can specify a path to an alternate skeleton with the -s option, but it like other ideas

jasonblewis:
While on this topic, it would also be nice  to have a --tag-style option to change the tags that the template is generated with.

Admittedly while this would only be useful for some types of templating engines, it would still be useful

cromedome:
@jasonblewis I'll go so far as to implement the request in #1112 - is that enough for you?

cromedome:
I think this is sufficiently handled in #1735 - see also the comment in #1111.

jasonblewis:
> [@jasonblewis](https://github.com/jasonblewis) I'll go so far as to implement the request in [#1112](https://github.com/PerlDancer/Dancer2/issues/1112) - is that enough for you?

@cromedome Sadly I don't use dancer currently so I don't feel qualified to answer that.

Thanks for looking into it, I hope i can find a need for d2 again sometime soon

PR #1283, opened 2016-10-18
add example of setting a title

jasonblewis:
I think it would be nice to show how to set the title of the html page in the generated app

### DIFF
--- a/share/skel/lib/AppFile.pm
+++ b/share/skel/lib/AppFile.pm
@@ -4,7 +4,8 @@ use Dancer2;
 our $VERSION = '0.1';
 
 get '/' => sub {
-    template 'index';
+    template 'index',
+      { title => '[d2% appname %2d]'};
 };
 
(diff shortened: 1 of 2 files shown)


racke:
Good idea :-) :+1:

xsawyerx:
While I think it's a nice example, it also turns the global example into a more specific use.

I think we would have a greater value of making the scaffolding script allow different profiles (there's a ticket for it) and then have one of them contain a title, amongst other stuff.

jasonblewis:
maybe that example isn't great but i think it would be valuable to have in the default template, an example of setting a variable in the app, and using it in the template

xsawyerx:
Agreed.

jasonblewis:
@xsawyerx is refering to https://github.com/PerlDancer/Dancer2/issues/1115

xsawyerx:
Merged!

PR #1287, opened 2016-10-21
Call route_exception hook before logging error

abeverley:
One might want to do "other things" when an exception has been raised, such as raising other exceptions. In this case, you probably don't want to also call the error logger. Therefore, this patch calls the logger after the hook rather than before. It saves dirty hacks like this:

https://metacpan.org/source/MARKOV/Log-Report-1.18/lib/Dancer2/Logger/LogReport.pm#L64


cromedome:
Ooh, nice. 👍  Thanks!

SysPete:
Makes sense to me :+1:

cromedome:
Merged, thank you!

Issue #373, opened 2013-08-09
Possible incorrect request parsing


xsawyerx:
Given this situation:

```
use Plack::Builder;
use MyWiki;

builder {
    mount '/wiki' => MyWiki->dance;
};
```

The specs for `Plack::App::URLMap` (which is what's used in the `mount` command from `Plack::Builder`) indicate that the app (`MyWiki`) will see the requests as `/` instead of `/wiki`, because it will munge the _PATH_INFO_.

The problem is that it doesn't see it that way. It still sees the mounting prefix. This means we're handling it incorrectly. You currently need to add `prefix => '/wiki'` to get it to work cleanly.

We should look into it.

PR #1288, opened 2016-10-24
Stop removing script_name from request's dispatch_path.

jbarrett:
This appears to be unnecessary and in the case where the mount point matches the start of the requested path, the request ends up with the mount point part removed from the path.

This gist should demonstrate the issue:

https://gist.github.com/jbarrett/458baf48a11802ad51f60888e7c9305d

Thanks!

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -278,15 +278,6 @@ sub dispatch_path {
 
     my $path = $self->path;
 
-    # Want $self->base->path, without needing the URI object,
-    # and trim any trailing '/'.
-    my $base = '';
-    $base .= $self->script_name if defined $self->script_name;
-    $base =~ s|/+$||;
-
-    # Remove base from front of path.
-    $path =~ s|^(\Q$base\E)?||;
-    $path =~ s|^/+|/|;
     # PSGI spec notes that '' should be considered '/'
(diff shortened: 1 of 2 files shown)


jbarrett:
I updated the gist above to try make what is happening more clear. I would expect a request to '/s/sub' to succeed in this setup.

veryrusty:
@jbarrett Thanks for reporting this. The gist is fine and makes it clear dispatch is not as it should be.

If you are interested, there is some history in #373 as to why dispatch_path was introduced. However your gist makes it clear that at some time the path handling used for dispatch has changed. I'm planning on bisecting to find where.

Note that the changes to `dispatch_path` in this Pr make that method the same as `request->path`. Hopefully after bisecting we can see if we really need that.

veryrusty:
git bisect between v0.153000  (good) and current HEAD (bad) using the test case in the gist points to the dispatch behaviour changing at commit:   c6af1c0.

At a quick look: that commit fixed the issues with `request->path` that `request->dispatch_path` was working around.

veryrusty:
Alternate PR in #1322 that deprecates `dispatch_path`.

jbarrett:
@veryrusty 👍 Thanks! I had also considered the deprecation approach, but my tuit collection is poor (but hey, whose isn't?).

PR #1290, opened 2016-11-02
Fix in _read_to_end

MarketaWachtlova:
The bug in the original code: the buffer appears empty, if it reads the string '0', which means false in Perl.
Therefore we should test buffer length instead.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -520,7 +520,7 @@ sub _read_to_end {
 
     my $body = '';
     if ( $content_length && $content_length > 0 ) {
-        while ( my $buffer = $self->_read() ) {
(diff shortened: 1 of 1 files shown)


SysPete:
Good catch. A quick scan of the code suggests that `length` could be replaced by `defined` which should be much faster.

MarketaWachtlova:
Of course. You are right.

racke:
Also it would be nice to have a test that exposes this bug.

xsawyerx:
Is this fix really waiting this long just for a test?

racke:
I suppose rather for thumbs ups :-).

cromedome:
👍  I had meant to write the test some time ago as part of the release, but it hasn't happened yet.

xsawyerx:
:+1: Ready for merge.

cromedome:
Merged. Thank you!

PR #1292, opened 2016-11-23
Improve support for "has" in Dancer2::Plugin

nfg:
Hello! I was writing a Dancer2 plugin for $WORK, and couldn't figure out why my attributes weren't pulling their values from the plugin configs. Eventually I figured out it was because I was defining multiple attributes at once in my plugin, a la:

````perl
#!/usr/bin/env perl

package PluginTest;
use Dancer2::Plugin;
has [qw(field1)]    => ( is => 'ro', from_config => 1 ); # THIS BREAKS!
has 'field2'        => ( is => 'ro', from_config => 1 ); # This works

plugin_keywords 'test_fields' => sub {
    my $self = shift;
    foreach my $field (qw(field1 field2)) {
        print STDERR "$field:" . ($self->$field // '(undef)') . "\n";
    }
};

package main;
use Dancer2;
PluginTest->import();

set plugins => {
    "PluginTest" => {
        field1 => 'woo',
        field2 => 'hoo',
    },
};

get '/' => sub {
    test_fields();
    return "TEST";
};
dance;
````

When I run that script and hit localhost:3000/ , I get:

````
➜  Foo perl server.pl 
>> Dancer2 v0.204001 server 14578 listening on http://0.0.0.0:3000
[main:14578] core @2016-11-22 17:49:30> looking for get / in /home/nfg/.plenv/versions/5.20.3/lib/perl5/site_perl/5.20.3/Dancer2/Core/App.pm l. 34
[main:14578] core @2016-11-22 17:49:30> Entering hook core.app.before_request in (eval 115) l. 1
field1:(undef)
field2:hoo
[main:14578] core @2016-11-22 17:49:30> Entering hook core.app.after_request in (eval 115) l. 1
````

This PR should fix the issue.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -342,8 +342,17 @@ sub _exporter_plugin {
             our \@EXPORT = ( ':app' );
 
             around has => sub {
-                my( \$orig, \$name, \@args ) = \@_;
-                \$orig->( ${caller}->_p2_has( \$_, \@args) )
+                my( \$orig, \$name, \%args ) = \@_;
+
+                if (ref \$name eq 'ARRAY'
+                        && exists \$args{'plugin_keyword'}
+                        && ref \$args{'plugin_keyword'} eq 'ARRAY') {
+
+                    Carp::croak('Setting "plugin_keyword" to an array is disallowed'
+                        . ' when defining multiple attributes simultaneously');
+                }
+
+                \$orig->( ${caller}->_p2_has( \$_, \%args) )
                     for ref \$name ? @\$name : \$name;
             };
 
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -342,8 +342,9 @@ sub _exporter_plugin {
             our \@EXPORT = ( ':app' );
 
             around has => sub {
-                my( \$orig, \@args ) = \@_;
-                \$orig->( ${caller}->_p2_has( \@args) );
+                my( \$orig, \$name, \@args ) = \@_;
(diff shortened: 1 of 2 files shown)


nfg:
FWIW, there's weirdness with from_config and plugin_keyword with non-1 values. IE:

````perl
has [qw(foo bar)] => (
  is => 'ro',
  from_config => 'foo',
  plugin_keyword => [qw(foo bar)],
);
````

Both "foo" and "bar" will use the key "foo" for their default value, which is probably OK... but the keywords "foo" and "bar" will both use the "bar" attribute, which is unintuitive. I was thinking of maybe adding a warning to Dancer2::Plugin::_p2_has_keyword if $class->keywords->{$_} already exists, or a warning in the documentation. What do you guys think?

xsawyerx:
:+1:

nfg:
Hello again! Just pinging. Is there anything else you'd like me to do, or is this good?

xsawyerx:
No. It's good. We just need a bit of time. The Advent Calendar is starting tomorrow so we've been pressed for time.

xsawyerx:
As for your question, I think we should not allow defining two attributes at the same time with two plugin keywords attribute.

There can be different conditions checked here, but I think the simplest one would be "If you're defining an attribute with an array, you cannot provide a `plugin_keyword` that is an array."

nfg:
Cool! I've added that check and a quick unit test (shoe-horned into from-config.t).

veryrusty:
Looks good to me  :+1:   Sorry its taken soooooo long to look over it :(

veryrusty:
Cherry-picked onto current master and merged.  @nfg++

Apologies again for it taking soooo long.

PR #1294, opened 2016-12-12
Feature/find plugin

xsawyerx:
Per the conversation with @yanick.

This purposefully doesn't error out, so the plugin developer can decide what kind of error to throw.

I have implemented it both in the App and the Plugin, just in case.

I need it approved quickly, so I can add this to an article.


yanick:
looks good to me. ==b o.o

xsawyerx:
Added documentation. @SysPete, any chance for a :+1: ?

racke:
:+1:

SysPete:
:+1:

xsawyerx:
Merged!

PR #1295, opened 2016-12-14
Fix public_dir option

### LINKED ISSUE #975: public_dir option does not seem to work
The Dancer2::Config POD documents the `public_dir` option, but it doesn't seem to do anything.

After looking at the `Dancer2::Handler::File` code I can't figure out when the Handler object is actually created and registered. In any case, `DANCER_PUBLIC` seems to work (although undocumented), so it might be an issue of loading the config after the handler is registered.

Steps to reproduce:

```
$ dancer2 -a MWE::PublicDirDoesNotWork
$ cd MWE-PublicDirDoesNotWork
$ mv public foo
# edit config.yml, add "public_dir: foo"
$ plackup bin/app.psgi
# visit http://localhost:5000 or wherever your Plack::Handler lives: all requests to static assets 404
```

xsawyerx:
This resolves #975.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,9 @@
 {{$NEXT}}
 
+    [ BUG FIXES ]
+    * GH #975: Fix "public_dir" configuration to work, just like
+      DANCER_PUBLIC. (Sawyer X)
+
     [ DOCUMENTATION ]
     * GH #1282: Typo in Cookbook. (Kurt Edmiston)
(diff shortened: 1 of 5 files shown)


cromedome:
:+1:

xsawyerx:
@PerlDancer/core, any other approvals?

mickeyn:
seems legit

PR #1301, opened 2017-01-05
capture @xsawyerx async code example to doc

mohawk2:
Possibly needs moving elsewhere, maybe to `Manual.pod`?

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -410,6 +410,39 @@ You can handle the error yourself by providing an C<on_error> handler:
         };
     };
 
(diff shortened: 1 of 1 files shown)


xsawyerx:
This class should document the internal core class, not the keyword. I think this should fit under the keyword in the Manual.

@mohawk2++ Thanks for doing this. :)

xsawyerx:
Now seeing it again, I think the note should go with the example, since the example is on streaming a file. I just want the note there to say "While this is streaming a while, `send_file` also streams files and will do it asynchronously, so don't worry" (which your comment phrases better).

xsawyerx:
:+1:

PR #1314, opened 2017-02-08
A few tweaks to the manual

bigpresh:
A few simple doco tweaks in the manual - mostly just adding cross-linking.

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -187,8 +187,8 @@ requests.
 =head2 Route Handlers
 
 The route action is the code reference declared. It can access parameters
-through the specific C<route_parameters|/route_parameters>, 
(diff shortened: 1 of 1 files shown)


cromedome:
:+1: Thanks, @bigpresh!

xsawyerx:
I had rewritten the Manual in #1069 and I want to merge this week. Any chance you could take a look if the bits you added here are missing there?

(When rebasing it yesterday I noticed one incident in which something I added was also added to the doc from the time I branched off, so it has happened at least once before.)

bigpresh:
Let's get #1069 merged, then I'll re-visit this PR and re-apply the tweaks I made, if they still make sense.

SysPete:
Closing since this was merged as part of v0.206000 via ae3abfcaf0be36f4eb2deee0ff5765375d620b81

PR #1322, opened 2017-02-17
Bugfix/deprecate dispatch path

veryrusty:
`request->dispatch_path` was introduced due to a buggy `request->path` implementation. (See #373).  Commit c6af1c0 fixed the `request->path` implementation; but caused `dispatch_path` to incorrectly fail route match when SCRIPT_NAME was also a prefix of the PATH_INFO.

After c6af1c0 was applied; it was safe to dispatch against the corrected `request->path` implementation. This PR deprecates `request->dispatch_path` (warns then calls `request->path`).

This is an alternate approach to #1288. Thanks to @jbarrett for reporting the issue there.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -274,22 +274,8 @@ sub uri_base {
 }
 
 sub dispatch_path {
-    my $self = shift;
-
-    my $path = $self->path;
-
-    # Want $self->base->path, without needing the URI object,
-    # and trim any trailing '/'.
-    my $base = '';
-    $base .= $self->script_name if defined $self->script_name;
-    $base =~ s|/+$||;
-
-    # Remove base from front of path.
-    $path =~ s|^(\Q$base\E)?||;
-    $path =~ s|^/+|/|;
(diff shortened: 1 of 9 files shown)


xsawyerx:
Whoa! I was reading this thinking "oh, we fixed it? Great" and then realized it was me who fixed it. Heh.

racke:
:+1: Good work, @veryrusty !

xsawyerx:
I've looked at this again, and I think we should have a deprecation period. Two releases maybe, but at least one release in which we warn you it's going to happen. We could then make it die later, too, to make sure no one misunderstands.

Do applications that use `dispatch_path` instead of `path` work?

veryrusty:
On 19/2/17 8:17 pm, Sawyer X wrote:

> Do applications that use |dispatch_path| instead of |path| work?
Yes;  currently, provided SCRIPT_NAME is not a prefix of PATH_INFO, then 
`dispatch_path` and `path` return the same. If SCRIPT_NAME *is a prefix* 
of PATH_INFO, you get the broken behaviour this PR fixes.

xsawyerx:
:+1: Ready for merge.

PR #1325, opened 2017-02-24
Test & support multi-value cookies under XS

JRaspass:
The issue is that we join on ampersand **before** escaping the value in the XS path which leads to incorrectly escaped ampersands in the final HTTP header.

A better fix would of course be to work out a calling convention to support multi-value cookies with HTTP::XSCookie, but in the mean time this at least restores working behaviour for people who opt into the XS backend.

### DIFF
--- a/lib/Dancer2/Core/Cookie.pm
+++ b/lib/Dancer2/Core/Cookie.pm
@@ -31,9 +31,12 @@ BEGIN {
 sub xs_to_header {
     my $self = shift;
 
+    # HTTP::XSCookies can't handle multi-value cookies.
+    return $self->pp_to_header(@_) if @{[ $self->value ]} > 1;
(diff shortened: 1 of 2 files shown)


xsawyerx:
:+1: 

Is there a corresponding `HTTP::XSCookies` ticket?

JRaspass:
No ticket yet, I wanted the discussion first, as I don't think multi value cookies are even standardised, I found no reference to them in the RFC :-(

xsawyerx:
I suggest opening a ticket and seeing what @gonzus thinks.

JRaspass:
Done https://github.com/gonzus/http-xscookies/issues/3

xsawyerx:
Merged, thanks! :+1:

Issue #1332, opened 2017-03-11
t/cookie.t may fail (Dancer2-0.205000)


eserte:
On some of my smoker systems t/cookie.t fails:
```
#   Failed test at t/cookie.t line 165.
#          got: 'same-site=strict; Path=/'
#     expected: 'same-site=strict; Path=/; SameSite=Strict'

#   Failed test at t/cookie.t line 165.
#          got: 'same-site=lax; Path=/'
#     expected: 'same-site=lax; Path=/; SameSite=Lax'
# Looks like you failed 2 tests of 120.
t/cookie.t ............................................. 
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/120 subtests 
```
Statistical analysis unfortunately does not give any results --- maybe the difference is caused by different versions of an optional module?

xsawyerx:
This depends on the version of `HTTP::XSHeaders` you have. We can have the test check for it.

eserte:
It's more likely to be an issue with `HTTP::XSCookies`. Here's the statistical analysis (theta=-1 indicates 100% fail):
```
****************************************************************
Regression 'qr:(HTTP..XSCookies.*)'
****************************************************************
Name                   Theta          StdErr     T-stat
[0='const']           1.0000          0.0000    6507504573462532.00
[1='eq_HTTP::XSCookies         0.000007 0.000006']           -1.0000          0.0000    -6068279304399642.00
[2='eq_HTTP::XSCookies         0.000007 0.000007']            0.0000          0.0000       1.51
[3='eq_HTTP::XSCookies         0.000007 missing']             0.0000          0.0000       1.18
```

This means there are only failures with HTTP::XSCookies 0.000006, no failures if 0.000007 or nothing is installed.

xsawyerx:
Pinging @gonzus.

gonzus:
Will check later today. Thanks for the heads up.

xsawyerx:
I read too quickly and pinged @gonzus for no reason.

@eserte, we *do* recommend XSCookies 0.000007, not under. Is there something else you think we can do to resolve this?

cromedome:
Resolved by #1333.

PR #1333, opened 2017-03-12
add HTTP::XSCookies to 'requires' if old version is installed

SysPete:
For #1332 

This works for cpanminus and I assume will work for any other clients which use cpanfile. Can't think of any other way round this so far.


cromedome:
Oooh, clever! :+1:

cromedome:
Merged! Thanks!

PR #1334, opened 2017-03-15
Fix Cookbook prefix example

mpcabd:
Previously the example looked like this

  prefix => '/users' => sub {

but that generates warnings, and I believe it was just a typo.
So I'm removing the => after prefix.


cromedome:
Merged. Thanks for fixing this!

PR #1341, opened 2017-04-02
Dancer2::Plugin is the only plugin class, so clear out the "alternate…

racke:
… plugin" sentence from the POD.


cromedome:
Merged!

PR #1352, opened 2017-06-06
Update +app.psgi

p-stefanov:
Actually don't use middleware in the first block.


ambs:
The bug had three patches, and just one could be merged, so, sorry but closing yours :-)

PR #1354, opened 2017-06-18
Changing logic around testing if a template exists.

kielstr:
Check for error and provide feedback.

### DIFF
--- a/lib/Dancer2/Template/TemplateToolkit.pm
+++ b/lib/Dancer2/Template/TemplateToolkit.pm
@@ -78,11 +78,19 @@ sub layout_pathname {
 
 sub pathname_exists {
     my ( $self, $pathname ) = @_;
-    return eval {
+
+    eval {
         # dies if pathname can not be found via TT2's INCLUDE_PATH search
         $self->engine->service->context->template( $pathname );
         1;
     };
+
+    if ( $@ ) {
+        warn $@;
+        return -1;
+    }
+    
+    return 1;                 
 }
 
 1;


cromedome:
I normally prefer `Try::Tiny` to error checking of `eval{}`, but since we use `eval{}` all over the place I can't really complain 😄 

However, this PR does cause `t/auto_page.t` to fail. I haven't had a chance to see if it's just a poorly written test, or if this change actually breaks the auto page functionality. If you have a chance to investigate and report back/fix, great. I will also try to look at this in the next day or two so we can release this ASAP.

Thanks for your PR! I like it, and will be eager to merge it once we can figure out what the auto page issue is.

veryrusty:
`pathname_exists` should return a false value if the file doesn't exist. This PR alters the return semantics; `1` for file exists and `-1` when not existing.  However `-1` isn't a perl false value.

Something like the following would be more appropriate (logic wise)
```
sub pathname_exists {
    my ( $self, $pathname ) = @_;

    my $exists = eval {
        # dies if pathname can not be found via TT2's INCLUDE_PATH search
        $self->engine->service->context->template( $pathname );
        1;
    };
    warn $@ if ! $exists;
    return $exists;
}
```

It'd be nice to emit the warning via an app's logger too - but that's a little trickier!

kielstr:
Hey all,

Thanks for sorting this. I was down a rabbit hole yesterday.

Kiel R Stirling.

veryrusty:
Merged a variant of this into master via commit e8a9f0c1. This implements the above suggestion to ensure the (boolean) nature of the return is maintained, and logs the missing template (at debug level) via the app's logger.

PR #1357, opened 2017-07-11
Issue in skeleton of file bin/+app.psgi

pierre-vigier:
The plack middleware Deflate was activated by default, and
was not only in the commented section. A small issue as well
with the comment of the third block, which is not related
to Deflate middleware


ambs:
Merged. Thanks.

PR #1359, opened 2017-07-11
Converted the rest of the ref checks to Ref::Util

mickeyn:
So there's the rest of them.

Question: do we want to convert the likes of:
`isa_ok( $app, 'CODE'...` 

I didn't touch those yet (we have many).


bigpresh:
This looks good to me.

PR #1362, opened 2017-07-16
Cookies set by cookie keyword should be httpOnly by default

### LINKED ISSUE #1361: cookies don't default to HTTPOnly, despite the doco
albino reported on #dancer that they were seeing cookies set without HTTPOnly.

https://metacpan.org/pod/Dancer2::Core::Cookie#http_only claims "By default, cookies are 
 created with a property, named HttpOnly, that can be used for security, forcing the cookie to be 
 used only by the server (via HTTP) and not by any JavaScript code. If your cookie is meant to be 
 used by some JavaScript code, set this attribute to 0."

Yet, the http_only attribute defaults to 0 - https://metacpan.org/source/CROMEDOME/Dancer2-0.205001/lib/Dancer2/Core/Cookie.pm#L128-133

So, there's a choice to make here:

* Update the doco to match what the code currently does
* Update the code to do what the doco claimed it did.  This *could* bite some people who were accessing cookies via JS and having it work "by accident", but would be more secure, and align with the behaviour of session engines which *do* set httpOnly by default, so I think I'd prefer this.

I'd like feedback from the rest of the team as to which option they prefer, though.

bigpresh:
Fixes #1361 

We document that we set httpOnly on cookies by default, but we don't, for ones set using `cookie`; we do, however, for session cookies.

This PR makes us set httpOnly by default for cookies - but as per my comment on #1361, I'd like other's opinions on whether changing this behaviour now is reasonable.  There's a slight chance that people could get bitten by it, if they were accessing cookies set via `cookie()` from Javascript and it was working "by accident", but setting httpOnly by default as we documented that we did is more secure, and aligns with that we do for session cookies, so I think this is probably the best approach.

It should be mentioned clearly in the changelog if we go with this, though :)

### DIFF
--- a/lib/Dancer2/Core/Cookie.pm
+++ b/lib/Dancer2/Core/Cookie.pm
@@ -129,7 +129,7 @@ has http_only => (
     is       => 'rw',
     isa      => Bool,
     required => 0,
-    default  => sub {0},
+    default  => sub {1},
 );
 
 has same_site => (
--- a/t/cookie.t
+++ b/t/cookie.t
@@ -68,11 +68,7 @@ sub run_test {
     is $cookie->secure(0) => 0, "disabling \$cookie->secure returns new value";
     ok !$cookie->secure, "\$cookie->secure flag is disabled";
 
-    ok !$cookie->http_only, "no http_only by default";;
-    is $cookie->http_only(1) => 1,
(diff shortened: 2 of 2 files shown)


ambs:
Adding to Milestone. Seems like a simple fix, and just a rewrite of tests.

PR #1365, opened 2017-07-20
Serializer mutable custom mapping

yanick:
Allows Dancer2::Serializer::Mutable to be configured with a different content-type/serializer mapping table.


veryrusty:
I've cleaned up rebased this onto the current master as the base for my work in #1457.

cromedome:
Resolved as part of #1457. Thanks!

PR #1370, opened 2017-08-14
Store route executing in request

bigpresh:
Currently, it seems there's no way for a hook / route handler to inspect which route the request matched, which is a limitation that is biting us at work in attempting to transition from Dancer1 to Dancer2.

It seems silly not to be able to do this - we were able to in Dancer1, and made good use of that ability.

This PR adds a `route` attribute to the D2::Core::Request object, which is populated with the D2::Core::Route object for the route the request matched by the dispatcher.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -653,6 +653,14 @@ sub _shallow_clone {
     return $new_request;
 }
 
+
+sub route {
+    my ($self, $route) = @_;
+    if ($route) { $self->{route_matched} = $route; }
+
+    return $self->{route_matched};
+}
+
 1;
 
 __END__
@@ -1005,6 +1013,10 @@ Alias to the C<method> accessor, for backward-compatibility with C<CGI> interfac
 =method request_uri
 
 Return the raw, undecoded request URI path.
+
+=method route
+
+Return the L<route|Dancer2::Core::Route> which this request matched.
 
 =method scheme
 
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1452,6 +1452,8 @@ DISPATCH:
             my $match = $route->match($request)
                 or next ROUTE;
 
+            $request->route($route);
+
             $request->_set_route_params($match);
             $request->_set_route_parameters($match);
 
--- a/t/dispatcher.t
+++ b/t/dispatcher.t
@@ -21,21 +21,21 @@ $app->setting( logger      => engine('logger') );
 $app->setting( show_errors => 1 );
 
 # a simple / route
-$app->add_route(
+my $simple_route = $app->add_route(
     method => 'get',
     regexp => '/',
     code   => sub {"home"},
 );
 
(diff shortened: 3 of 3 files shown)


lib/Dancer2/Core/Request.pm
```diff
@@ -653,6 +653,14 @@ sub _shallow_clone {
     return $new_request;
 }
 
+
+sub route {
+    my ($self, $route) = @_;
+    if ($route) { $self->{route_matched} = $route; }
+
+    return $self->{route_matched};
```
xsawyerx: I'm not sure this is the best way to accomplish this. This creates a hashref slot without an appropriate attribute and thus introduces state we cannot account for. It also makes `route` a `rw`, which we are trying to avoid doing.

Why can't `route` be a `ro` attribute and that's it?

bigpresh: > Why can't route be a ro attribute and that's it?

Because D2::C::Request isn't a Moo class, and does all the OO itself.  There may be an argument that it *should* be converted into a Moo class, but I'd consider that a separate tech debt issue personally.

xsawyerx: Oh yeah. I did that. Sorry.

It was moved to be a class on its own because it needs to inherit from `Plack::Request`. It can't do that with Moo because it will not allow some things. The commits should provide more information on that.

I would at least make a read-only attribute from this. Let me get home and add a different suggestion branch.

bigpresh: Hmm - we could tweak it so you can only set it once, if that would make you happier. 

I'd be worried, though, that that could cause problems if a request object is kept the same when a route fowards internally, and thus goes through the dispatcher a second time...

Or, I could make the `route()` method just an accessor, and have a private `_set_route()` setter?

I'm not quite sure I see the problem, to be honest?

xsawyerx: State is the problem. Dancer2 is able to optimize some things based on attribute values never changing. We're using it in several places and I want to use it in more. If the state changes here, we have to account for it.

I was surprised by `forward` so I look. The code seems to create a new request. We do a shallow clone of the existing request and then update the cookies. This might be wrong in general. However, it does combine new options, so the new option could be the next route. This means we won't be breaking `forward`.

Let me try and hack on it when I get home.

bigpresh: Isn't any such optimisation in the Moo'd stuff?  I don't see how it applies here - it's just an attribute in a Request object, which will only live as long as the request being processed.

Interested to see what you come up with, though.

xsawyerx: The concept of state is generally useful, even outside Moo.

However, I played with it now and I don't really see how to make this not-rw, since it needs to be set *after* the request. I would still prefer `set_route` instead of `route` as rw method, and `matched_route` as an entry. Mine: https://github.com/PerlDancer/Dancer2/pull/1371

bigpresh:
Closing in favour of @xsawyerx's #1371

PR #1371, opened 2017-08-15
Sawyer/store route executing in request

xsawyerx:
My version of it.


bigpresh:
Right, let's get this merged.

PR #1374, opened 2017-09-09
Delete env var before testing

### LINKED ISSUE #1373: tests fails when DANCER_ENVIRONMENT is set
A lot of tests fail when DANCER_ENVIRONMENT is set...

ambs:
Fixes #1373

### DIFF
--- a/t/classes/Dancer2-Core-Runner/environment.t
+++ b/t/classes/Dancer2-Core-Runner/environment.t
@@ -3,6 +3,10 @@ use warnings;
 use Test::More tests => 6;
 use Dancer2::Core::Runner;
 
+# undefine ENV vars used as defaults for app environment in these tests
+local $ENV{DANCER_ENVIRONMENT};
+local $ENV{PLACK_ENV};
+
 {
     my $runner = Dancer2::Core::Runner->new();
     isa_ok( $runner, 'Dancer2::Core::Runner' );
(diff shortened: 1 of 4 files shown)


veryrusty:
Thanks @ambs!

Dancer2::Core::Runner uses `$ENV{DANCER_ENVIRONMENT} || $ENV{PLACK_ENV} || 'development'` as the default for its `environment` attribute.  If `$ENV{PLACK_ENV}` is set, a similar set of failures would be seen too (`t/config_reader.t` also fails). For completeness/consistency, these tests should also ensure `$ENV{PLACK_ENV}` is not defined.

This may be a personal preference, but I'd prefer to localize those ENV vars; and the `BEGIN` blocks shouldn't be necessary (its a lazy attribute). i.e. use something like.
```
# undefine ENV vars used as defaults for app environment in these tests
local $ENV{DANCER_ENVIRONMENT};
local $ENV{PLACK_ENV};
```

ambs:
@veryrusty sure. Would do it.

ambs:
@veryrusty in one of the tests I really needed to use delete. Localizing wasn't working and I wasn't understanding why, so, gone for delete.

ambs:
@veryrusty do you mind giving this fix a look? Tempted to merge as it is. Thanks.

veryrusty:
@ambs `t/config_reader.t` still fails for me:

```
export DANCER_ENVIRONMENT=foo
export PLACK_ENV=bar
prove -r t
```

We need to localize (or under) those two in that test case (and then merge :)

ambs:
And here it goes.

PR #1382, opened 2017-10-03
Remove middleware boilerplate from basic app

nunorc:
the Deflater middleware is not a dependency requirement for Dancer2. The app fails to run if it is included in the skeleton script.


ambs:
Duplicate fix. Thanks any way :-)

Issue #1386, opened 2017-10-08
Add examples to the repository and maybe even install them as well


szabgab:
e.g. create and examples/ directory and in there
create examples/single for examples that only have a single psgi file and thus don't interfere with each other.
create a separate subdirectory for every other example that has multiple files templates etc.
Add tests for each example

cromedome:
Closed by 63db51bfb83ce18f38f7f219cb90ca37a1db6d58. Thanks!

PR #1387, opened 2017-10-08
Add Hello World example and a test with it #1386

szabgab:


cromedome:
:+1: Thanks @szabgab !

cromedome:
:+1: Merged, thank you!

PR #1389, opened 2017-10-08
Keep config file reading order, independent of file extension

### LINKED ISSUE #1304: Config files are read in sorted order, rather than environment always overriding main config
Playing around with Dancer2, I arranged the configs approximately like this:

    /app/lib/config.yml        # main config
    /app/conf/production.yml   # environment config

I was surprised to find that the environment config wouldn't override the main config. There is a `sort` call in `Dancer2::Core::Role::ConfigReader` that puts them in the unexpected order. I don't see why it should sort them.

I believe it should consistently evaluate files in this order:

- main
- main local
- environment
- environment local

ambs:
Fixes #1304

### DIFF
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -129,22 +129,23 @@ sub _build_config_files {
         }
     }
 
-    foreach my $ext (@exts) {
-        foreach my $file ( [ $location, "config.$ext" ],
-            [ $self->environments_location, "$running_env.$ext" ] )
-        {
-            my $path = path( @{$file} );
-            next if !-r $path;
+    # Keep order of files, even if they have different extensions
+    foreach my $file ( [ $location, "config" ],
+		       [ $self->environments_location, "$running_env" ] )
+      {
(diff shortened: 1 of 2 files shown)


lib/Dancer2/Core/Role/ConfigReader.pm
```diff
@@ -129,22 +129,23 @@ sub _build_config_files {
...
         }
-    }
+      }
 
     # Look for *_local.ext files
     @files = map {
         (my $l = $_) =~ s/(\w+)(\.\w+)$/${1}_local$2/;
         $_, (-r $l ? $l : () );
-    } sort @files;
+    } @files;
```
veryrusty: We should remove the `map` here by putting the logic for `_local.$ext` inside the nested `foreach` loops above.
(I'll clean this up in the next day or two!)

ambs:
Oooooh... this is OOOOOLD!. Will you fix it? Thanks

veryrusty:
Ooooooold it is. Trying to clean up the low hanging fruit in preparation for the next release!

Reworked and merged into master in commit 2c761784.

Thanks @ambs!!

Issue #1380, opened 2017-09-19
Custom error pages at wrong place


niceperl:
I have built a Dancer2 app (version 0.205001) with the command line interface:

`%> dancer2 -a my_app`

The template toolkit was configured in this project.
The custom error pages created by the builder were placed at the "public" folder (`404.html` and `500.html`). Then in the application code, I tried to send a 404 error, but Dancer2 used its own page (and ignored the custom one). 

It works properly if I create a `404.tt` page at "views" folder.
Is this a bug in the CLI or a kind of feature that should be documented?

cromedome:
Merged and fixed!

PR #1391, opened 2017-10-10
Issue #1380 error templates

1nickt:
Fixes bug where the class of error was not taken into account when deciding whether to show or not show a custom static HTML error page with `show_errors` on.

Updates Manual.pod to document the now-working proper behaviour, as is documented already in Cookbook.pod


ambs:
Looks Good!

cromedome:
:+1: thanks!

cromedome:
Merged, thanks!

Issue #1368, opened 2017-08-08
XSS in Dancer2::Tutorial example


diocles:
I noticed there is no HTML escaping in the template in the tutorial:

`<li><h2>[% entries.$id.title %]</h2>[% entries.$id.text %]`

Similarly to the paragraph on SQL injection, I think it would be nice to fix the XSS, and perhaps mention briefly why escaping matters.

guest20:
This could be fixed by having the boiler plate apps use [Template::AutoFilter](https://metacpan.org/pod/Template::AutoFilter), who's default is to html escape content.

It does come with a "unstable until 1.0" warning though.

PR #1392, opened 2017-10-10
don't allow XSS in tutorial

simbabque:
I decided to take the simplest approach possible to fix this, but to include a bit of explanation although it's not a TT tutorial. If SQL injection is explained, this should be explained too. I believe new users are often unfamiliar with how the web works and surely appreciate additional information that they can read if they want to.


cromedome:
Merged, thanks!

cromedome:
Resolves #1368.

simbabque:
Github says there are unmerged commits in my branch. Weird.

Issue #1285, opened 2016-10-20
Add "Default template variables" section to manual


racke:
These variables are documented in https://metacpan.org/pod/Dancer2::Core::Role::Template#DESCRIPTION,
which is out of reach for people reading the manual.

This information should be also available in the manual:
https://metacpan.org/pod/distribution/Dancer2/lib/Dancer2/Manual.pod#TEMPLATES

Bonus points for more detailed descriptions and links to the keywords/objects.

PR #1393, opened 2017-10-10
Add "Default template variables" section to manual

simbabque:
There was already something there about the vars, but that seemed to be outdated and in general way less information.


cromedome:
Merged, thanks!

cromedome:
Resolves #1285.

PR #1394, opened 2017-10-12
Fix some links in the Manual pod

### LINKED ISSUE #1385: Error Handling and File Rendering self link in docs
On https://metacpan.org/pod/distribution/Dancer2/lib/Dancer2/Manual.pod

Error Handling have link that lead to itself. I think because there are two sections that are called `Error Handling`

File Rendering has a link titled `File Rendering` leading to file-handler.

Both seem to be a bit confusing.

ambs:
Fixes #1385 (needs to be properly tested)

### DIFF
--- a/Changes
+++ b/Changes
@@ -5,6 +5,7 @@
     * GH #1373: Unset Dancer environment vars before testing (Alberto Simões)
     * GH #1383: Remove Deflater from default app skeleton (Pierre Vigier)
     * GH #1362: Make cookies http_only by default (David Precious)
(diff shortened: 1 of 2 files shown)


ambs:
@szabgab do you have the patience to generate the HTML file and test? Thanks

ambs:
Added to the milestone for tomorrow, as I think it is not a problematic merge.

cromedome:
:+1:

Went ahead and merged this. I can reopen if it is still a problem later. This looked great to me though!

PR #1395, opened 2017-10-12
add a chapter about advanced customization of TT via subclass of D2::…

simbabque:
This content is based on my Stack Overflow answer https://stackoverflow.com/a/46645835/1331451. I suggested to add this to the D2 documentation and @cromedome approved, so here we go.


cromedome:
Merged, thank you!

PR #1397, opened 2017-10-15
Honour no_server_tokens config in error responses

veryrusty:
For apps configured to NOT include server tokens in responses,
responses on error should do the same thing too. So ..

When the Error object has an app, use its no_server_tokens config setting,
or fall back to $ENV{DANCER_NO_SERVER_TOKENS} (as the Runner does) if
no app is present.

Resolves 1390.


racke:
👍

cromedome:
Looks good! :+1:

cromedome:
Merged, thanks!

PR #1398, opened 2017-10-30
Pr/work on issue 1317

sdeseille:
Hello

I added documentation about serializer configuration in Dancer2::Core::Role::Serializer module and in Dancer2::Config.pod.

I hope to haven't done too much mistake.

regards

sdeseille

### DIFF
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -40,9 +40,9 @@ Next, Dancer2 will look for a file called F<config_local.EXT>. This file
 is typically useful for deployment-specific configuration that should
 not be checked into source control. For instance, database credentials
 could be stored in this file.  Any settings in this file are merged into
-the existing configuration such that those with the same name in your 
-global configuration file will take precedence over those settings in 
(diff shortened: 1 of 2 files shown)


lib/Dancer2/Config.pod
```diff
@@ -227,6 +227,54 @@ extension, and are text files, set the C<default_mime_type> to
 C<text/plain>.
 
 
+=head2 Serializing responses
+
+=head3 serializer (string)
```
sdondley: I can't speak to the content of the patch, so I'll be a doc Nazi and suggest that `(string)` should probably be replaced with something more perlish like ($string). Thanks!

sdeseille: hello, it is really (string) and not $string.

You have to understand (string) has a datatype like per example (integer).
regards

sdeseille

racke:
Indeed, that is consistent with the rest of the POD. This change looks good to me 👍

cromedome:
:+1:

Merged, thank you!

PR #1406, opened 2017-12-06
Remove HTTP::Body from Dancer2::Core::Request

### LINKED ISSUE #1090: Move away from `HTTP::Body`
I've been wanting to do this for a very long time.

`Plack` did this in https://github.com/plack/Plack/pull/537.

veryrusty:
Turned out to be _mostly_ straightforward. Some things to look closely at before merging:

  * While upload filenames are decoded to UTF-8 characters, upload headers are NO longer decoded.
  * Issue #794 arose again while removing HTTP::Body. If I understood whats going on correctly, HTTP::Body would attempt to grok the JSON content and die, causing the 500 response. Instead, this PR locally overrides the `log_cb` code in the serializer to catch the failure while deserializing the request body. Its a ~hack~ workaround - would appreciate others thoughts on better approaches.
  * Bumps minimum Plack version to v1.040 (first release without HTTP::Body).
  * Added a test that creating multiple Request objects from the same PSGI env succeeds. This will hang (indefinitely) if it fails.. Room for improvement there

Wanted to get this out there for review ASAP, it could be a little while before I get back to it. Resolves #1090.


veryrusty:
Anyone else want to comment ? It would be awesome to see this released.

veryrusty:
Merged. Thanks all.

PR #1426, opened 2017-12-31
Move section on modules for extra speed to Deployment

melo:
This section is important to both old Dancer1 users but also to new
Dancer2 users and the latter group would probably not read the
Migration page.

Moved the section from Migration to Deployment, and left a link
pointing to the new placement.

### DIFF
--- a/lib/Dancer2/Manual/Deployment.pod
+++ b/lib/Dancer2/Manual/Deployment.pod
@@ -675,5 +675,38 @@ Launch your application:
 
     plackup -s FCGI --listen /tmp/fcgi.sock bin/app.psgi
 
+
+=head2 Performance Improvements
+
+The following modules can be used to speed up an app in Dancer2:
+
+=over 4
+
+=item * L<URL::Encode::XS>
+
+=item * L<CGI::Deurl::XS>
+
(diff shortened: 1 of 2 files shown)


racke:
Looks good to me 👍. I would also put a link to the performance section into the "DESCRIPTION" section, so it doesn't get overlooked.

melo:
I thought those links were dynamically generated by Metacpan :).

I'll look into this later tonight when I get home and update them then.

racke:
The links are generated on MetaCPAN, that's correct. I think this is an important piece of information, so in my opinion it deserves an extra link in the description.

melo:
I'm confused now... If the links are generated automatically, there is nothing left to fix here. I used a `=head2` tag so the proper link will be generated.

If I add a link to the description, this would be the only section of the document to have it.

If this section is important, maybe we can just move it to the top of the file?

racke:
OK, we keep your original version. Just need one more 👍 for merging.

cromedome:
:+1: I really like this change. I hadn't really looked at it that way before, and this change makes a lot of sense to me.

Merged, thanks!

PR #1430, opened 2018-02-19
Use File::Share on top of File::ShareDir

perlpunk:
File::Share provides its own dist_dir()/dist_file() which also work
in local source directories. Otherwise it will delegate to File::ShareDir
as usual.

Currently the `dancer2 gen` command will not work from the source directory
if there is no Dancer2 installed yet.

See https://perlmonks.org/?node_id=1209472

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -11,7 +11,7 @@ requires 'File::Basename';
 requires 'File::Copy';
 requires 'File::Find';
 requires 'File::Path';
-requires 'File::ShareDir';
+requires 'File::Share';
 requires 'File::Spec';
 requires 'File::Temp';
 requires 'Hash::Merge::Simple';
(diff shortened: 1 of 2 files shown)


xsawyerx:
Need to probably update cpanfile but otherwise, I'm in for merging it.

racke:
I'm for merging it as well :+1:. Thanks for the patch, @perlpunk

perlpunk:
@xsawyerx hm, I *did* update the cpanfile? =)

xsawyerx:
@perlpunk You did and I need to sleep more. :)

perlpunk:
@xsawyerx "sleep"? what's that?

xsawyerx:
I saw this wasn't merged so I went ahead. Sorry for the delay. We will also be releasing a new version soon!

PR #1432, opened 2018-02-23
Support Content-Disposition of inline in send_file()

d5ve:
When called with a "filename" arg, send_file() sets the
Content-Disposition header type to "attachment". This triggers a "Save"
dialogue in some browsers when opening a link in a new tab or window.

+ Add an optional "content_disposition" argument to send_file() which can be
set to "inline" to have the browser display the file inline.

+ Add a note to the manual about this new behaviour.

+ Add tests for "attachment" and "inline" behaviours.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1067,7 +1067,7 @@ sub send_file {
     # content disposition
     ( exists $options{filename} )
       and $self->response->header( 'Content-Disposition' =>
-          ($options{content_disposition} // "attachment") . "; filename=\"$options{filename}\"" );
+          ($options{content_disposition} || "attachment") . "; filename=\"$options{filename}\"" );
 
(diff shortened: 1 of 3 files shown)


d5ve:
The AppVeyor CI failure doesn't seem related to the contents of the PR.

It looks like Strawberry perl can't install Dist::Zilla, so the build fails.

veryrusty:
This looks like a good improvement to me.  @d5ve we _try_ and support perl 5.8 if its at all possible.
Any change you could modify the PR to remove the use of exclusive or (`//`) ?

d5ve:
I changed from `//` to `||` which has basically the same functionality in this case.

cromedome:
@d5ve - I really like this improvement. Thanks for the change to help us support 5.8 as well. This gets a :+1: from me.

SysPete:
Nice change :+1:

cromedome:
Merged, thank you!

PR #1434, opened 2018-03-05
Validate Session ID's

veryrusty:
Add a `validate_id` method to SessionFactory that can "reject" (return false) if the given ID is not considered a suitable string for the session engine to go fetch from the store.

The default pattern is `m/^[A-Za-z0-9_\-~]+$/`, which covers Base64 (for url applications) encoding plus the `~` character that Session::Cookie uses to combine the elements that make up its cookie string.

### DIFF
--- a/lib/Dancer2/Core/Role/SessionFactory.pm
+++ b/lib/Dancer2/Core/Role/SessionFactory.pm
@@ -149,6 +149,11 @@ sub create {
     }
 }
 
+sub validate_id {
+    my ($self, $id) = @_;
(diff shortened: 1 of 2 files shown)


SysPete:
:+1: looks good. I've tested against all non-core session engines based on their current GH master branches and all are good. Thanks @veryrusty!

veryrusty:
@SysPete Thanks for testing all the non-core session engines !

veryrusty:
Rebased and merged to master. Thanks for taking a look @SysPete

PR #1436, opened 2018-03-08
Missing prerequisite 20180308

jkeenan:


veryrusty:
Thanks for taking the time to put together the PR @jkeenan.

The only place `XML::Simple` is mentioned in the Dancer2 codebase is within documentation mentioning the non-core XML serializer. As such, I don't see why XML::Simple should be a dependency for Dancer2.  Closing PR without merging.

PR #1438, opened 2018-03-12
Handle XSCookies to return an arrayref of values

### LINKED ISSUE #1435: t/request.t fails due to change in HTTP::XSCookies?
t/request.t  fails with ```Use of uninitialized value $value in split at /root/.cpanm/work/1520488903.54892/Dancer2-0.205002/lib/Dancer2/Core/Request.pm line 599.```

t/request.t sets cookies as follows:
``` HTTP_COOKIE          => 'cookie.a=foo=bar; cookie.b=1234abcd; no.value.cookie;',```
Where ```no.value.cookie``` looks to be an intentionally bad value, with a possible expectation that  HTTP::XSCookies will discard it (see #1087) .  

Given the following snippet:
```perl
use strict;
use warnings;
use Data::Dumper;
use HTTP::XSCookies 'crush_cookie';

my $str = 'cookie.a=foo=bar; cookie.b=1234abcd; no.value.cookie;'; #undef; #'lalala';
my $cookies = crush_cookie($str);

print Dumper $cookies;
```

Using HTTP::XSCookies version 0.000014 will return:
```
$VAR1 = {
          'cookie.a' => 'foo=bar',
          'cookie.b' => '1234abcd'
        };
```

And using HTTP::XSCookies version 0.000016 will return:
```
$VAR1 = {
          'cookie.b' => '1234abcd',
          'no.value.cookie' => undef,
          'cookie.a' => 'foo=bar'
        };
```
The change in HTTP::XSCookies is from issue [crush_cookie bug? #5](https://github.com/gonzus/http-xscookies/commit/01e36b20c0e7d48d3104977a975538c12617b72d)

I'm not entirely sure if it is more appropriate to open the request here or against HTTP::XSCookies, but the documentation for the crush_cookies sub says, "Parse a (properly encoded) cookie string into a hashref with the individual values" which would indicate that it shouldn't be used to validate the cookie string.

veryrusty:
Both Cookie::Baker and Cookie::Baker::XS return a string for the cookie
value, for which we then split on `&` for handling multiple values. The
alternative XS implementation for cookie crushing, HTTP::XSCookies does
the split, returning an arrayref.

Update the cookie object construction code to accept a string (which we
split), or an arrayref of values (which we leave alone). Allows use of
Cookie::Baker(::XS) or HTTP::XSCookies with minimal code to maintain.

Closes #1435.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -598,7 +598,8 @@ sub _build_cookies {
     while (my ($name, $value) = each %{$cookies}) {
         $cookies->{$name} = Dancer2::Core::Cookie->new(
             name  => $name,
-            value => [split(/[&;]/, $value)]
+            # HTTP::XSCookies v0.17+ will do the split and return an arrayref
+            value => (is_arrayref($value) ? $value : [split(/[&;]/, $value)])
         );
     }
     return $cookies;


veryrusty:
Thanks @SysPete - merged.

PR #1443, opened 2018-04-04
Update copyright year

frazjp65:


cromedome:
I made the change in `Dist::Zilla` so this will get propagated everywhere we need the copyright to appear. I have credited you with the change. Thank you for bringing this to our attention! :+1:

PR #1447, opened 2018-04-10
Added missing build requires as suggested by CPANTS.

manwar:
Hi Team Dancer2,

Please review the PR.

Many Thanks.
Best Regards,
Mohammad S Anwar

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -81,9 +81,9 @@ test_requires 'Test::Fatal';
 test_requires 'Test::More';
 test_requires 'Test::More', '0.92';
 
-build_requires 'Test::CPAN::Meta'
-build_requires 'Test::NoTabs'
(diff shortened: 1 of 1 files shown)


veryrusty:
Thanks @manwar. 

The test cases that use `Test::NoTabs` and `Test::Pod` are both begin with `author-` and are gated with logic such that they WONT run unless the auth is testing, viz:
```
BEGIN {
  unless ($ENV{AUTHOR_TESTING}) {
    print qq{1..0 # SKIP these tests are for testing by the author\n};
    exit
  }
}
```

As such, I think these should be listed as `author_requires`, not `build_requires`.

cromedome:
:+1: Thank you!

cromedome:
Merged. Thanks again!

PR #1450, opened 2018-04-21
Add calculator example with appropriate tests

szabgab:


cromedome:
Nice illustration of using the various parameter methods. Treating the example as docs, so merging this with my :+1: on it. 

Thanks!

szabgab:
Thanks. It is the result of a Pair Programming session of @manwar and myself.

PR #1451, opened 2018-04-23
Added set-engines documentation - Close #1417

### LINKED ISSUE #1417: Doc bug: Configuring serializers
Common enough case, especially during prototyping: Most content going out is JSON, with a few items returning text. using JSON as a global serializer works for all but non-ref scalars, which given an error indicating that I should set "allow_nonref" to true.

Catch: Nothing in the JSON serializer doc tells me how to set this in config or code.

I'd suggest adding one snippet to the JSON doc:

```
=head2 Configuring the JSON Serializer using "set engines"

The JSON serializer optoins can be configured via "set engines". The most
common settings are:

=over 4

=item   allow_nonref

Ignore non-ref scalars returned from handlers. With this set the "Hello, World!"
handler returning a string will be dealt with properly.

=back

Set engines should be called prior to setting JSON as the serializer:

 set engines =>
 {
     serializer =>
     {
         JSON =>
         {
            allow_nonref => 1
         },
     }
 };

 set serializer      => 'JSON';
 set content_type    => 'application/json';

=head2 Returning non-JSON data.

Handlers can return non-JSON via "send_as", which overrides the default serializer:

 get '/' =>
 sub
 {
     send_as html =>
     q{Welcome to the root of all evil...<br>step into my office.}
 };

Any other non-JSON returned format supported by 'send_as' can be used.
```

deirdreemoran:
Added recommended documentation as per lembark's ticket #1417

### DIFF
--- a/lib/Dancer2/Serializer/JSON.pm
+++ b/lib/Dancer2/Serializer/JSON.pm
@@ -82,9 +82,9 @@ Deserializes a JSON string into a Perl data structure.
 
 
 
-=head2 Configuring the JSON Serializer using "set engines"
+=head2 Configuring the JSON Serializer using C<set engines>
 
-The JSON serializer optoins can be configured via "set engines". The most
+The JSON serializer options can be configured via C<set engines>. The most
 common settings are:
 
 =over 4
@@ -114,7 +114,7 @@ Set engines should be called prior to setting JSON as the serializer:
 
 =head2 Returning non-JSON data.
 
-Handlers can return non-JSON via "send_as", which overrides the default serializer:
+Handlers can return non-JSON via C<send_as>, which overrides the default serializer:
 
  get '/' =>
  sub
--- a/lib/Dancer2/Serializer/JSON.pm
+++ b/lib/Dancer2/Serializer/JSON.pm
@@ -79,3 +79,48 @@ Serializes a Perl data structure into a JSON string.
 =method deserialize($content)
 
 Deserializes a JSON string into a Perl data structure.
+
+
+
+=head2 Configuring the JSON Serializer using "set engines"
+
+The JSON serializer optoins can be configured via "set engines". The most
+common settings are:
+
+=over 4
+
+=item   allow_nonref
+
+Ignore non-ref scalars returned from handlers. With this set the "Hello, World!"
+handler returning a string will be dealt with properly.
+
+=back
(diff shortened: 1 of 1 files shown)


deirdreemoran:
Added recommended documentation as per lembark's ticket #1417

lib/Dancer2/Serializer/JSON.pm
```diff
@@ -79,3 +79,48 @@ Serializes a Perl data structure into a JSON string.
 =method deserialize($content)
 
 Deserializes a JSON string into a Perl data structure.
+
+
+
+=head2 Configuring the JSON Serializer using "set engines"
+
+The JSON serializer optoins can be configured via "set engines". The most
```
cromedome: Rather than wrapping set engines in quotes, wrap it in `C<>`. `perldoc` will know to format `C<set engines>` as code when it sees it.

lib/Dancer2/Serializer/JSON.pm
```diff
@@ -79,3 +79,48 @@ Serializes a Perl data structure into a JSON string.
...
+         },
+     }
+ };
+
+ set serializer      => 'JSON';
+ set content_type    => 'application/json';
+
+=head2 Returning non-JSON data.
+
+Handlers can return non-JSON via "send_as", which overrides the default serializer:
```
cromedome: Same here. Use `C<>` rather than quotation marks.

lib/Dancer2/Serializer/JSON.pm
cromedome: Also, `options` is misspelled.

cromedome:
Welcome to the Dancer community! Thanks for submitting this - it requires a couple of very minor tweaks and I will merge this. 

FYI, these are things I'd normally have just fixed when merging, but you mentioned you were a beginner, so I wanted to give you a chance to see some back and forth on a PR.

Again, thank you!

deirdreemoran:
Thanks for letting me learn!  I think this is all set, but let me know if not, thanks for helping me!

cromedome:
:+1: Perfect! Thank you so much... just merged this.

I will try to be more proactive about tagging issues as Beginner Suitable. You are free to tackle others as well.

Thanks again for your contribution, and welcome to the Dancer community!

PR #1454, opened 2018-05-12
Fix typos

gilmagno:
Simple typos


cromedome:
Merged, thanks for finding and fixing those!

PR #1456, opened 2018-06-19
Engines should respect log level

### LINKED ISSUE #1328: Can not set loglevel for session module
I use Dancer2 version 0.204004 

I have written my own session module Dancer2::Session::Shibboleth
It uses the interface 'Dancer2::Core::Role::SessionFactory';

To log i use `$self->log_cb` for instance

```
sub _retrieve {
    my ( $self, $id ) = @_;
	$self->log_cb->("debug", "shibboleth Session _retrieve $id");
```

The loglevel I set in the app has no effect

```
set 'logger'       => 'console';
set 'log'          => 'info';
```

I traced back the problem to Dancer2::Core::App in the methode  _build_session_engine
I replaced the line
        `log_cb => sub { $weak_self->logger_engine->log(@_) },`
with
        `log_cb => sub { $weak_self->log(@_) },`
and then is seems to work ok

there is potentialy the same issue with `_build_template_engine` and `_build_serializer_engine`

pdl:
This hopefully fixes #1328. The logger callback log_cb which is set on the engines should include the code which checks that only messages which are at a sufficiently high log level are actually logged.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -244,7 +244,7 @@ sub _build_session_engine {
         %{$engine_options},
         postponed_hooks => $self->postponed_hooks,
 
-        log_cb => sub { $weak_self->logger_engine->log(@_) },
+        log_cb => sub { $weak_self->log(@_) },
     );
 }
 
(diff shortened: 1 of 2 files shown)


veryrusty:
👍 Looks good to me.  Thanks @pdl !

SysPete:
:+1: Looks good. Thanks @pdl

veryrusty:
Merged. Thanks again @pdl  !

PR #805, opened 2014-12-09
Make Serializer::Mutable very mutable

dboehmer:
As part of my working on fixing GH #795 I see an opportunity to make this module capable of using mixed input/output formats.

This might not be the most common case but while rewriting the test it just happened and here you have the complete feature incl. test and updated documentation. Hope you like it.

PR #1457, opened 2018-06-20
Serializer mutable with custom mapping

### LINKED ISSUE #901: Dancer2 appears to remember the response serializer from one request to the other: if no serializer is set the last one set is remembered
when running https://github.com/PerlDancer/Dancer2-Plugin-REST/blob/master/t/02_prepare_serializer_for_format.t :

the last pair of tests (output and content type for path => '/') fail because the response is generated with the serializer set for the previous request ( path => '/foo.yml', content_type => qr'text/x-yaml')

the solution I found was to
  delete $dsl->response->{serializer}   (
when path is '/' but it might be a problem that the response serializer survives requests

### LINKED ISSUE #973: Mutable and other serializers cannot deserialize different content_types
Since 598ffa7922014b47158954a87e1b90f3516e26bb Dancer2::Serializer::Mutable (or other serializers) is no longer able to deserialize multiple content_types.

Previously, when Dancer2::Core::Request was about to deserialize, it first made a call to Serializer->support_content_type($ct)
Mutable took this opportunity to set its own content type to that so that when it came to deserialize, it knew the content type.
This is obviously quite hacky but I assume it worked (I've not tested that version).

https://github.com/PerlDancer/Dancer2/blob/8d544721bf9e706cdfb9c160ab75a63a9c579a99/lib/Dancer2/Core/Request.pm#L288

Currently, Mutable cannot deserialize anything but the default content type application/json

I would suggest the the serializer be passed the request's content_type as a parameter in the deserialize method.

veryrusty:
Its about time the mutable serializer "just worked".  This PR pulls together several existing branches and PR's to ensure the mutable serializer does indeed work as it should, but maintains back-compatibility.

* Built from a cleaned-up (and rebased) version of @yanick's PR #1365 that allows for customising the content-type/serializer mapping table
* Includes the tests by @dboehmer from #805 ensuring intended fixes the mutable serializer are indeed fixed.
* Removes the unused `support_content_type` method (see #973).
* Add the request object to the current serializer (if we hadn't done so already), resolving the chicken-and-egg issues that the mutable serializer needs the request headers to function. This should also help resolve #973. May need @shumphrey's feedback on that. 

Also resolves #901.

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -199,6 +199,10 @@ sub deserialize {
         $serializer_fail = $_[1];
         $serializer_log_cb->(@_);
     };
+    # work-around to resolve a chicken-and-egg issue when instantiating a
+    # request object; the serializer needs that request object to deserialize
+    # the body params.
+    $self->serializer->has_request || $self->serializer->set_request($self);
     my $data = $serializer->deserialize($body);
     die $serializer_fail if $serializer_fail;
 
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -202,7 +202,8 @@ sub deserialize {
     # work-around to resolve a chicken-and-egg issue when instantiating a
     # request object; the serializer needs that request object to deserialize
     # the body params.
-    $self->serializer->has_request || $self->serializer->set_request($self);
+    Scalar::Util::weaken( my $request = $self );
+    $self->serializer->has_request || $self->serializer->set_request($request);
     my $data = $serializer->deserialize($body);
     die $serializer_fail if $serializer_fail;
 
--- a/lib/Dancer2/Serializer/Mutable.pm
+++ b/lib/Dancer2/Serializer/Mutable.pm
@@ -6,7 +6,9 @@ use Carp 'croak';
 use Encode;
 with 'Dancer2::Core::Role::Serializer';
 
-has '+content_type' => ( default => sub {'application/json'} );
+use constant DEFAULT_CONTENT_TYPE => 'application/json';
+
+has '+content_type' => ( default => DEFAULT_CONTENT_TYPE() );
 
 my $serializer = {
     'YAML'   => {
@@ -60,7 +62,7 @@ sub serialize {
     my ( $self, $entity ) = @_;
 
     # Look for valid format in the headers
-    my $format = $self->_get_content_type();
+    my $format = $self->_get_content_type('accept');
 
     # Match format with a serializer and return
     $format and return $serializer->{$format}{'to'}->(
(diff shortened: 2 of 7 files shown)


lib/Dancer2/Core/Request.pm
```diff
@@ -199,6 +199,10 @@ sub deserialize {
         $serializer_fail = $_[1];
         $serializer_log_cb->(@_);
     };
+    # work-around to resolve a chicken-and-egg issue when instantiating a
+    # request object; the serializer needs that request object to deserialize
+    # the body params.
+    $self->serializer->has_request || $self->serializer->set_request($self);
```
shumphrey: Request has a serializer, and serializer has the request.

The request attribute in Dancer2::Core::Role::Engine isn't a weak_ref.

Could that cause a memory leak?

veryrusty: Possible but unlikely; the code in `Dancer2::Core::App->dispatch()` is where the request object is created then sets the request object into each defined engine without weakening the request object ref; that has no known memory leak.

Line 205 is above pre-empting a call to `$self->set_request($request, $defined_engines);` ( currently line 1444 of Dancer2::Core::App ) to resolve the chicken-and-egg issue.

xsawyerx: `Dancer2::Core::App` weakens a copy when it creates the engines. Check `_build_serializer_engine`, `_build_session_engine`, and `_build_template_engine`. You will find it there.

veryrusty: We were already setting the request in the serializer instance as part of the app's dispatch, which hadn't caused any known problems, hence I wasn't concerned. Pretty trivial to weaken it to ensure it never bites us though.

Added commit 57834d3 to weaken the request object in *both* places is was being set in any engine instance.

lib/Dancer2/Serializer/Mutable.pm
```diff
@@ -63,47 +76,53 @@ sub serialize {
...
-                return $formats->{$value};
+    my ($self, $header) = @_;
+
+    if ( $self->has_request ) {
+        # Search for the first HTTP header variable which specifies
+        # supported content. Both content_type and accept are checked
+        # for backwards compatibility.
+        foreach my $method ( $header, qw<content_type accept> ) {
+            if ( my $value = $self->request->header($method) ) {
+                if ( my $serializer = $self->mapping->{$value} ) {
```
bigpresh: Should we compare case-insensitively here, just for Postel's sake, in case someone said e.g. `Content-Type: text/HTML` or similar?  (They shouldn't, but it's nice to just DYWM sometimes).

lib/Dancer2/Serializer/Mutable.pm
```diff
@@ -31,25 +25,44 @@ my $serializer = {
...
 
-        return 1;
-    }
 
-    return 0;
-}
+        return {
+            'text/x-yaml'        => 'YAML',
+            'text/html'          => 'YAML',
+            'text/x-data-dumper' => 'Dumper',
```
bigpresh: I do wonder whether Dumper should be in the default mapping at all - I imagine its use in real-world Dancer apps is pretty unlikely, and IIRC there's some potential security issues with deserializing untrusted data with it (although Dancer2::Serializer::Dumper does use Safe.pm to do the eval in a compartment, but it still makes me nervous).  I think I'd be in favour of just dropping it - it's easy enough to add back in if required!

t/serializer_mutable_custom.t
```diff
@@ -0,0 +1,126 @@
...
+                    my $content = $req->content;
+                    $content =~ s/\s//g;
+                    is( $req->code, 200, "status" );
+                    is( $content, $s->{last_val}, "content" );
+                }
+            }
+        }
+    }
+
+}
```
bigpresh: The code repetition here irks me somewhat - feels like the original test script could just create two apps, one that has custom stuff in its mappings and one that doesn't, then loop over them doing all the tests, so no need for a second file of copy & pasted code.  Or, perhaps, even simpler, just provide a custom mapping and test that it all works, then test the default mapping is returned as expected separately - or even don't do that, as we can trust a Moo attribute's default to work properly :)

cromedome:
I am merging this and moving forward, as I think this is important enough to get in the next release (as this has been needed/desired for some time). However, I think the concerns @bigpresh has raised are important and need to be addressed when possible, and to that end, I have created #1488 to ensure that these don't get neglected.

Thanks to everyone for their contributions in making this happen!

cromedome:
:+1: from me. Merged in at 56029527e2816f14c5e3b85f7f06dd6c5e0c2088.

PR #1458, opened 2018-06-20
Allow layout_dir to by configured by set keyword

### LINKED ISSUE #1427: Changing layout_dir doesn't work
Changing layout_dir with `set layout_dir` doesn't work for me (I don't use any config file). I found a workaround by adding following code at Core/App.pm line 269 : 

`$engine_attrs->{layout_dir} ||= $config->{layout_dir} || path( $self->location, 'layouts' );`

I'm not sure if it's really a bug or not.

veryrusty:
Ensure `set layout_dir => 'foo';` updates config and the app's template engine if its been instantiated.

  * `layout_dir` attribute in Role::Template is now read-write
  * `config->{layout_dir}` passed to template engine when being instantiated
  * Added a trigger to update an existing template engine's `layout_dir` attribute.

Resolves #1427.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -153,6 +153,12 @@ has '+local_triggers' => (
                 $self->template_engine->layout($value);
             },
 
+            layout_dir => sub {
+                my $self  = shift;
+                my $value = shift;
+                $self->template_engine->layout_dir($value);
+            },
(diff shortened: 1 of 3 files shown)


cromedome:
Nice work! :+1: I approve!

racke:
Looks good to me as well :+1:

Issue #1410, opened 2017-12-07
ETag support via middleware failed for HEAD requests.


amaxcz:
Please add the support for ETags? Also, there is should be a way to specify params for Plack::Middleware::ETag

`
--- App.pm.orig 2017-03-10 22:38:06.000000000 +0100
+++ App.pm      2017-12-06 19:48:35.582374394 +0100
@@ -17,6 +17,7 @@
 use Plack::Middleware::Head;
 use Plack::Middleware::Conditional;
 use Plack::Middleware::ConditionalGET;
+use Plack::Middleware::ETag;

 use Dancer2::FileUtils 'path';
 use Dancer2::Core;
@@ -1416,6 +1417,9 @@
         );
     }

+    # Add experimental ETag support.
+    $psgi = Plack::Middleware::ETag->wrap( $psgi );
+
     # Apply Head. After static so a HEAD request on static content DWIM.
     $psgi = Plack::Middleware::Head->wrap( $psgi );
     return $psgi; 
`

amaxcz:
Suggested fix is here because now, Dancer2 isn't ready to serve Etags for HEAD requests (even via builder). Because HEAD requests drops body, so etags calculation for HEAD requests is not working absolutely. That is a problem.

amaxcz:
also we SHOULD calculate Etags before ConditionalGET. and before dropping body on the HEAD requests.. 
so... Please implement ETag support, it's important.

amaxcz:
Linked bugs are:

Issue #600, opened 2014-06-04
Separate the middleware to a builder


xsawyerx:
In a discussion on plugins (#599), we raised the idea of using a `Plack::Builder` all the time, and then allow plugins to add middlewares.

I know exactly how I want to implement it, in this case:
Have a Builder attribute and add `handles` to the `add_middleware` method and have that available in the DSL.

This is a two part process. This issue is the first one. I'm opening another issue for the second one.

xsawyerx:
Renewed position: Middlewares are not application level and we shouldn't confuse them. Closing this issue. IF someone still sees this as a valued enhancement, please reopen.

Thanks!

PR #1459, opened 2018-06-26
Feature no default middleware

veryrusty:
Add a `no_default_middleware` config option that when set to a true value, does not wrap Dancer2 apps in the following default middleware:

  * Plack::Middleware::FixMissingBodyInRedirect
  * Plack::Middleware::Head

These are both "nice to have" and completely sane for general use. However there are some cases where a developer may want to manipulate the response headers for say a HEAD request in their PSGI stack; but as we drop the request content (via Plack::Middleware::Head) its currently not possible to do so.

This `no_default_middleware` approach was based on @xsawyerx comment of "Middlewares are not application level and we shouldn't confuse them" (see https://github.com/PerlDancer/Dancer2/issues/600#issuecomment-248428286). Going this way also resolves issues such as #1410.

### DIFF
--- a/lib/Dancer2/Config.pod
+++ b/lib/Dancer2/Config.pod
@@ -172,6 +172,19 @@ hard-configured in the proxy if possible. For Apache this would be:
     RequestHeader set X_FORWARDED_PROTO "https"
     RequestHeader set X_FORWARDED_HOST "www.example.com"
 
+=head3 no_default_middleware (boolean)
+
+If set to true, your Dancer2 application will B<NOT> be wrapped with the default
+PSGI middleware. The default middleware wrappers are:
+
+=over 4
+
+=item * L<Plack::Middleware::FixMissingBodyInRedirect>
+
+=item * L<Plack::Middleware::Head>
+
+=back
+
 =head2 Content type / character set
 
 =head3 content_type (string)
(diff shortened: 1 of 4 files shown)


xsawyerx:
In this case, we determined some middlewares to be implementations of features we want in the web framework, but do not want to write it ourselves. Kind of like how you decide you see a plugin you want in core and just make it core, despite it written as a plugin.

Both `Head` and `FixMissingBodyInRedirect` make sense to be part of core because they're not extra functionality, they're part of what the HTTP spec says. (I thought of implementing the latter in core to prevent another subroutine wrapping for each request - which is not a bad thing. The former could also fit.)

veryrusty:
I _think_ we are stating the same thing different ways. Agreed that both `Head` and `FixMissingBodyInRedirect` make sense to be part of core by default. This PR doesn't change that.

If you are building a PSGI app for which Dancer2 is just one part; there may be situations where you require other middleware to act on the response *before* Plack::Middleware::Head is applied. Adding eTag headers is one example; you can't calculate a (weak) eTag in middleware wrapping a Dancer2 app for a HEAD request once the content has been removed by our current use of Plack::Middleware::Head.

Adding more middleware into our stack isn't the way to go; so this PR goes the other way and allows the two middleware we use by default to be removed.

cromedome:
I like this, @veryrusty! :+1:

Might I suggest (in a separate PR) that we add some examples of ETags and ContentMD5 in the cookbook? You cover the why in this PR very well, it might be nice to show people how this change is best used.

PR #1461, opened 2018-06-26
Relax redirect() to match RFC 7231

### LINKED ISSUE #1460: redirect '?foo=bar+%26+baz' does wrong thing
When one feeds a relative URL to `redirect` Dancer2 attempts to make it an absolute URL. It however has a bug in that the URL is URI decoded first, causing `?foo=bar+%26+baz` to become `...?foo=bar+&+baz` which obviously means something VERY different.

It does this because, as the [comment in the code](https://github.com/PerlDancer/Dancer2/blob/master/lib/Dancer2/Core/App.pm#L1272) points out, RFC 2616 requires the URL in a Location header to be absolute. However [RFC 2616 is dead](https://www.mnot.net/blog/2014/06/07/rfc2616_is_dead) and this rule has been relaxed in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.2), some even go as far as to suggest that that [was a bug in RFC 2616](https://stackoverflow.com/questions/8250259/is-a-302-redirect-to-relative-url-valid-or-invalid).

I therefore suggest we remove the offending code and add a test case which will simplify the code and make it more correct :-P

I am of course more than happy to write the commit, I just wanted to get some consensus first, thoughts @xsawyerx and co? Ran into this bug at $day_job.

JRaspass:
Remove code in redirect() that tried to ensure we only return full
URLs with a host as RFC 2616 required. This RFC has since been
superceeded by RFC 7231 which relaxes this rule.

By deleting this code we both simplify the code greatly, win, and
fix a bug where URI encoded hostless URLs were decoded prior to
sending.

Fixes #1460

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1273,16 +1273,6 @@ sub redirect {
     my $destination = shift;
     my $status      = shift;
 
-    # RFC 2616 requires an absolute URI with a scheme,
-    # turn the URI into that if it needs it
-
-    # Scheme grammar as defined in RFC 2396
-    #  scheme = alpha *( alpha | digit | "+" | "-" | "." )
-    my $scheme_re = qr{ [a-z][a-z0-9\+\-\.]* }ix;
-    if ( $destination !~ m{^ $scheme_re : }x ) {
-        $destination = $self->request->uri_for( $destination, {}, 1 );
-    }
-
     $self->response->redirect( $destination, $status );
 
     # Short circuit any remaining before hook / route code
--- a/t/redirect.t
+++ b/t/redirect.t
@@ -15,6 +15,7 @@ subtest 'basic redirects' => sub {
         get '/bounce'   => sub { redirect '/' };
         get '/redirect' => sub { response_header 'X-Foo' => 'foo'; redirect '/'; };
         get '/redirect_querystring' => sub { redirect '/login?failed=1' };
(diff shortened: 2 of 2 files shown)


SysPete:
Thank you @JRaspass :+1: from me. @cromedome?

JRaspass:
I'd forgotten about this change, the diff reads a lot better with "Hide whitespace changes", I could try and redo this patch without the whitespace change if that's preferred? I presume my editor did something by default, sorry about that.

SysPete:
@JRaspass would be great if you can easily get rid of the whitespace changes. Always best to have smallest possible changes. Don't worry if you don't have time.

JRaspass:
Amended and rebased.

SysPete:
@JRaspass thank you!

cromedome:
Looks great! :+1: from me. Thank you!

cromedome:
Merged. Thank you!

roberto-from-infn:
Hello, I m just a final user, but I'd like to notify that after this modification my redirect's don't work anymore.
I mount the app in a sub-directory and use apache + starman with 

ProxyPass /bee/ http://localhost:5252
ProxyPassRevers /bee/ http://localhost:5252

When the program says:

 redirect '/login';

I get https://myhost.mydomain/login ----> NOT found 
where before I used to get the correct URL 
https://myhost.mydomain/bee/login

Do I have to add uri_for to all my redirect? This is not documented anywhere...
Thank you for your answer.

PR #1470, opened 2018-10-01
use session for flash and explain in detail

### LINKED ISSUE #1465: Slight error in documentation tutorial
Hi there,

While working through the Dancr example at https://metacpan.org/pod/distribution/Dancer2/lib/Dancer2/Tutorial.pod I came accross a slight error (if you can call it that)

Instead of a verbatim copy of the completed file I worked my way through and typed the code as I went along. During this "work in progress" a function called &get_flash() is called, but it is never defined.

Reading the completed file it is actually there... but not in the expanation above.

Thought you might wanted to know this

Regard and keep rocking

Peter

simbabque:
When we use a file-wide lexical variable as a closure in the flash functions, we risk that a different user sees the flash message because it's shared between all users. This would occur if a different user's request happened between the time the server returns the 302 response and their browser following it. Saving it to the session removes this problem, which would be hard to explain to a beginner and gives a chance to explain what a session is in more detail.

Closes #1465

### DIFF
--- a/lib/Dancer2/Tutorial.pod
+++ b/lib/Dancer2/Tutorial.pod
@@ -249,9 +249,33 @@ database, then we set a flash message for the user and redirect her back to
 the root URL.
 
 It's worth mentioning that the "flash message" is not part of Dancer2, but a
-part of this specific application.
+part of this specific application. We need to implement it ourself.
 
-=head1 Logins and sessions
+  sub set_flash {
+      my $message = shift;
+ 
+      session flash => $message;
+  }
+ 
+  sub get_flash {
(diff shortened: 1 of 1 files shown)


cromedome:
:+1: Thank you for doing this!

@SysPete - I think that is a great idea. What would you think of adding a "Further Reading" or some such section to the tutorial that provides links to modules and other sites that explain some of the tutorial topics in further detail? It might work in conjunction with (or replace) the See Also section that is currently there.

simbabque:
Sure, I can do that. Sounds like a good idea.

PR #1471, opened 2018-10-01
Don't die if our session dir disappears mid-run (fixes #1462)

### LINKED ISSUE #1462: existence of sessions directory is only checked at start-up?
This app.psgi:

```
use Dancer2;
set session => 'YAML';

get '/' => sub {
    my $count = session('counter') // 0;
    return 1;
};
```
I run it as `plackup app.psgi`

```
curl http://localhost:5000
```
This creates the 'sessions' subdirectory without any content.

If I had a 'sessions' subdirectory when I start the application and I remove it and only then I access the page:

```
rm -rf sessions
curl http://localhost:5000
```

Then the directory is not created.

simbabque:


SysPete:
Part of me wants to scream out "don't delete your sessions dir" but I guess trying to protect user from themselves is not a bad thing so :+1: from me.

racke:
Actually I think it is not worth the performance hit (extra system call on each session retrieval) for a real corner case. So :-1:

cromedome:
I am in the middle on this one. I like to help our users not shoot themselves in the foot, but at the same time, there's the performance issue to consider that @racke raised. Is there a middle ground here? Should we use a configuration setting or environment variable to toggle this functionality?

Thanks for the PR, @simbabque! I appreciate you taking the time to submit this.

simbabque:
I agree with the performance considerations that @racke has raised. I thought long and hard if there's another way without introducing anything new (which I wouldn't just do as an outside contributor).

Maybe this could be a plugin that users who run on _volatile_ (for the lack of a better word) environments can load. But then I don't really see a use-case for it anyway.

SysPete:
@racke but would you use file-based sessions in production? We generally recommend a non-file approach so this is possibly mostly hitting very small sites and some dev setups. If you have strong objections then I think we should close the original issue as "won't fix".

racke:
@SysPete I usually don't, but still *way more* reasonable than removing the session directory. So my recommendation would be to close with wontfix.

cromedome:
Per discussion, closing/won't fix.

PR #1473, opened 2018-10-04
Show support resources after generating new app

cromedome:
For new users, it may be handy to have a little reminder of the
resources they have available to them when help is needed.

This was - somehow - inspired by the default MOTD you get after
installing FreeBSD.


racke:
Yeah, good idea. :+1:

cromedome:
Merged!

PR #1474, opened 2018-10-04
Use the correct URL for HAProxy.

cromedome:
Self-explanatory :)


racke:
I think we should nowadays promote the https:// URL if available.

cromedome:
Fixed, thanks. Just copied, pasted, and didn't look at the URL. Whoops.

racke:
:+1:

PR #1475, opened 2018-10-04
Add a section to manual regarding security concerns

cromedome:
This has been absent from the manual for far too long,

### DIFF
--- a/lib/Dancer2.pm
+++ b/lib/Dancer2.pm
@@ -133,7 +133,7 @@ Dancer2 is easy and fun:
 
     use Dancer2;
     get '/' => sub { "Hello World" };
-    dance;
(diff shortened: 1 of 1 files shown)


lib/Dancer2.pm
```diff
@@ -133,7 +133,7 @@ Dancer2 is easy and fun:
 
     use Dancer2;
     get '/' => sub { "Hello World" };
-    dance;
+    dance; Show support resources after generating new app Show support resources after generating new app
```
SysPete: erm?

cromedome: Interesting, not sure what happened there. Thanks for pointing it out!

cromedome: Fixed, thanks.

Issue #1127, opened 2016-02-05
route parameters with type constraints


SysPete:
Would be nice to have type constraints for route parameters as suggested by @shadowcat-mst 

```
<mst> get '/user/:id(Int)/'
<mst> or similar
<mst> which would default to Types::Standard
<SysPete> absolutely
<mst> but then you can create a MyApp::Types to declare your own in
<mst> using Type::Library + inheritance
<mst> in Catalyst we tried doing it via getting people to import the type constants, but that gets disappointingly crappy fast
<mst> whereas
<mst> set type_library => 'MyApp::Types';
<mst> is relatively nice
<SysPete> yup
<mst> note: this design may also have problems. but my last attempt definitely did so :)
```

shumphrey:
This would be neat.

SysPete:
@shumphrey I have a branch which works for simple type declarations for types available in `Dancer::Core::Types` (from my `Type::Tiny` branch so includes `Types::Standard`). Only works for simple types atm but could do more. Do you have any expectations?

bigpresh:
Definitely sounds like a cool feature.  @SysPete, you saying your branch already provides simple support for this?  Is there a link look at it for the lazy?

SysPete:
@bigpresh this is the commit with my initial implementation: https://github.com/SysPete/Dancer2/commit/ccdffaddaa30fb3b2ccc553cfc3561731d1ae370. No pod updates and no tests pushed yet but that will happen in the next day or so. Changes are really small.

SysPete:
@bigpresh this feature depends on my [Type::Tiny PR](https://github.com/PerlDancer/Dancer2/pull/1120) (hint, hint) and is not feature-complete: needs to support user-specified type library for starters.

shumphrey:
@SysPete No expectations. I like the idea of the route matching happening on arbitrary types. Currently, if we do a regex match, and the route doesn't match, it goes on to the next one. This allows two routes, one for a string, one for an int.
This removes the need for me to write if/else logic etc as this is handled by D2.
Extending this with types, would further allow routes to be distinguished without the need for conditional logic.
In general, I think the regex routes cover the majority of cases, where route definitions conflict, but adding types would just make this much more flexible.

SysPete:
@bigpresh, @shumphrey  something close to what will appear in a PR can be seen here: https://github.com/SysPete/Dancer2/compare/feature/type_tiny...SysPete:feature/gh1127_route_params_with_types

[...]

PR #1476, opened 2018-10-07
Route params with types

SysPete:
Hopefully this PR will generate some discussion as to how we would like to proceed with typed params. This code his been around for > 2 years and PR created after I was pinged about it by @cromedome.

Allows types to be specified in route param definitions allowing such things as:

```
get '/thing/:id[Int]' => sub {};

# if not Int then perhaps a Num...
get '/thing/:id[Num]' => sub {};
```

In original issue @yanick suggested:
> Interesting venue to explore. We could even do something like (off-the-cuff pseudo-code below)

```
subtype UserId => as Int => validate { };  # don't remember the real Types::Tiny syntax

get '/api/user/!UserId' => sub { 
  ...;
  # dancer magic would use the type name to populate the right param name
  param('UserId'); 
};
```

This is not currently possible in current implementation but could be added later.

See also @xsawyerx's [Dancer2::Plugin::ParamTypes](https://metacpan.org/pod/Dancer2::Plugin::ParamTypes).

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1222,8 +1222,11 @@ sub add_route {
     my $self        = shift;
     my %route_attrs = @_;
 
-    my $route =
-      Dancer2::Core::Route->new( %route_attrs, prefix => $self->prefix );
+    my $route = Dancer2::Core::Route->new(
+        type_library => $self->config->{type_library},
+        %route_attrs,
+        prefix => $self->prefix,
+    );
 
     my $method = $route->method;
 
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -3,10 +3,12 @@ package Dancer2::Core::Route;
 
 use Moo;
 use Dancer2::Core::Types;
+use Module::Runtime 'use_module';
 use Carp 'croak';
 use List::Util 'first';
 use Scalar::Util 'blessed';
 use Ref::Util qw< is_regexpref >;
+use Type::Registry;
 
 our ( $REQUEST, $RESPONSE, $RESPONDER, $WRITER, $ERROR_HANDLER );
 
@@ -75,6 +77,12 @@ has _params => (
     default => sub { [] },
 );
 
+has _typed_params => (
+    is      => 'ro',
+    isa     => ArrayRef,
+    default => sub { [] },
+);
+
 sub match {
     my ( $self, $request ) = @_;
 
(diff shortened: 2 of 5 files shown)


cromedome:
Disclaimer: I have not yet had a chance to experiment with this, my feedback is based solely on looking through the PR without running tests or writing code.

I *really* like how easy this PR makes it for the developer to add types to their parameters in Dancer2. Being able to have Dancer2 validate the parameters as they are passed into the application will help to eliminate a lot of boilerplate code that does just that (at least that's the case from my own code). Pretty excited to clone this and try to adapt a couple of my existing apps to it. This goes a long way towards making things easy and fun again, as we like to preach.

It's my hope that with an eventual inclusion of this that @yanick's suggestion won't be too far behind ;)

I'll have more comments once I get a chance to work with this throughout the week. Thanks @SysPete!

cromedome:
For the most part, this has been awesome. A couple of weird things that I have run into however:

I have a route that ends with an optional parameter (example based on something at `$work`):
```
get '/pdf/:product[StrMatch[qr{(foo|bar|baz)}]]/:state[StrMatch[qr{(\w{2})}]]/:force[StrMatch[qr{([fxplt])}]]?' => sub {
    return "Hello! " . route_parameters->{ product } . " / " . 
        route_parameters->{ state } . 
        ' / force ' . route_parameters->{ force };
};
```
In this example, the URL `http://localhost:5000/pdf/bar/FL/f` matches, but `http://localhost:5000/pdf/bar/FL/` and `http://localhost:5000/pdf/bar/FL` does not (to be fair, I expected the latter would not).

Originally, the last parameter was optional, and this was handled via the regex:
```
get '/pdf/:product[StrMatch[qr{(foo|bar|baz)}]]/:state[StrMatch[qr{(\w{2})}]]/:force[StrMatch[qr{([fxplt]?)}]]' => sub {
    return "Hello! " . route_parameters->{ product } . " / " . 
        route_parameters->{ state } . 
        ' / force ' . route_parameters->{ force };
};
```
This, however, doesn't compile: `Error while loading ~/TestApp/bin/app.psgi: Expected R_BRACKET; got MYSTERY at ~/perl5/perlbrew/perls/dancer-5.28.0/lib/site_perl/5.28.0/Type/Parser.pm line 391.'`

I intend to investigate further. Posting this here for the record, and in case you get to it before I can.

Thanks! Really liking this so far.

SysPete:
`strict_typed_params` turned out to be a bad idea so I've removed it. I have now squashed all commits to get rid of a pile of noise and consider this PR ready for review.

This comment https://github.com/perldancer/dancer2/issues/1127#issuecomment-202900429 from @yanick is not yet addressed and I will look at adding that as a later enhancement.

cromedome:
I just spent some time reviewing this again, and I am really happy with it. My earlier issue is resolved, and I haven't run into any other bugs or unexpected behavior. This gets a big :+1: from me! (sorry it took so long to get back to it).

Fellow Dancers - What say ye? :smile:

yanick:
Let's.

cromedome:
It's go time! Thanks @SysPete!

PR #1479, opened 2018-10-27
remove arbitrary 5.10 requirements from tests

Grinnz:
Dancer2 works on 5.8.8+ (earlier perl versions cannot install Template Toolkit due to the AppConfig requirement). These tests should not arbitrarily require 5.10.


veryrusty:
:+1: Nice pickup @Grinnz .

cromedome:
Merged, thanks!

PR #1480, opened 2018-10-27
Correct dynamic HTTP::XSCookies requirement

Grinnz:
Dynamic prereqs cannot be specified in `cpanfile`. The cpanfile is not used on the installing system, the prereqs are only read from the generated META.json and the MYMETA.json that gets generated by ExtUtils::MakeMaker on the installing system. The current code will only be run by whoever is running `dzil release` and ends up inserting a hard HTTP::XSCookies requirement in the resulting distribution. So to specify this prereq dynamically for the installer we need to use [DynamicPrereqs].


cromedome:
Merged at 21afe35dd1b23230238cf8422191d98fc6795fdb. Thanks again!

Issue #1483, opened 2018-11-07
Dancer2 didn't install on Windows 32 for me, issue with Plack dependency


sdondley:
This is more of a heads up to Dancer2 distro maintainers in case it isn't on the radar. I was doing some testing on an old Win 8 virtual machine. Ran into issues getting Plack installed. Issue was with a Plack dependency, [Test::Time](https://metacpan.org/pod/Test::Time).

See https://github.com/cho45/Test-Time/issues

From what I can gather, bug was introduced into the module within the last year. Not sure what the best approach is for making sure new Dancer installs don't encounter this problem.

cromedome:
Closing as the original `Test::Time` issue has been resolved.

PR #1484, opened 2018-11-11
add Route attribute for storing extra data

sdondley:
This is a simple patch to `Dancer2::Core::Route` I'd like to have considered. It would be useful to me in the development of a plugin for attaching additional information associated with routes. In my case, I'd like to associate routes with a title for a menu item.

This patch proposes adding a new attribute, `xt_data`, for storing a hash reference for containing this kind of "extra data".

### DIFF
--- a/lib/Dancer2/Core/Route.pm
+++ b/lib/Dancer2/Core/Route.pm
@@ -42,6 +42,12 @@ has options => (
     predicate => 1,
 );
 
+has xt_data => (
+    is => 'ro',
+    isa => HashRef,
+    predicate => 1,
+);
+
 sub _check_options {
     my ( $self, $options ) = @_;
     return 1 unless defined $options;


sdondley:
Look like it failed due to this issue: https://github.com/PerlDancer/Dancer2/issues/1483

sdondley:
It turns out I don't need this change for my module. Will leave for consideration, though.

cromedome:
Can you provide an example of how you planned to use this?

If we'd move forward with this, will also need docs and tests to support this change.

Thanks!

cromedome:
@sdondley One more ping before I close this out. Can you provide some guidance on how you planned to use this so we can discuss this?

sdondley:
Any neurons that once fired on this topic are long dead.

cromedome:
@sdondley thanks for getting back with me. I'm going to close this for now. If you remember what it was and want to discuss it, please feel free to add to this issue.

PR #1485, opened 2018-11-11
change module requirement to allow Plack to be installed on Windows

sdondley:
Attempt at a fix for https://github.com/PerlDancer/Dancer2/issues/1483. After applying this patch locally, I was able to get Dancer2 installed on my Windows machine. (Note that there was a failing cookie test, however. I just commented that out but that commented out code is not in this patch).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,6 +22,7 @@ install:
   - cpanm Dist::Zilla::PluginBundle::Git
   - cpanm Dist::Zilla::PluginBundle::Dancer
   - cpanm Dist::Zilla::Plugin::Prereqs::FromCPANfile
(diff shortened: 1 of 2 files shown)


sdondley:
Just another note here, Plack will only fail to install with Test::Time if the `Cookie::Baker` module is not already installed.

cromedome:
This is a toolchain issue, and not something I think we should solve in Dancer, so this gets :-1: from me. If you haven't already done so, it might be best to submit an issue to https://github.com/plack/Plack/issues

Perhaps you could add this information to the wiki so others affected by this might be able to find a solution to it?

sdondley:
K, but not patching .appveyor will cause all those tests to fail.

cromedome:
I am pondering the wisdom of cherry-picking that bit out. I am not a big fan of `--notest`.

cromedome:
Closing as the original `Test::Time` issue has been resolved.

PR #1493, opened 2018-12-03
Fix body not being send on forward

jpiautinity:
According to my use cases forward does not forward the body which seems to happen because `_shallow_clone` assumes that the body is fully read and cached. This assumption appears to be wrong though.  Thus `delete $env->{CONTENT_LENGTH};` prevents the body from being accessible at all in the forwarded route.

Furthermore later in `_shallow_clone` the new requests hash gets `body ` set to the old body. To my knowledge this hash key is never accessed anywhere. Neither in Plack nor in Dancer2. I removed it in order to reduce dead and potentially misleading code.

I am using Starman as web server and a pretty trivial test case.
Note that it does not matter whether the body is used within the old route or not; neither does explicitly setting the method to post.
```Perl
prefix '/old_prefix';
post '/route/:variable' => sub {
    my $variable        = route_parameters->get('variable');
    forward "/$NEW_PREFIX/route/$variable";
};
```

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -566,11 +566,6 @@ sub _shallow_clone {
     # in $self, then merge any overridden values
     my $env = { %{ $self->env }, %{ $options || {} } };
 
-    # request body fh has been read till end
-    # delete CONTENT_LENGTH in new request (no need to parse body again)
-    # and merge existing params
-    delete $env->{CONTENT_LENGTH};
-
     my $new_request = __PACKAGE__->new(
         env         => $env,
         body_params => {},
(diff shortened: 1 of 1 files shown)


SysPete:
`delete $env->{CONTENT_LENGTH}` was added via this PR to fix crashes with forward when using JSON serializer: https://github.com/PerlDancer/Dancer2/pull/1040

Since the test added in the PR passes with this new change I'm leaning towards :+1: but would like @veryrusty's opinion since he reviewed that original PR and also the created dbaec0754 which moved this delete to the newly created `_shallow_clone` method.

veryrusty:
At the time of dbaec07, the request parsing used HTTP::Body. Commit dbaec07 was to stop the body being parsed again by removing the content length, causing the loop that read the body to not execute.

14 months later we replaced the use of HTTP::Body with HTTP::Entity::Parser (as Plack::Request had done). eg commit 65854ec. After that, Dancer2::Core::Request (via Plack::Request) uses other mechanisms to stop a request body being re-parsed. The workaround of deleting the content length on forward isn't necessary any longer.

This gets a 👍 from me. Thanks @jpiautinity for the PR, and thanks @SysPete for pinging me.

SysPete:
:+1: from me. Thank you @veryrusty for the explanation.

cromedome:
Merged. Thank you!

Issue #1492, opened 2018-11-26
Dancer2::Config documentation makes reference to non-existent module


sdondley:
https://metacpan.org/pod/distribution/Dancer2/lib/Dancer2/Config.pod#Logging,-debugging-and-error-handling mentions a `Dancer2::Config::Object` module but there is no such animal.

racke:
AFAICT, this is a relict from Dancer 1. Couldn't find strict_config other than in this POD.

racke:
I don't see that happening in the near future ... so I suggest to remove the invalid reference. It can be restored easily if needed.

PR #1494, opened 2019-01-18
Remove references to strict_config, Dancer2::Config::Object

haguenau:
This is meant to fix PerlDancer/Dancer2#1492. Note that that issue
only mentions `Dancer2::Config::Object` being a non-existent module,
but since the `strict_config` property does not seem to exist at all
either, I just dropped the whole documentation subsection.


SysPete:
:+1: This needs to be removed. Many thanks @haguenau

racke:
:+1:

PR #1495, opened 2019-02-06
Proposed fix for FAIL report where cookie value is missing.

manwar:
Hi Team Dancer2,

Please review the PR.

https://www.cpantesters.org/cpan/report/d3d83d54-e862-11e8-b9e4-d3d34df7d3e2

       Use of uninitialized value $value in split at /home/cpansand/.cpan/build/2018111422/Dancer2- 
       0.207000-1/blib/lib/Dancer2/Core/Request.pm line 552.
       Modification of non-creatable hash value attempted, subscript "no.value.cookie" at 
       /home/cpansand/.cpan/build/2018111422/Dancer2-0.207000-1/blib/lib/Dancer2/Core/Request.pm 
       line 552.
       # Tests were run but no plan was declared and done_testing() was not seen.
       # Looks like your test exited with 255 just after 26.
       t/request.t ............................................ 

Many Thanks.
Best Regards,
Mohammad S Anwar

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -549,11 +549,14 @@ sub _build_cookies {
 
     # convert to objects
     while (my ($name, $value) = each %{$cookies}) {
+        my $cookie_value;
+        if (defined $value) {
+            # HTTP::XSCookies v0.17+ will do the split and return an arrayref
+            $cookie_value = (is_arrayref($value) ? $value : [split(/[&;]/, $value)]);
+        }
         $cookies->{$name} = Dancer2::Core::Cookie->new(
             name  => $name,
-            # HTTP::XSCookies v0.17+ will do the split and return an arrayref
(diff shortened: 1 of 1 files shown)


xsawyerx:
This seems more cosmetic, as the problem is beyond the use of uninitialized value in a split, but rather an underlying problem of having no value for a cookie.

The questions instead should be, why do we get into this situation and what do we want to do when it happens?

xsawyerx:
This breakage could be HTTP::XSCookies 0.000016 which changed some stuff relating to this:

https://metacpan.org/changes/distribution/HTTP-XSCookies#L33

SysPete:
@xsawyerx how about we just bump recommends on `HTTP::XSCookies`?

cromedome:
What @SysPete said.

cromedome:
Seems to be resolved. Will reopen if I'm wrong. Thanks @manwar !

PR #1498, opened 2019-03-19
load Encode so we can use Loggers standalone without bang

simbabque:
When working on my Dancer2::Logger::Console::Colored module I spotted that Encode is missing in the Logger role. Talked to Pete on IRC too.


veryrusty:
:+1: Thanks @simbabque

cromedome:
Merged, thank you!

PR #1500, opened 2019-03-29
Dancer2::Compat: Add compatibility wrapper module

clayne:
* MIME::Base64::encode_base64url: This is only available in v3.11+
  of MIME::Base64 and as far as I can tell Perl 5.14+ (where it is
  a core module).

* List::Util::pairmap/pairgrep: Both of these are only available
  in 1.29+ and as a core module didn't show up until Perl 5.20+.

* The general approach is to use a separate module (this one) with
  it's own import routine which will attempt to use the original
  functions if available, and if not, export a wrapper version
  providing the equivalent functionality.

Background:

  This is an attempt to defend against a few module functions used
  by Dancer2 that are only available if running more recent versions
  of Perl and/or List::Util. In our case, we have *many* RHEL6 hosts
  which are still locked to Perl 5.10.x as the major version and in
  the case of some modules (notably both involved in this change) are
  packaged with perl itself - making it slightly more difficult to
  simply build and deploy newer packages for them. Up until now
  we've been making locally patched builds to work around the
  issues but in the spirit of opensource and wider compatibility
  for everyone, I figured a formal patch would be better.

### DIFF
--- a/lib/Dancer2/Compat.pm
+++ b/lib/Dancer2/Compat.pm
@@ -1,0 +1,112 @@
+package Dancer2::Compat;
+# ABSTRACT: Compatibilty wrappers for newer module functions
+
+use strict;
+use warnings;
+
+our @ISA = qw(Exporter);
+our @EXPORT_OK = qw(
+    encode_base64url
+    pairgrep
+    pairmap
+);
+
+sub import {
+    my $self = shift;
+
+    # This function is basically a dispatch table that executes
+    # an anonymous sub for each matching import requested. Within
+    # each sub an attempt is made to require and import the
+    # wrapped function - and if the import fails (because the
+    # function doens't exist), a wrapper version will be shimmed
+    # into the symbol table and exported as normal. The only change
+    # required on the caller's end is to use this module rather than
+    # the original module being wrapped.
+
+    {
+        # MIME::Base64::encode_base64url (3.11+)
+        encode_base64url => sub {
+            require MIME::Base64;
+            eval  { MIME::Base64->import('encode_base64url') };
+            return unless ($@);
+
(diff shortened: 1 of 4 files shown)


SysPete:
I've looked at this PR several times over the last 9 months and I think my final conclusion is that it should be rejected. Reasoning is as follows:

- List::Util v1.29 and MIME::Base64 v3.11 are both available from CPAN and are installable for the target Perl stack. 
- Merging this change assumes the requester will then install a new Dancer2 from CPAN. Why not then also install other dependencies from CPAN?
- Merging this change adds an additional maintenance burden on the already over-stretched Dancer2 core team, a burden that appears, to me at least, greater than having Dancer2 users install other dependencies from CPAN.

@clayne I really appreciate your work here (and elsewhere - other PRs and issues from you will result in what I expect to be more positive results), but I also hope you can understand why I prefer not to have this code merged.

I can still be overruled by other @PerlDancer members and would appreciate their views.

racke:
I agree with @SysPete

veryrusty:
If you must use "system perl" but can't update dual-life modules; there are other options, like local::lib.

I agree with @SysPete

cromedome:
I agree with @SysPete. I know what you are trying to do, but we cannot absorb the burden of this, not when there are too many alternatives out there.

That being said, I really, really appreciate the time, effort, and work you put into this, and I appreciate your enthusiasm for this project. Please don't let this deter you from any further contributions!

Dancer team: I move to close this. Agreed?

clayne:
Yep, I think everyone's rationale makes a lot of sense here and we can always use some local patches/workarounds on our end. I'll close it out.

Issue #1320, opened 2017-02-12
Real setup method for apps


xsawyerx:
One common annoyance I find more and more is that applications cannot declare a setup method to be run prior to creating the app.

You have two options:
1. Run code as soon as the class is loaded: This is problematic because it won't be tied to a caller cleanly and will not receive parameters.
2. Manually have a method called (I wrote about it [here](http://advent.perldancer.org/2014/18)) and hope no one forgot to call it before calling `to_app`.

I suggest creating a proper one, called `prepare_app`, based on the equivalent one in `Plack`.

My suggestion is to first call warning for any such found methods in apps, to make sure that the subroutine is not found or the users know about it. Then after a number of releases or enough time, to call such a method by default, having given any users that have `prepare_app` in their code a fair warning we intend to do it.

What do you think, @PerlDancer?

veryrusty:
#1331 has been in place for 2 years now. Think it is time for `prepare_app` to be a real thing.

I would like to see this in the next release. If anyone wants to have a go, the floor is yours :dancers:

xsawyerx:
Do we have a pending release?

cromedome:
I don't presently have one scheduled, but it would be nice to put one out before TPC::NA.

I just bought a house and am moving this weekend. After I get settled a bit, I will get the ball rolling on one.

Speaking of TPC::NA, any interest in putting together a Dancer BOF, or at the very least, a Dancer Happy Hour? :)

xsawyerx:
I started implementing this and realized there is a problem in the original design of a `prepare_app` subroutine. When you have an App that is based on another app, we need to know which subroutine to call (the parent app or the subclassing app). I think we should run both.

Example:

```perl
package Foo {
  use Dancer2;
  sub prepare_app {...}
}

package Bar {
  use Dancer2 'appname' => 'Foo';
  sub prepare_app {...}
}
```

In either case, the inheritance could be an arbitrary amount of levels, which would require us to somehow either inspect the calling  (which I'm not sure we have here) or we have to start tracking inheritance when people set `appname`.

An alternative is to register a callback:

```perl
package Foo {
  use Dancer2;
  prepare_app {...}; # notice the semicolon at the end
}

package Bar {
  use Dancer2 'appname' => 'Foo';
  prepare_app {...}; # notice the semicolon at the end
}
```

That way we just stack them up and call all of them in the order of registration.

Thoughts?

cromedome:
[...]

PR #1504, opened 2019-06-18
Implement prepare_app keyword

xsawyerx:
The new `prepare_app` keyword allows the user to have a setup method for their code.

If you have an app based on another app, both `prepare_app`s will be run, which I think would be correct behavior.

GH #1320

### DIFF
--- a/lib/Dancer2/Manual.pod
+++ b/lib/Dancer2/Manual.pod
@@ -2528,6 +2528,26 @@ name.
         # will result in two headers "x-my-header" in the response
     }
 
+=head2 prepare_app
+
+You can introduce code you want to run when your app is loaded, similar to the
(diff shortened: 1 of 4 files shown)


cromedome:
:+1: Awesome work.

lib/Dancer2/Manual.pod
```diff
@@ -2528,6 +2528,26 @@ name.
...
+C<prepare_app> in L<Plack::Middleware>.
+
+    prepare_app {
+        my $app = shift;
+
+        ... # do your thing
+    };
+
+You should not close over the App instance, since you receive it as a first
+argument. If you close over it, you B<will> have a memory leak.
```
SysPete: :+1:

cromedome:
Merged, thank you!

PR #1506, opened 2019-07-02
Explain what add_route() does with its arguments

tomhukins:


racke:
Now I'm wondering whether this method will return the new Route object :nerd_face:

tomhukins:
> Now I'm wondering whether this method will return the new Route object nerd_face

@racke, I approve of your "_yes, and.._" style so I've also submitted #1520 :grinning:

Issue #1507, opened 2019-07-04
Invalid body causes a 500


xsawyerx:
When making a call with an invalid body, `HTTP::Body` croaks, which triggers a 400.

I think we should instead return a 400 for an invalid request.

Objections?

cromedome:
No objection! :+1:

ice-lenor:
The issue is also present in Dancer v1
https://github.com/PerlDancer/Dancer/issues/1167

xsawyerx:
Reference tests for this: #1508

cromedome:
Resolved by #1509.

Issue #1496, opened 2019-02-14
Invalid GET request causes 500


JRaspass:
```
$ plackup -e 'use Dancer2; dance'
HTTP::Server::PSGI: Accepting connections at http://0:5000/
127.0.0.1 - - [14/Feb/2019:14:29:25 +0000] "GET / HTTP/1.1" 500 180 "-" "curl/7.64.0"
```
```
$ curl -so /dev/null -w "%{http_code}\n" -H 'Content-Type: multipart/form-data; boundary=a' -k http://localhost:5000
500
```

It seems if you pass a GET request which obviously has no body, but claim to have multipart/form-data in your body then Dancer2 dies trying to read said non-existent body.

This is similar in spirit to #585 and I think we should either skip processing the body when we don't have one, or return a 400.

SysPete:
400 response seems sane in this case.

cromedome:
This is resolved by #1509. Closing!

PR #1509, opened 2019-07-05
Request instantiation fails throw 400 Bad Response

veryrusty:
This started out as a patch for 1056; permitting deserialisation fails to use before hooks to alter the error thrown.

Since then, issues like #1496 and #1507 highlight that we want something more general in this regard. This PR changes behaviour, as what was previously a 500 Internal Server error when request object creation failed is now a 400 Bad Response error in line with the above issues.

Putting this out here for feedback and seeing if this is how we'd like to solve this.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1441,15 +1441,12 @@ sub dispatch {
     # Catch bad content causing deserialization to fail when building the request
     if ( ! $request_built_successfully ) {
         my $err = $@;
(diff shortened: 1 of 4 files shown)


SysPete:
:+1: this is a clear improvement over current 500 response. I can also live with the status and message attributes switching to `rw` as this allows flexibility when needed.

cromedome:
:+1: Sorry for not approving this sooner - I was ready to but had to think through any consequences to changing the status of those attributes, then promptly dropped the ball. This is a solid fix. Thank you!

cromedome:
Merged, thank you!

PR #1510, opened 2019-07-08
Multipart returning 400

xsawyerx:
This is just a test + cleanups at first.

I'm going to now add the fix for it.

### DIFF
--- a/t/multipart_content.t
+++ b/t/multipart_content.t
@@ -27,7 +27,7 @@ test_psgi $app, sub {
     my $cb = shift;
 
     is(
-        $cb->( GET '/' => ['Content-Type' => 'multipart/form-data'] )->code,
+        $cb->( GET '/' => 'Content-Type' => 'multipart/form-data' )->code,
         400,
         'multipart with incorrect boundary returns 400',
     );
@@ -37,10 +37,9 @@ test_psgi $app, sub {
 
     is(
         $cb->(
-            GET '/' => [
+            GET '/' =>
                 'Content-Type' =>
-                  'Content-Type: multipart/form-data boundary=------boundary-------'
-            ],
+                'Content-Type: multipart/form-data boundary=------boundary-------',
         )->code,
         200,
         'Providing multipart with correct boundary works',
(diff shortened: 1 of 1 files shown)


t/multipart_content.t
```diff
@@ -15,39 +20,42 @@ use Dancer2;
...
-        my $res = $cb->(HTTP::Request->new($http, '/', $headers));
-        is($res->code, 200);
-    }
-    { # passes
-        my $headers = [ 'Content-Type' => 'text/html; charset=UTF-8' ];
-        my $res = $cb->(HTTP::Request->new($http, '/', $headers));
-        is($res->code, 200);
-    }
+    is(
+        $cb->( GET '/' => ['Content-Type' => 'multipart/form-data'] )->code,
```
veryrusty: The syntax for the extra headers isn't correct.  Should be:
```
$cb->( GET '/', 'Content-Type' => 'multipart/form-data' )->code,
```

veryrusty:
Also look at #1509 (These tests with correct header syntax pass on that branch).

cromedome:
:+1: for merging this test. Thanks @xsawyerx.

veryrusty:
Trying to sort out the test failure for the April 2020 release.

If I grok this correctly; the 3rd test (https://github.com/PerlDancer/Dancer2/pull/1510/files#diff-1044d466c97c916209b2eab15abab50aR42) has a header that isn't what the test expects. The PSGI env gets
```
CONTENT_TYPE => 'Content-Type: multipart/form-data boundary=------boundary-------'
```
When the request is parsed; HTTP::Entity::Parser matches that against the registered types. None of those begin with `Content-Type`, so the request gets parsed as an OctetStream, and no error occurs.  ( Those details are here: https://github.com/kazeburo/HTTP-Entity-Parser/blob/master/lib/HTTP/Entity/Parser.pm#L53-L63 ).

The 4th test has a content type header that matches a registered parser. That it handed off to HTTP::MultiPartParser, which throws an error in parsing the request. With #1509 merged, the response is now a 400 (which I believe is what #1507 wanted). The error returned (generated by HTTP::MultiPartParse) is `End of stream encountered while parsing preamble`. 

To clean this up:
Delete the current 3rd test; it isn't doing what we expect.  The current 4th test should check it gets a 400 response.

cromedome:
Resolved by #1509.

PR #1516, opened 2019-08-05
Add Types::Standard to cpanfile. 

### LINKED ISSUE #1515: Dependency on Types::Standard not declared
Dancer2::Core::Types extends `Types::Standard`, but we do not list Type::Tiny or Types::Standard as a dependency.   I suspect an upstream dependency changed, exposing this.

You can verify with a clean perl install or use Docker to makes it easy: `docker run --rm -ti perl:5.14 cpanm -v Dancer2`.

veryrusty:
Add Types::Standard to cpanfile, which is a dependency for Dancer2::Core::Types.

Closes #1515

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -40,6 +40,7 @@ requires 'Template';
 requires 'Template::Tiny';
 requires 'Test::Builder';
 requires 'Test::More';
+requires 'Types::Standard';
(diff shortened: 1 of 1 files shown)


cromedome:
👍 Thanks! Merged.

PR #1517, opened 2019-08-20
Add test and update doc for pluggable route handlers

1nickt:
Corrects bad doc on how to configure a pluggable route handler. Adds a test for a pluggable route handler, but wraps the tests in `TODO` since pluggable route handlers currently don't get registered when you run the app with `to_app` (if you start with `dance` in a script, they are registered and work OK).

Issue: (https://github.com/PerlDancer/Dancer2/issues/1518)


cromedome:
We are going to close the PR and look at the original ticket some more. Let's not let correct documentation paper over a bug we think should be fixed :)

PR #1519, opened 2019-08-22
Fix incorrect documentation

tomhukins:


SysPete:
Good catch! :+1: Thanks @tomhukins

tomhukins:
> Merged, thanks!

You're welcome!  Would you mind reviewing the older pull request I have open at #1506?  It's just as simple and would have helped me understand Dancer2 more easily.

PR #1524, opened 2019-11-13
Fix nonsensical documentation

tomhukins:
It's still hard to read, but better than it was..


cromedome:
:+1: Thank you!

SysPete:
:+1: Thank you @tomhukins!

cromedome:
Merged, thanks again!

tomhukins:
I apologise if I'm missing something, but I don't see this change on the master branch.  I still see the old wording in `lib/Dancer2/Config.pod` and on https://metacpan.org/pod/Dancer2::Config so I suspect you forgot to merge this change.

cromedome:
I'm not sure what happened but I think this is in master now. Sorry about that.

PR #1525, opened 2019-11-15
remove use of Return::MultiLevel, and implement stack frame jumping manually

haarg:
Return::MultiLevel obscures the fact that returning across multiple
stack frames is very evil.  Rather than hide the evil we are doing, just
embed the stack breaking via a label and last.

This avoids a Return::MultiLevel dependency, which avoids pulling in
Scope::Upper, another module with too much magic in it.  It also is
slightly faster, although that is a very minor benefit.


cromedome:
Thanks haarg! I love seeing changes like this. :+1:

PR #1526, opened 2019-11-24
docs: Mention that TT2 config start_tag/end_tag need escaping

cxw42:
Sawyer, Jason, it was a pleasure to meet you earlier this year at TPCiP!  I am trying out Dancer2 and tripped over a minor documentation point.  This is my first PR, and attempts to help others avoid what confused me.

Context: I am trying to use TT2 with PHP-style tags (`<? ?>`).

Dancer2::Template::TemplateToolkit does not mention that the TT2 `start_tag` and `end_tag` are regexes, so need backslash escapes. This commit adds that mention, and also a direct link to Template::Manual::Config.

Noted (and tested) with Dancer2 v0.208001, Cygwin x64, running under plackup.

Thank you for considering this PR!

PR #1527, opened 2019-12-01
update travis dist to 'trusty'

bor:
travis moved to using 'xenial' dist by default a few months ago
and for 'xenial' there are no support of perl less than 5.22
  https://docs.travis-ci.com/user/reference/xenial#perl-support
for 'trusty' using perlbrew with any version of perl is still ok
  https://docs.travis-ci.com/user/reference/trusty#perl-images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: perl
 
 perl:


cxw42:
This is the cause of the build failures in #1526 as well.

racke:
Kind of sad that we need to use an old distribution. But that is a good find, I noticed a similar problem with several other Perl projects.

cromedome:
:+1: Thank you!

cromedome:
Merged, closing!

PR #1528, opened 2019-12-01
skel/config.yml: Note that "engines:" must be merged

cxw42:
Another thing that tripped me up, and that I would like to save others from.  If config.yml has:

    engines:
      template:
        ...
    engines:
      session:
        ...

(i.e., two `engines:` keys), only the session-engine configuration will be loaded.  The template-engine configuration will be silently discarded when the YAML file is loaded.  This commit updates `share/skel/config.yml` to tell readers that you instead need to use:

    engines:
      template:
        ...
      session:
        ...

This commit also adds a note about TT2 start_tag and end_tag being regexes, per #1526.

### DIFF
--- a/share/skel/config.yml
+++ b/share/skel/config.yml
@@ -2,6 +2,8 @@
 # env-related settings should go to environments/$env.yml
 # all the settings in this file will be loaded at Dancer's startup.
 
+# === Basic configuration ===
+
 # Your application's name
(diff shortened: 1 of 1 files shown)


cromedome:
Nice catch! I can see how that would be confusing. :+1:

cromedome:
Merged, thanks!

PR #1536, opened 2020-02-24
Fix storing objects in YAML sessions

tomhukins:
This worked fine, until YAML version 1.30 disabled loading blessed objects by default.  In many circumstances that makes sense.  Untrusted user input should not refer to Perl objects.  However, sessions are always created by the application, so it seems reasonable to trust their contents.  Preventing developers from storing objects in sessions is a regression and removes a useful feature.


SysPete:
Since this is a regression I have to :+1: this change.

cromedome:
Seconded. 👍 from me. Thank you @tomhukins!

cromedome:
Merged. Thanks again!

Issue #1538, opened 2020-03-20
Documentation fix (Named matching)


TimothyAlexisVass:
change
```
get '/hello/:name?' => sub {
    my $name = route_parameters->get('name') //= 'Whoever you are';
    "Hello there, $name";                      ^  ^
};                                             HERE
```

to
```
get '/hello/:name?' => sub {
    my $name = route_parameters->get('name') // 'whoever you are';
    "Hello there, $name";
};
```

cromedome:
Merged. Thanks!

PR #1539, opened 2020-03-20
Documentation fix (Named matching)

SysPete:
This is issue #1538 

Thank you @TimothyAlexisVass 

defined or assignment in docs should be defined without assignment


SysPete:
@cromedome another one for next release?

racke:
I suggest that we also fix the implicit return in this example.

cromedome:
Merged. Thanks!

Issue #1541, opened 2020-04-07
t/issues/gh-811.t fails on some systems (Dancer2-0.300001)


eserte:
On some of my smoker systems the test suite fails:
```
    #   Failed test 'Redirects to correct url'
    #   at t/issues/gh-811.t line 43.
    #          got: '/get'
    #     expected: 'http://localhost/get'
    # Looks like you failed 1 test of 3.

#   Failed test 'Creating a session'
#   at t/issues/gh-811.t line 46.

    #   Failed test 'Correct response'
    #   at t/issues/gh-811.t line 53.
    #          got: '{}'
    #     expected: '{"foo":"bar"}'
    # Looks like you failed 1 test of 2.

#   Failed test 'Retrieving a session'
#   at t/issues/gh-811.t line 54.
# Looks like you failed 2 tests of 2.
t/issues/gh-811.t ...................................... 
Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/2 subtests 
```

racke:
Can you please indicate the Perl versions installed on the failed smokers?

racke:
Reproducible with Perl 5.30.

racke:
Different behaviour was introduced by f80ec210cd238f280dcc233fda85d4e31cc94e14 (#1461).

SysPete:
Sadly only reproducible if `Dancer2::Session::Cookie` is installed, which does not happen on travis. I'm creating a PR.

cromedome:
Just merged the fix. Thanks for reporting, and thanks @SysPete for fixing.

PR #1542, opened 2020-04-07
fix broken test that is skipped under CI

SysPete:
This is issue #1541 

Fix test that was broken by switch to RFC 7231 (#1461).

My original PR pulled in an authordep on Dancer2::Session::Cookie, which of course caused a circular dependency on D2 itself, so that was removed.


SysPete:
I'm not convinced that this test belongs in D2 itself, and wonder what other @PerlDancer think. For now this gets rid of the test failures.

cromedome:
@SysPete Are you thinking we move the test to the `Dancer2::Session::Cookie` distribution?

In any case, :+1: for the PR, and another :+1: for moving the test :)

PR #1550, opened 2020-06-29
Add Types::Standard to cpanfile. Ref #1515

gerlacher:
test


SysPete:
Merged via commit 6b08b3619. Thanks @gerlacher

PR #1554, opened 2020-08-01
bump dep on Role::Tiny to 2.000008

SysPete:
Avoids issue with Role::Tiny 2.000007 and Types::Standard undefined warnings

See: https://github.com/PerlDancer/Dancer2/issues/1515#issuecomment-518305067

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -32,7 +32,7 @@ requires 'Plack::Middleware::FixMissingBodyInRedirect';
 requires 'Plack::Middleware::RemoveRedundantBody';
 requires 'POSIX';
 requires 'Ref::Util';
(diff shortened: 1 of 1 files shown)


veryrusty:
We could also exclude the specific broken point release for Role::Tiny via
```
requires 'Role::Tiny', '>=2.000000, !=2.000007';
```
Which would make anyone with the broken release upgrade,  but still permit versions between 2.000000 and 2.000006.

SysPete:
Thank you @veryrusty, I'd forgotten such magic was possible. Updated.

veryrusty:
:+1: Thanks for the update @SysPete.  Feel free to merge this to master

Issue #802, opened 2014-12-06
warnings flag?


xsawyerx:
Is the `warnings` flag even used?
(This is related to `DANCER_WARNINGS` environment variable.)

PR #1555, opened 2020-08-07
remove all leftovers after removing 'warnings' config setting

bor:
which was happened in GH #802 (0bc2da1)


cromedome:
:+1: Merged. Thanks!

PR #1557, opened 2020-08-28
Pr/remove workaround for fixed http xscookies bug

ccntrq:
Hey,

in #1428 I worked with @bigpresh on a workaround for a bug in `HTTP::XSCookies`.

The bug has since been fixed in version `0.000015` of `HTTP::XSCookies` by the original author. The issue has been reported in Issue https://github.com/gonzus/http-xscookies/issues/7.

This pull requests increases the minimum version on the recommended dependency to ` 0.000015`  and reverts the hack from #1428. The testcase from #1428 still passes with these changes.


cromedome:
Merged, thanks!

PR #1558, opened 2020-08-28
provide a testcase for #1289

### LINKED ISSUE #1289: request->uploads are discarded after a forward
In the first handler of a route, the `request->uploads` hash contains the expected upload objects.  Following a `forward`, the `request->body` still contains the uploaded content but the `request->uploads` hash is empty and `upload($file)` is undefined.

ccntrq:
Hey, 

this PR adds a testcase for the issue described in #1289 

It seems the problem has been remedied since the issue has been opened. The testcase passes with the latest release of Dancer2.

Can we merge this and close #1289 or is there anything else that should be done before?

### DIFF
--- a/t/issues/gh-1289.t
+++ b/t/issues/gh-1289.t
@@ -1,0 +1,40 @@
+use strict;
+use warnings;
+use Test::More tests => 1;
+use Plack::Test;
+use HTTP::Request::Common;
+
+{
+
+    package App;
+    use Dancer2;
+
+    post '/forwards' => sub {
+        forward('/still_contains_uploads');
+    };
+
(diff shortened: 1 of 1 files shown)


veryrusty:
Thanks @ccntrq!  I suspect the issue was "resolved" when we migrated the Request object to inherit on Plack::Request.

Test looks good. There's a `$url` variable that's defined but not used. Mind cleaning that up ?

ccntrq:
I don't mind. Actually thanks for the hint. I amended the commit accordingly.

cromedome:
Merged! Thanks!

PR #1561, opened 2020-10-04
Bugfix/1546 static handler mime types

### LINKED ISSUE #1546: Adding a Mime::Type for files who get served from public/
I've stumbled upon the problem that a `*.mjs` file in `public/` didn't have the correct mime type assigned.

Trying to add it via `mime->add_type( mjs => 'text/javascript' );` didn't work, it seems that does not affects files who are served from `public/`.

I've worked around it by renaming it is `*.js` but it would be nice to be able to manipulate the mime types for files in `public/`. Or at least know how to do that, maybe I tried it the wrong way.

veryrusty:
* Change static content handler from `mime_type->from_name` to `mime_type->from_file` so the extension is used to determine the content type of the file.
* Minor optimisation to use `$_[0]` rather than shift.
* Tests for dsl `mime` keyword 

Resolves #1546.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1396,7 +1396,7 @@ sub to_app {
         # Use App::File to "serve" the static content
         my $static_app = Plack::App::File->new(
             root         => $self->config->{public_dir},
-            content_type => sub { $self->mime_type->for_name(shift) },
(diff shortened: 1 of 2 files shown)


cromedome:
Nice work! 👍

cromedome:
Merged, thanks!

yanick (commented):
Looks good!

PR #1562, opened 2020-10-17
Fix some stuffs in the Dancr2 Tutorial

cloveistaken:
Just a couple of small changes to some hyperlinks in Tutorial.pod. The old hyperlinks are currently not pointing correctly to the specific sections in metapcpan.


cloveistaken:
The final code in the tutorial also misses the snippet for database_error hook (introduced in the previous part). I have added it in the 2nd commit.

cromedome:
Merged, thank you!

PR #1563, opened 2020-10-27
made few spelling changes to perlcritic.rc file present in xt file

Sloth-Panda:


AchyutPanda:
Hey, just made few spelling corrections

veryrusty:
Thanks for correcting `popping`.

However `Readonly` (without a hyphen) refers to another perl module: https://metacpan.org/pod/Readonly

cromedome:
Reverted the `Read-only` change but kept the other. Thanks.

Issue #1564, opened 2020-11-04
redirect disregards path to app


nfg:
Hey!

The `redirect` function has stopped respecting the path leading to the Dancer2 app. If my PSGI looks something like

```perl
use strict;
use warnings;
use lib qw<lib>;

use Plack::Builder;

use MainApp;
use SubApp;

builder {
    mount '/' => MainApp->to_app();
    mount '/subapp' => SubApp->to_app();
};
```

Then `redirect "/page"` in MainApp works fine, but `redirect "/page"` in SubApp goes to `/page` instead of `/subapp/page`.

This changed in #1461 . I can fix this on my end by finding everywhere I call `redirect` and calling `uri_for` first, but it's unexpected behaviour.

Cheers!

cromedome:
Resolved by #1565. Thanks for reporting!

Issue #1416, opened 2017-12-14
Guidance on writing a plugin for managing included javascripts on a page


sdondley:
I'm thinking of writing a plugin to help me learn how to do write one. The first plugin I'm thinking of writing would allow developers to easily inject and remove `<script>` tags into a page to include javascript libraries. I'm thinking it would work something like this:

1) The plugin would detect if the library was external to the site. If it is, it would download the script to the local machine.

2) The plugin would write the necessary code to fall back to using the local copy of the code if external js is unavailable.

3) The developer could easily turn on and off libraries on specific pages. Some libraries could default to "always on" and would have to be removed by the developer while others would default to "always off" and have to be specifically added by the developer. 

4) The developer could indicate which javascript libraries could have dependencies. So if the developer includes jQuery Mobile, jQuery would automatically be loaded.

5) The developer could indicate where in the code it should go: head, body_top, body_bottom

5) The developer would configure these settings in the config file:

```
plugins:
  JavaScriptLoader:
    external_code: 'http://cdn.com/javascript'
    name: jQuery_mobile
    default: on
    depends_on: jQuery
    location: head
```

I think it would be straightforward, but I'm not sure of the best way to getting the inject the code into the page. I guess it would be up to the app developer to provide the appropriate variables in their template and then the plugin would have to somehow supply the html for those variables. But I'm not sure how that can best be achieved. Are there plugins I could look to as an example to see how this might be achieved?

Thanks.

StylusEaterCTP:
Interesting idea but as a developer, I would be concerned about the security implications. I don't think I would use this in a production setting unless, possibly, I were referencing internal repositories. Even then I'd want some sort of verification mechanism (maybe a checksum to go along with it? to make sure no one changed a file on where I'm pulling it from?). 

This might be of use [OWASP - XSS](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet)

sdondley:
It is common practice to load external libraries through `<script>` tags to inject them into the site (from reputable sources, of course). If the developer preferred to load the libraries locally off the server, it would be easy enough make that an option. So instead of using `extenal_code` they could load the js right from the server with something like `local_file: /js/jQuery.js`.

sdondley:
Some kind of checksum would be cool, though, I think. Would help give the developer some piece of mind. Not sure how much that might slow things down, though.

sdondley:
[...]

PR #1565, opened 2020-11-05
Add script_name to redirects beginning with /

nfg:
Hey! I'm not sure if this is the best solution, but at least the test script shows the problem.

I thought about copying some of the logic from uri_for, but you guys deliberately removed it in the first place. This keeps the developer responsible for the actual location, but will prefix the path if necessary.

EDIT: This is a potential fix for #1564

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1273,6 +1273,14 @@ sub redirect {
     my $destination = shift;
     my $status      = shift;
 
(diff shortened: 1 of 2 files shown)


SysPete (approved):
A good fix, thank you @nfg! Restores behaviour for in-app redirects to what it was before #1416 was merged, also providing behaviour that matches the docs.

nfg:
Cool! Thanks for taking a look, @SysPete !

cromedome:
:+1:  merged... thank you!

PR #1567, opened 2020-12-16
Fixed CSS so error message does not display outside of <pre> html element

elliotholden:
The error messages where displaying outside of the `<pre>` tag. Minor CSS fix.

### **Before**
before-css-fix

### **After**
after-css-fix


cromedome:
Thanks! :+1:

SysPete:
:+1:

cromedome:
Merged... thanks again!

PR #1575, opened 2020-12-26
Bugfix: mutable serializer & yaml dsl helpers loading depencendies.

### LINKED ISSUE #1568: Serializer::Mutable doesn't (auto) load other serializers.
Advent driven development: The mutable serializer has a couple of class loading bugs.  `YAML` isn't loaded, nor are any serializer classes in configured mapping. Serializers also eat errors, making chasing this down unfriendly.

The simple app below highlights the problem. Both these requests return no output and no errors :(`
 * `curl -H 'text/x-yaml' http://localhost:5000/` 
 * `curl -H 'text/xml' http://localhost:5000/`

```
#!/usr/bin/env perl
use warnings;
use strict;

{
    package Example::App;
    use Dancer2;
    # Serializer loading bugs
    # use YAML;
    # use Dancer2::Serializer::XML;

    set engines => {
        serializer => {
            Mutable => {
                mapping => {
                    'text/x-yaml'      => 'YAML',
                    'text/html'        => 'YAML',
                    'text/x-json'      => 'JSON',
                    'application/json' => 'JSON',
                    'text/xml'         => 'XML',
                }
            }
        }
    };
    set serializer => 'Mutable';

    get '/' => sub {
        return { some => 'data' };
    };
}

Example::App->psgi_app;
```

veryrusty:
🎄 ADD: Advent Driven Development 🎄

The easy part:
* `require` non-core serializers before calling serializer_class->new

The core tries hard to not load YAML.pm when it is not needed. This makes it possible to call the to/from yaml DSL methods without YAML loaded leading to silent serialization failures. Those helpers are also used by the mutable serializer.
* Use deferred subs to load YAML when called as class methods. 
* Tests for `to_yaml` and `from_yaml` dsl keywords, which also require YAML to be loaded.

Resolves #1568.

### DIFF
--- a/lib/Dancer2/Serializer/Mutable.pm
+++ b/lib/Dancer2/Serializer/Mutable.pm
@@ -4,6 +4,7 @@ package Dancer2::Serializer::Mutable;
 use Moo;
 use Carp 'croak';
 use Encode;
+use Module::Runtime 'require_module';
 with 'Dancer2::Core::Role::Serializer';
 
 use constant DEFAULT_CONTENT_TYPE => 'application/json';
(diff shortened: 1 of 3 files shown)


cromedome:
:+1: Nice work!

SysPete (approved):
:+1: Good to merge. Thanks @veryrusty!

cromedome:
Merged! Thanks!

PR #1581, opened 2021-01-29
disable scheduled GitHub action on forks

szabgab:


racke:
Can you please elaborate on this change, @szabgab ?

szabgab:
In the previous version of the Github Actions the job is executed on push events, on pull requests, and also using the scheduler. This worked both in the central repository and on every fork as well. This pull request disables the scheduled execution on the forks. Or at least I hope it does that.

racke:
OK, but what is wrong with running the action on your fork? Doesn't that need to enabled explicitly like for Travis?

szabgab:
AFAIK it is enabled by default.

racke:
Yes, that is true. Why is that a bad thing? With your change I can't run the actions on my fork if I want to.

szabgab:
This change, if I am not mistaken, only disables the scheduled job on the forks. It will still run all the jobs hen you push new code to your fork.

racke:
Yes, it still works on the fork when pushing commits there. :+1:

SysPete:
:+1:

cromedome:
Merged, thanks!

PR #1586, opened 2021-02-15
Run version checks against MetaCPAN.

cromedome:
When creating a new application, Dancer2 will check your current version against what is on CPAN, and if you aren't current, gives you a notification. This check was done against search.cpan.org which is no more. Rewritten to use the MetaCPAN API for the version check.

### DIFF
--- a/lib/Dancer2/CLI/Command/gen.pm
+++ b/lib/Dancer2/CLI/Command/gen.pm
@@ -266,7 +266,7 @@ sub _version_check {
     return if $version =~  m/_/;
 
     my $latest_version = 0;
-    my $resp = _send_http_request( 'https://fastapi.metacpan.org/module/Dancer2' );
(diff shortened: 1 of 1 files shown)


lib/Dancer2/CLI/Command/gen.pm
```diff
@@ -265,20 +266,20 @@ sub _version_check {
     return if $version =~  m/_/;
 
     my $latest_version = 0;
-    my $resp = _send_http_request('http://search.cpan.org/api/module/Dancer2');
+    my $resp = _send_http_request( 'https://fastapi.metacpan.org/module/Dancer2' );
```
veryrusty, line 269: I'd `s/module/release/` here  (though it really doesn't matter.. think the data returned is the same).

cromedome: Thanks for the tip! Changed.

yanick (approved):

cromedome:
Merged.

PR #1589, opened 2021-02-18
Generate a Dockerfile when creating a new app

cromedome:
Giving credit where credit is due: I shamelessly stole the sample Dockerfile from Mojolicious. To be fair, this has been on my list for about six months now, but them releasing the feature finally motivated me to do something about it.

You can now run `dancer2 gen` with a `-c` or `--docker` option to produce a minimalist Dockerfile for your application. The `gen` command will even give you some instructions for how to run it.

This could use a proper deployment guide, but that can be handled in one of the various documentation issues that exist.

PR #1590, opened 2021-02-18
Improve install/documentation of XS modules

cromedome:
This came up in discussing #1589 with @racke. This is really an independent fix from the Docker enhancement.

Make it easier to install the XS modules for the performance boost. Change the deployment pod to reflect this change. The Docker PR uses the new feature tag in `cpanfile` to install the XS modules when building the container image.

### DIFF
--- a/share/skel/cpanfile
+++ b/share/skel/cpanfile
@@ -1,9 +1,37 @@
 requires "Dancer2" => "[d2% dancer_version %2d]";
 
-recommends "YAML"             => "0";
-recommends "URL::Encode::XS"  => "0";
-recommends "CGI::Deurl::XS"   => "0";
-recommends "HTTP::Parser::XS" => "0";
+recommends "YAML"                    => "0";
+recommends "URL::Encode::XS"         => "0";
+recommends "CGI::Deurl::XS"          => "0";
+recommends "HTTP::Parser::XS"        => "0";
+recommends "CBOR::XS"                => "0";
+recommends "YAML::XS"                => "0";
+recommends "Class::XSAccessor"       => "0";
+recommends "Cpanel::JSON::XS"        => "0";
+recommends "Crypt::URandom"          => "0";
+recommends "HTTP::XSCookies"         => "0";
+recommends "HTTP::XSHeaders"         => "0";
+recommends "Math::Random::ISAAC::XS" => "0";
+recommends "MooX::TypeTiny"          => "0";
+recommends "Scope::Upper"            => "0";
+recommends "Type::Tiny::XS"          => "0";
+
+feature 'accelerate', 'Accelerate Dancer2 app performance with XS modules' => sub {
+    requires "URL::Encode::XS"         => "0";
+    requires "CGI::Deurl::XS"          => "0";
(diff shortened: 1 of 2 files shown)


share/skel/cpanfile
```diff
@@ -1,9 +1,37 @@
...
+recommends "CBOR::XS"                => "0";
+recommends "YAML::XS"                => "0";
+recommends "Class::XSAccessor"       => "0";
+recommends "Cpanel::JSON::XS"        => "0";
+recommends "Crypt::URandom"          => "0";
+recommends "HTTP::XSCookies"         => "0";
+recommends "HTTP::XSHeaders"         => "0";
+recommends "Math::Random::ISAAC::XS" => "0";
+recommends "MooX::TypeTiny"          => "0";
+recommends "Scope::Upper"            => "0";
```
SysPete, line 16: We don't use this anymore

share/skel/cpanfile
```diff
@@ -1,9 +1,37 @@
 requires "Dancer2" => "[d2% dancer_version %2d]";
 
-recommends "YAML"             => "0";
-recommends "URL::Encode::XS"  => "0";
-recommends "CGI::Deurl::XS"   => "0";
-recommends "HTTP::Parser::XS" => "0";
+recommends "YAML"                    => "0";
+recommends "URL::Encode::XS"         => "0";
+recommends "CGI::Deurl::XS"          => "0";
+recommends "HTTP::Parser::XS"        => "0";
```
SysPete, line 6: I don't see HTTP::Parser as a dep of either D2 or Plack, so I don't think this should be here.

share/skel/cpanfile
```diff
@@ -1,9 +1,37 @@
...
+recommends "CBOR::XS"                => "0";
```
SysPete, line 7: I don't see the reason for this one either.

share/skel/cpanfile
```diff
@@ -1,9 +1,37 @@
...
+recommends "YAML::XS"                => "0";
+recommends "Class::XSAccessor"       => "0";
+recommends "Cpanel::JSON::XS"        => "0";
```
SysPete, line 10: Since we have a dep on `JSON::MaybeXS` then this `recommends` feels wrong.

SysPete (changes requested):
A few deps should be removed, I think. Otherwise :+1:

share/skel/cpanfile
cromedome: Removed.

share/skel/cpanfile
cromedome: Done!

share/skel/cpanfile
cromedome: Author dependency. My bad. Removed.

share/skel/cpanfile
cromedome: Leaving this one per our conversation. We also have this `recommends` in the project's `cpanfile`.

cromedome:
Thanks for the feedback! This looks much better.

SysPete (approved):

racke:
:+1:

PR #1597, opened 2021-03-05
update Core/Hook.pm pod

clemep-sios:
The Core/Hook.pm pod is a little out of date. This PR removes references to non-existent "apps" property and properties parameter from register_hook pod. 
Also a minor typo fix.


cromedome:
Thanks for correcting this piece of the docs. :+1:

cromedome:
Merged, thanks!

PR #1603, opened 2021-03-15
Bundle App::Cmd 0.331

bigpresh:
See #1602 - App::Cmd > 0.331 depends on much newer perl versions, we've always supported older perls where practical, and we shouldn't suddenly pull the rug out from under our users on older perls trying to install Dancer2.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,4 +1,3 @@
-requires 'App::Cmd::Setup';
 requires 'Attribute::Handlers';
 requires 'Carp';
 requires 'Clone';
--- a/dist.ini
+++ b/dist.ini
@@ -89,3 +89,10 @@ match    = \.(png|ico|jpg)$ ; these are all binary files
 
 [ ReadmeAnyFromPod / MarkdownInRoot ]
 filename = README.mkdn
+
+
+# For anything we've had to bundle our own copy of with Dancer2, we have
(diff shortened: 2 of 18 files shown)


racke:
Sorry, but I don't really like to putting 3K lines of duplicate code into Dancer just because of the commandline app which is used by a minority of users. :-1: :-1:

bigpresh:
> Sorry, but I don't really like to putting 3K lines of duplicate code into Dancer just because of the commandline app which is used by a minority of users.

Neither do I, it's an ugly solution, but a temporary one to make Dancer2 installable for users on older perls, and allow them to mint new apps too.  The better alternative, replacing use of App::Cmd with MooX::Options isn't something any of us are likely to be able to rock up tonight and get to CPAN soon.  The other temporary band-aid solution, removing it as a dep (maybe adding to recommends instead), and having the dancer2 command-line script check for it at runtime and tell you to install it if not, is doable, but I'm unlikely to have the time for that tonight, and I still think it's a little user-hostile (following any Dancer2 guide, they're going to expect `dancer2 gen ...` to work after installing Dancer2...)

veryrusty:
Would fatpacking App::Cmd into `script/dancer2` itself be sufficient ? It'd still pull in 3K lines of code .. but it'd all reside in the script.

cromedome:
Closing in favor of #1604.

PR #1607, opened 2021-03-17
Fix broken tests following App::Cmd removal

cromedome:
The `Test::Compile` plugin to `Dist::Zilla` tries to build a compile test with all of our modules, except now that we have removed the requirement for `App::Cmd::Setup`, `App::Cmd::Setup` might not be installed, and our test suite fails. This PR addresses the problem in two ways:

* Excludes the `Dancer2::CLI` namespace from the automatic compile test generation

* Builds a compiles test specifically for the CLI that skips all if `App::Cmd::Setup` is not installed

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -28,6 +28,8 @@ add_files_in = README.mkdn
 -remove = AutoPrereqs
 -remove = GithubMeta
 -remove = ModuleBuild
+; Don't add CLI files to the auto-generated test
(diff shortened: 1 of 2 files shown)


veryrusty:
Let's add comments to dist.ini and the new test case flagging that these are temporary and can be undone/removed when #1319 is addressed.

SysPete:
:+1: with @veryrusty's comment

PR #1608, opened 2021-03-19
Migrate CLI to CLI::Osprey

### LINKED ISSUE #1319: App::Cmd is a quite a heavy (mostly optional) dependency

cromedome:
In light of the aftermath caused by recent changes to `App::Cmd` (see #1604 and #1607), move our CLI away to something lighter. For this, I chose `CLI::Osprey`. It is really lightweight, and doesn't have any dependencies we are already using. Merging of this will resolve #1319.

I may yet write some additional tests here, but this matches what we had in terms of code and testing with `App::Cmd`. I wanted to get this out here to give people a chance to review. 

I was unsure about whether or not to leave the documentation overview in `script/dancer2`, so please leave your feedback and I will address accordingly.


lib/Dancer2/CLI/Gen.pm
```diff
@@ -0,0 +1,322 @@
...
+    my( $self, $app_path ) = @_;
+    print qq{
+Your new application is ready! To run it:
+
+        cd $app_path
+        plackup bin/app.psgi
+
+If you need community assistance, the following resources are available:
+- Dancer website: http://perldancer.org
+- Mailing list: http://lists.preshweb.co.uk/mailman/listinfo/dancer-users
```
bigpresh, line 162: Probably looks more professional on the project to use lists.perldancer.org there (it's a configured vhost, it'll work) - and also decouples it from me in case hosting of the list ever needs to be moved elsewhere - bus factor and all that :)

lib/Dancer2/CLI/Gen.pm
```diff
@@ -0,0 +1,322 @@
...
+    my $version = $self->parent_command->_dancer2_version;
+    return if $version =~  m/_/;
+
+    my $latest_version = 0;
+    my $resp = HTTP::Tiny->new( timeout => 5 )->get( 'https://fastapi.metacpan.org/release/Dancer2' );
+    if( $resp->{ success } ) {
+        if ( decode_json( $resp->{ content } )->{ version } =~ /(\d\.\d+)/ ) {
+            $latest_version = $1;
+        } else {
+            die "Can't understand fastapi.metacpan.org's reply.\n";
```
bigpresh, line 308: Should this be a non-fatal warning?  If you're trying to scaffold a new app, it's nice to be warned if the version you have is a bit old, but it seems a bit OTT to prevent scaffolding your new app at all if we're unable to check (unless you ask not to; if this stays fatal, then this error message should probably tell you how to bypass the check).

script/dancer2
```diff
@@ -94,93 +81,26 @@ how to convert a Dancer (1) based application to Dancer2.
...
-The application is ready to serve:
-
-    $ cd MyWeb-App
-    $ plackup bin/app.psgi
-    >> Listening on 127.0.0.1:3000
-    == Entering the development dance floor ...
+    dancer2 <command> --help
 
-=head1 AUTHOR
+The lastest list of available commands can be displayed by:
```
bigpresh, line 91: typo - lastest -> latest

script/dancer2
```diff
@@ -94,93 +81,26 @@ how to convert a Dancer (1) based application to Dancer2.
...
+    dancer2
```
bigpresh, line 93: i'm torn here; on the one hand, I think the list of commands should be documented in a way you can easily see without having to run the code to find out.

On the other hand, that means we have to remember to keep that documentation in sync with the actual CLI app implementation... hmm.

bigpresh: I wonder if we could get cute and auto-generate the doco for it during the build process?  (If we think that would be good, that's probably one to raise as a new ticket to implement separately if anyone chooses to do so, and not block this PR with...)

bigpresh (approved):
Excellent stuff!

This looks great... and I think I may be using CLI::Osprey myself for the next scripts I write, looks nice and clean!

lib/Dancer2/CLI/Gen.pm
cromedome: LOL, duly noted. Changed!

lib/Dancer2/CLI/Gen.pm
cromedome: You are absolutely correct. I changed the way this was handled. I think it is a lot better now!

script/dancer2
cromedome: Fixed!

script/dancer2
cromedome: The two subcommands we have are documented when you run `perldoc dancer2`. I think having the option to see the subcommands when running the command without args is nice.

As far as auto-generating the docs for this script - I would LOVE this. I will add it to my list of possible documentation fixes.

lib/Dancer2/CLI/Gen.pm
```diff
@@ -301,20 +303,19 @@ sub _version_check {
...
+        if ($latest_version gt $version) {
+            print qq{
 The latest stable Dancer2 release is $latest_version. You are currently using $version.
 Please check https://metacpan.org/pod/Dancer2/ for updates.
 
 };
+        }
+    } else {
+        warn "\nCouldn't determine latest version of Dancer2. Please check your internet
+connection, or use pass -x to gen to bypass this check in the future.\n\n";
```
bigpresh, line 317: Terribly minor: "use -x" or "pass -x" would both make sense (the former more so, I'd say) but "use pass -x" sounds as though it's talking about running a `pass -x` command to me, rather than passing the `-x` option

cromedome: I started with use, rewrote *most* of it, and forgot to remove that! Thanks for catching... fixed!

SysPete (approved):
Looks great. Thank you @cromedome!

cromedome:
Merged, thanks for jumping on this!

PR #1570, opened 2020-12-22
Add test to check whether all DSL keywords appear in the manual

racke:
The following keywords are missing according to this test:
```

not ok 3 - Checking for keyword app
not ok 9 - Checking for keyword context
not ok 13 - Checking for keyword dancer_app
not ok 14 - Checking for keyword dancer_major_version
not ok 19 - Checking for keyword delayed
not ok 21 - Checking for keyword done
not ok 22 - Checking for keyword dsl
not ok 27 - Checking for keyword flush
not ok 34 - Checking for keyword header
not ok 35 - Checking for keyword headers
not ok 38 - Checking for keyword log
not ok 50 - Checking for keyword push_header
not ok 57 - Checking for keyword response
not ok 61 - Checking for keyword runner

```

PR #1609, opened 2021-03-22
Document and test missing keywords

cromedome:
Unsure what I did on #1587, but this is a cleaner version of that PR. Has all the fixes requested in #1587 with one or two extras.

As with #1587:
* Incorporates the test to ensure all DSL keywords are documented from #1570
* Splits the DSL keywords into a separate document
* Alphabetizes them
* Adds documentation for keywords that have none.


t/dsl/pod.t
```diff
@@ -0,0 +1,52 @@
...
+            if (exists $dsl_keywords->{$keyword}) {
+                $dsl_keywords->{$keyword}->{found} = $entry->[1]->{startline};
+            }
+        }
+    }
+}
+
+# go through the authoritative list and test we have a corresponding POD entry
+for my $keyword ( sort keys %$dsl_keywords ) { 
+    ok( exists $dsl_keywords->{$keyword}->{found}, "Checking for keyword $keyword" );
```
bigpresh, line 49: I'd suggest a clearer test title, e.g. "Keyword $keyword is documented in D2::Manual::Keywords" or similar, so any failures are more immediately understandable

bigpresh (approved):
One comment on making the test message clearer for the checks for documented keywords, but otherwise looks good!

t/dsl/pod.t
cromedome: Done, thanks!

cromedome:
Merged, thank you. These docs (along with our other docs) could still use improvement, but this is definitely some improvement.

PR #1614, opened 2021-04-18
Generate a Dockerfile when creating a new app

cromedome:
This is the same PR as #1589 with the same features and syntax, just reimplemented for the new CLI.

### DIFF
--- a/lib/Dancer2/CLI/Gen.pm
+++ b/lib/Dancer2/CLI/Gen.pm
@@ -72,6 +72,14 @@ option skel => (
     },
 );
 
+option docker => (
+    is       => 'ro',
+    short    => 'c',
+    doc      => 'create a dockerfile (container definition)',
+    required => 0,
+    default  => 0,
+);
+
 # Last chance to validate args before we attempt to do something with them
 sub BUILD {
     my ( $self, $args ) = @_;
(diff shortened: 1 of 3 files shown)


bigpresh (approved):
Nice!

I wonder if the instructions on how to run it via Docker should mention that the `-p 5000:5000` arg is the port inside Docker and the port it's exposed as, so if you want this app listening on a different port on the host machine, change the latter number?  Or, maybe, we should have a Dancer2::Manual::Docker or similar, or just a new section in Dancer2::Manual::Deployment, which could start out quite simple for now, and have it refer to that for more information on running D2 apps via Docker?

lib/Dancer2/CLI/Gen.pm
```diff
@@ -149,13 +163,35 @@ following commands:
...
+
         cd $app_path
         plackup bin/app.psgi
+};
+    } else {
+        print qq{
+Your new application is ready! To run it:
 
+        cd $app_path
+        plackup bin/app.psgi
```
bigpresh, line 191: It feels like this bit could be in a var like `$plackup_cmds` or similar that both branches of the if can use, rather than copied & pasted - just so that if that bit wants changing in future we don't have to remember to change both.  Then again, it's right next to each other, so not that hard!

veryrusty:
> I wonder if the instructions on how to run it via Docker should mention that the `-p 5000:5000` arg is the port inside Docker and the port it's exposed as

Note that `-p x:y` implies `x` is the locally exposed port, `y` is the port used within the running container. Using different values for x and y makes it clearer.

> Or, maybe, we should have a Dancer2::Manual::Docker or similar, or just a new section in Dancer2::Manual::Deployment, which could start out quite simple for now, and have it refer to that for more information on running D2 apps via Docker?

This is a can of worms I don't want to open. There are many way to deploy docker apps that we couldn't possibly cover them all. Some external links to docker deployment examples would be better than writing it ourselves.

cromedome:
@veryrusty If you have any resources you like in particular, please share and I can add them to the deployment manual. I think that's a better way to go than to try and write something ourselves.

lib/Dancer2/CLI/Gen.pm
cromedome: I eliminated the redundancy. Thanks for spotting that!

cromedome:
I changed the port the app runs on inside the container to highlight the difference between the internal and external ports. The output of the generation now points this out.

veryrusty (approved):
Thanks for differentiating the internal/external ports.

Looks good to me :+1:

veryrusty:
Also, I have no specific references to link to for Docker deployment docs. Once your image is built, deployment is not Dancer2 or Perl specific. Google will be a more up-to-date reference that anything we could document.

cromedome:
Merged. Thanks everyone!

PR #1617, opened 2021-05-05
Bugfix/1611 redirect forwardslash

### LINKED ISSUE #1611: redirect '/' doesn't always work as expected
When the root of a Dancer app is running under a non-root path on a website, e.g. with an app rooted at:

    https://www.example.com/dancerapp/

I would expect (although perhaps I'm wrong?) that `redirect '/'` would redirect to the URL above, but instead it results in a `Location: /` header redirecting to the root of the website.

I believe the regex on line 1278 below should be `m{^/(?!/)}`.

https://github.com/PerlDancer/Dancer2/blob/b9c740f9cb849c1c3bc9bc9a5e3449411f346052/lib/Dancer2/Core/App.pm#L1273-L1284

(As a workaround one can use `redirect '/?'`.)

veryrusty:
Consistent behaviour for `redirect '/'` and `redirect '/any/path/beginning/with/single/slash` when mounted on a non-root path. Both will get the scriptname prepended to the Location header path.

This uses @gurnec's fix from #1611, with added tests.

Resolves #1611.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1275,7 +1275,7 @@ sub redirect {
     my $destination = shift;
     my $status      = shift;
 
-    if ($destination =~ m{^/[^/]}) {
+    if ($destination =~ m{^/(?!/)}) {
         # If the app is mounted to something other than "/", we must
         # preserve its path.
         my $script_name = $self->request->script_name;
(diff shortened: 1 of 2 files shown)


cromedome:
:+1: Nice work, @veryrusty!

cromedome:
Merged, thank you!

Issue #1615, opened 2021-05-04
Remove template simple?


ghost:
I would like to discuss if it would be possible to remove template::simple in favor of using the bundled implementation of template::tiny as the default template engine in Dancer2?

At the moment, the project is bundling both template::simple and template::tiny. The docs recommend users not to run template::simple in production, so it may make more sense to steer people towards template::tiny and template::toolkit.

If nothing else, the code would become slightly smaller by removing template::simple.

Any thoughts on this?

cromedome:
Honestly, I'd like to remove our forks of `Template::Simple` and `Template::Tiny` and just use `Template::Tiny` as it exists on CPAN. Our fork of the latter isn't even used.

There are some things that we need to work through first, and most of them affect people that may have built their own skeleton apps for use with `dancer2 gen`. I am hoping to make some progress on this soon.

Thanks for bringing this idea to our attention!

masu47:
Sounds good to me.

I made a branch and removed Template::Simple, at least. Is it ok to send a pull request, for review?

cromedome:
Absolutely!

masu47:
Great!
Just created one, #1616

veryrusty:
My 2 cents:

Dancer::Template::Simple isn't a fork of Template::Simple. It was written from scratch and unfortunately has the same name as another template module on CPAN.

Dancer2::Template::Simple is there to ease anyone using Dancer::Template::Simple to migrate to D2.

Dancer::Template::Tiny was never part of Dancer's core.

Dancer2::Template::Tiny includes a fork of Template::Tiny with extra functionality. The request to merge that into Template::Tiny has sat there for over 9 years, and with Ovid's fork for Template::Tiny::Strict in the past week, safe to assume will never be included.

If I had to remove one template engine from D2 core, I'd remove Dancer2::Template::Tiny. We can turn that into its own module on CPAN for those that need or want it.

If we remove Dancer2::Template::Simple, we need to add to migration docs what the alternatives are in case someone uses that in the D1 app and wants an upgrade path.

masu47:
Good input, thank you!

Perhaps add to the migrations docs and set a deprecation time, if a decision is made to remove Template::Simple?

I can start adding to the migration docs in the pull request.

cromedome:
[...]

PR #1619, opened 2021-05-08
Clarify D2::Template::Simple's role in life

cromedome:
This is one thing to come out of #1615. Got me to thinking about why we have two template systems in core when we really want to steer devs to other template engines.

After talking with @SysPete and @veryrusty, it was decided we should update the pod here to clarify what role `Dancer2::Template::Simple` fills, and to consider splitting `Dancer2::Template::Tiny` to another distribution.

### DIFF
--- a/lib/Dancer2/Template/Simple.pm
+++ b/lib/Dancer2/Template/Simple.pm
@@ -153,15 +153,14 @@ To use this engine, you may configure L<Dancer2> via C<config.yaml>:
 
 =head1 DESCRIPTION
 
-This template engine is provided as a default one for the Dancer2 micro
-framework.
+This template engine is primarily to serve as a migration path for users of 
+L<Dancer>. This should be fine for development purposes, but you would be 
+better served by using L<Dancer2::Template::TemplateToolkit> or one of the
+many alternatives available on CPAN to power an application with Dancer2 
+in production environment. 
 
-This template engine should be fine for development purposes but is not a
-powerful one, it's written in pure Perl and has no C bindings to accelerate the
-template processing.
-
-If you want to power an application with Dancer2 in production environment, it's
-strongly advised to switch to L<Dancer2::Template::TemplateToolkit>.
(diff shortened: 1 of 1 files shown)


lib/Dancer2/Template/Simple.pm
```diff
@@ -153,15 +153,14 @@ To use this engine, you may configure L<Dancer2> via C<config.yaml>:
...
-framework.
-
-This template engine should be fine for development purposes but is not a
-powerful one, it's written in pure Perl and has no C bindings to accelerate the
-template processing.
-
-If you want to power an application with Dancer2 in production environment, it's
-strongly advised to switch to L<Dancer2::Template::TemplateToolkit>.
+This template engine is primarily to serve as a migration path for users of 
+L<Dancer>. This should be fine for development purposes, but you would be
```
racke, line 157: Maybe "It should be fine" instead of "This"?

masu47:
@cromedome This sounds very good! I did some work in #1616 to try and remove Template::Simple, but I keep running in to issues with generating a new application and using Template::Tiny to create the config  and app.psgi files from skeleton.
It is probably a better way forward to clarify the docs and the try removing the forked version of Template::Tiny instead.

lib/Dancer2/Template/Simple.pm
cromedome: Agreed. Fixed. Thanks!

bigpresh (approved):
Approved - although I think yanking it out of core entirely into its own CPAN dist might not be a bad thing.  D1 users migrating to D2 who used to use Dancer::Template::Simple would have to then go install it, but I don't think many D1 users used D::T::Simple - I think many would have switched it out for a more full-featured engine, and as long as we document clearly on the migration doco that you'll need to install it if your app used D::T::Simple, that sounds fair to me.

SysPete (approved):

cromedome:
Merged, thanks!

PR #1627, opened 2021-06-15
Proposal: Set minimum Perl to 5.10.1

SysPete:
We can no longer install on 5.8 due to upstream dependencies, so I think it is time to admit defeat and bump our min dep. This also allows us to use defined-or and state variables.

Changes:
- use 5.10.1 in Dancer.pm
- add 5.10.1 dep to cpanfile
- remove all mentions of 5.10 from docs when describing named captures
- remove special-casing in code and tests for < 5.10

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,3 +1,4 @@
+requires 'perl', 5.010001;
 requires 'Attribute::Handlers';
 requires 'Carp';
 requires 'Clone';
--- a/lib/Dancer2.pm
+++ b/lib/Dancer2.pm
@@ -2,6 +2,7 @@ package Dancer2;
 
 # ABSTRACT: Lightweight yet powerful web application framework
 
+use 5.10.1;
 use strict;
 use warnings;
 use List::Util 'first';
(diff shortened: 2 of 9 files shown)


cromedome:
:+1:... Very enthusiastically! 

While I have no empirical proof, I suspect this affects a very small subset of our uses. 5.8 is 19 years old already! It's time to put support for it out to pasture. As it is, 5.10 is almost 15 years old... I think that is plenty far back for us to support. Especially given the useful features we get from making it or minimum version. Let's make this happen!

racke:
I support it as well :+1: @SysPete  Please check if we mention the minimum Perl version in the docs and adjust if needed.

SysPete:
I've removed all mention of use of Perl < 5.10 from docs, and removed special-casing, as per updated initial comment. Docs do not specifically mention minimum Perl version, but that should show on metacpan due to `MIN_PERL_VERSION` in generated `Makefile.PL`.

cromedome:
Thanks again! Sorry for the delay.

Issue #1628, opened 2021-07-23
Replace strings in Dancer2 config with environment variables


mikkoi:
I want to run my Dancer2 application in a docker but I want to run the
container without mounting any local directories and I want to run the same
once-built container in all environments, not compile it separately for different
environments.

I would like to pass parts of the configuration as environment variables, e.g.
**DB_HOST** and **SITE_URL**, and then apply these over the values in
config.yaml when I start Dancer2 application.

Suggestion:
Change the command `config` (or the underlying file reader function)
so that it can replace environment variables in config strings, e.g.

    plugins->database->connections->main->host: ${ENV:DB_HOST}

mikkoi:
A script which creates config_local.yaml on the fly and then starts Dancer2 app would otherwise work but it means that container won't be immutable. Immutable is one of the architectural best practices listed by Google.

As it happens, this container is planned to run on Google Container Engine. You mentioned there are other ways to do this in Kubernetes. How?

Using a tmpfs is also not an option because it would mean that container is dependent on the way it is being run (the runtime environment).


Substituting variables in config.yaml is not overly complicated in my mind. Many platforms use environment variables to pass information to containers, docker or others, e.g. Java jar. I have used Openshift platform and it gave me a very good example of how env vars are governed in order to provide secrets to the container.

By stating the env vars in the configuration, we make it clear to the programmer and to operations what is expected from the runtime environment. Besides this, it also makes local development easier if you "practice dotenv" - what The Twelve-Factor App recommends. I keep my environment in a .env file and I also use the same file when testing the container with local docker.


See:
* https://cloud.google.com/architecture/best-practices-for-operating-containers
* https://docs.docker.com/storage/tmpfs/
* https://12factor.net/config

racke:
Yes, I can see the need to override Dancer config with environment variables without a start script. From my view a better place for the transformation would be a Dancer plugin if that is technically possible instead of changing Dancer's core.

mikkoi:
Plugin would be a great way to do it. Then there wouldn't be any compatibility issues. But the configuration system is not "pluggable", i.e. there is no hooks where a plugin could attach, AFAIK.

cromedome:
On Mon, Aug 23, 2021, at 3:12 AM, Mikko Koivunalho wrote:
> Plugin would be a great way to do it. Then there wouldn't be any compatibility issues. But the configuration system is not "pluggable", i.e. there is no hooks where a plugin could attach, AFAIK.
> 
[...]

PR #1629, opened 2021-07-23
Replace strings in Dancer2 config with environment variables

mikkoi:
Ref #1628 

What this PR changes?

1) After reading the whole config from different files, it traverses the whole tree and if it sees a string with a substring like "${ENV:SOME_ENV}", it changes it to the corresponding environment variable, if that environment variable is defined.
2) Before doing 1), it checks if the config item "env_var_replace" is defined and is boolean true. If so, then it executes 1). Otherwise it skips. This is to maintain backwards compatibility.
3) Unit tests included.

I am using this solution in my build and web application.

### DIFF
--- a/lib/Dancer2/Core/Role/ConfigReader.pm
+++ b/lib/Dancer2/Core/Role/ConfigReader.pm
@@ -232,6 +232,9 @@ sub load_config_file {
     }
 
     # TODO handle mergeable entries
+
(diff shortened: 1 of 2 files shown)


xsawyerx:
I see now that this was closed. I commented on the original ticket with a way that I think we can do this in core.

PR #1633, opened 2021-12-14
Fix YAML example in the tutorial

perlpunk:
Semicolons are not special in YAML and certainly not comments.

### DIFF
--- a/lib/Dancer2/Tutorial.pod
+++ b/lib/Dancer2/Tutorial.pod
@@ -1161,11 +1161,11 @@ of this tutorial. Let's edit the C<development.yml> file and add our database
 configuration there. Below the last line in that file, add the following lines,
 being careful to keep the indentation as you see it here:
 
-  plugins:                 ; all plugin configuration settings go in this section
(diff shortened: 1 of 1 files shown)


bigpresh (approved):
Good catch!

bigpresh:
The CI fail looks entirely unrelated - failed installing Getopt::Long::Descriptive:

```
 --> Working on Getopt::Long::Descriptive
Fetching http://www.cpan.org/authors/id/R/RJ/RJBS/Getopt-Long-Descriptive-0.110.tar.gz ... OK
! Configure failed for Getopt-Long-Descriptive-0.110. See /github/home/.cpanm/work/1639476716.45/build.log for details.
Configuring Getopt-Long-Descriptive-0.110 ... N/A
```

SysPete:
> The CI fail looks entirely unrelated - failed installing Getopt::Long::Descriptive:

More Perl min version creep :unamused: thanks to Rik. We need to see if we can drop that dep, as otherwise we're going to have to bump D2 to 5.12.

PR #1636, opened 2022-01-23
Feature/new config system

mikkoi:
Completely new design for the config system.

Config are loaded by classes which implement the role `Dancer2::Core::Role::ConfigReader`.
User can create custom classes to use instead of `Dancer2::ConfigReader::FileSimple`.
Custom classes are activated via env var `DANCER_CONFIG_READERS`.

### DIFF
--- a/lib/Dancer2/ConfigReader/FileSimple.pm
+++ b/lib/Dancer2/ConfigReader/FileSimple.pm
@@ -1,0 +1,144 @@
+# ABSTRACT: Ye Olde Config Reader
+package Dancer2::ConfigReader::FileSimple;
+
+use Moo;
+
+use File::Spec;
+use Config::Any;
+use Hash::Merge::Simple;
+use Carp 'croak';
+use Module::Runtime 'require_module';
+
+use Dancer2::Core::Factory;
+use Dancer2::Core;
+use Dancer2::Core::Types;
+use Dancer2::FileUtils 'path';
+
+with 'Dancer2::Core::Role::ConfigReader';
+
+has name => (
+    is      => 'ro',
+    isa     => Str,
+    lazy    => 0,
+    default => sub {'FileSimple'},
+);
+
+has config_files => (
+    is      => 'ro',
+    lazy    => 1,
+    isa     => ArrayRef,
+    builder => '_build_config_files',
+);
+
+sub read_config {
+    my ($self) = @_;
+
+    my $config = Hash::Merge::Simple->merge(
+        map {
+            warn "Merging config file $_\n" if $ENV{DANCER_CONFIG_VERBOSE};
+            $self->_load_config_file($_)
+        } @{ $self->config_files }
+    );
+
+    return $config;
+}
+
+sub _build_config_files {
+    my ($self) = @_;
+
+    my $location = $self->config_location;
+    # an undef location means no config files for the caller
+    return [] unless defined $location;
+
+    my $running_env = $self->environment;
+    my @available_exts = Config::Any->extensions;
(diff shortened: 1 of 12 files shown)


xsawyerx:
I started reviewing this and stopped because it's a bit difficult. There are movements done between files with some renaming and I need to keep going back-and-forth to see what was moved, what was altered, etc. There is no explanation in the commit. I started with comments, then saw this was the original code copied, etc. I would have preferred this to be done either with several commits - each accomplishing one clear thing, or with a description with the commits of what they're doing. (Commits that both move code *and* change the code are some of the most difficult to properly review and that's when multiple commits are *really* useful.)

I'll start with a few small comments and try to review it more thoroughly meanwhile.

1. I don't like the description "Ye Olde Config Reader." Many won't understand and it's not helpful.
2. I don't like the name `FileSimple`. Other options: `File::Simple`, `Simple::File`, or just `File`.

xsawyerx:
Here is partial coverage of what this PR does:

* It creates `Role::Config` which takes away most of `Role::ConfigReader`
  * The following attributes were moved from `Role::ConfigReader`: `default_config`, `config`, `config_files`, `local_triggers`, and `global_triggers`.
  * The following attributes were added: `config_location`, `environments_location`, `environment`, and `config_readers`.
  * The following methods were moved from `Role::ConfigReader: `_build_default_config`, `_build_environment`, `_build_config_files`, `_build_config`, `_set_config_entries`, `_set_config_entry`, `_normalize_config`, `_compile_config`, `settings`, `setting`, `has_setting`, `$_normalizers`, `_normalize_config_entry`, and `_compile_config_entry`.
  * The following methods were added: `_build_config_readers`
* After gutting `Role::ConfigReader`, it leaves it with `environment` and `location`
* It also introduces `FileSimple` which uses `Role::ConfigReader` and has a method to read the config.

This feels like a bit of a maze and I'm still making sense of it. A map would be useful. :)

bigpresh:
Thanks for the summarisation there @xsawyerx - definitely makes it easier for other reviewers, kudos.

In general, it seems like a good idea allowing much more flexibility in config reading - thanks @mikkoi for your work!

I'll try to review in more detail as soon as I get some free time.

I think it could do with a bit more documentation adding - i.e. documenting in the new D2::Core::ConfigReader::FileSimple that it is the original file-based config reader, and briefly documenting in D2::Core::Role::ConfigReader how to implement custom config readers based on it.

My understanding from a quick review is that the basic idea is for `D2::Core::Role::Config` to handle the "how to get config" stuff, and for `D2::Core::Role::ConfigReader` to be used by config readers, e.g. `Dancer2::ConfigReader::FileSimple` by default, to handle the actual reading of a config from a source.  That seems like a pretty sane design.

mikkoi:
I agree, it confuses when one commit has both file renames and changes.
I created a new PR.
Separated the commits and improved the docs a little.
https://github.com/PerlDancer/Dancer2/pull/1637

Also removed "Ye Olde Config Reader." and renamed FileSimple to File::Simple.

mikkoi:
I am wondering if I should have named `Dancer2::Core::Role::ConfigReader` something else. Now it confuses. Maybe `ConfigBuilder`, `ConfigAssembler` or `ConfigCreator`?

xsawyerx:
I appreciate the new PR very much. I think we can close this one in favor of that. I was able to review it, provide comments, and a summary with what I think we should do next.

mikkoi:
Development moved to https://github.com/PerlDancer/Dancer2/pull/1637

Issue #634, opened 2014-07-29
Defaults for environment and location in Logger::File fail.


veryrusty:
The defaults for `environment` and `location` in Logger::File were using the `context` attribute from the engine which was removed as part of the great context "untanglement".

PR #1637, opened 2022-02-02
Feature/new config system (improved)

mikkoi:


lib/Dancer2/Core/Role/Config.pm
```diff
@@ -84,9 +91,11 @@ has global_triggers => (
             },
         };
 
+        no warnings 'once'; # Disable: Name "Dancer2::runner" used only once: possible typo
```
xsawyerx, line 94: It might be worthwhile exploring this warning.

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -107,62 +116,45 @@ sub _build_environment { 'development' }
...
 
     $config = $self->_normalize_config($config);
     return $config;
 }
 
+sub _build_config_readers {
+    my ($self) = @_;
+
+    my @config_reader_names = $ENV{'DANCER_CONFIG_READERS'}
+                            ? (split qr{ [[:space:]]{1,} }msx, $ENV{'DANCER_CONFIG_READERS'})
```
xsawyerx, line 147: Is there a reason the `split` is written this way with POSIX character classes instead of `split /\s+/xms, $ENV{'DANCER_CONFIG_READERS'}` ?

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -107,62 +116,45 @@ sub _build_environment { 'development' }
...
+                            : ( q{Dancer2::ConfigReader::File::Simple} );
+
+    return [ map {
+        use_module($_)->new(
+            location => $self->location,
+            environment => $self->environment,
+            );
+    } @config_reader_names ];
```
xsawyerx, line 155:
```perl

    return [
        map use_module($_)->new(
            location    => $self->location,
            environment => $self->environment,
        ), @config_reader_names
    ];
```

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -281,8 +255,46 @@ __END__
 
 =head1 DESCRIPTION
 
-Provides a C<config> attribute that feeds itself by finding and parsing
-configuration files.
+This is the redesigned C<Dancer2::Core::Role::ConfigReader>
+to manage the Dancer2 configuration.
```
xsawyerx, line 259: I don't think it is advantageous to write the documentation as a "diff" from the previous version. It should just document the way it is.

At most, if you worried people who read this will be confused,d you could say, "starting from version XXXX, the L<ConfigReader> works this way..." or something. However, writing this in a role that is only used by developers is probably not necessary.

It's even more confusing because the class isn't called `Role::ConfigReader` but `Role::Config` and refers to a previously existing `Role::ConfigReader` even though there is a different `Role::ConfigReader` at the end of these commits.

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -281,8 +255,46 @@ __END__
...
+
+It is now possible for user to control which B<ConfigReader>
+class to use to create the config.
```
xsawyerx, line 262: "Users can control which B<ConfigReader> class to use when creating the config."

Same comment as above.

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -281,8 +255,46 @@ __END__
...
+
+Use C<DANCER_CONFIG_READERS> environment variable to define
+which class or classes you want.
+
+    DANCER_CONFIG_READERS='Dancer2::ConfigReader::File::Simple Dancer2::ConfigReader::CustomConfig'
+
+If you want several, separate them with whitespace.
```
xsawyerx, line 269: "If you want several configuration classes, express it as a string, separated by spaces."

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -281,8 +255,46 @@ __END__
...
+Configs are read in left-to-write order where the previous
+config items get overwritten by subsequent ones.
```
xsawyerx, line 271: I think an example here will help. I think the merge strategy is to only override leaf nodes - values, not references. A simple example can express that.

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -281,8 +255,46 @@ __END__
...
+
+You can create your own custom B<ConfigReader>.
+The default is to use C<Dancer2::ConfigReader::File::Simple>
+which was the only way to read config files earlier.
```
xsawyerx, line 275: No need to refer to "earlier."

lib/Dancer2/Core/Role/Config.pm
```diff
@@ -84,9 +91,11 @@ has global_triggers => (
             },
         };
 
+        no warnings 'once'; # Disable: Name "Dancer2::runner" used only once: possible typo
         my $runner_config = defined $Dancer2::runner
                             ? Dancer2->runner->config
                             : {};
+        use warnings 'once';
```
xsawyerx, line 98: Might be easier to write:

```perl
        my $runner_config;
        {
            no warnings 'once';
            $runner_config = defined $Dancer2::runner
                           ? Dancer2->runner->config
                           : {};
        }
```

I would still want to know if we could avoid that warning.

lib/Dancer2/ConfigReader/File/Simple.pm
```diff
@@ -0,0 +1,147 @@
...
+}
+
+1;
+
+__END__
+
+=head1 DESCRIPTION
+
+This class provides the same features to read configuration files
+as were earlier done by C<Dancer2::Core::Role::Config>.
```
xsawyerx, line 111: Again, no need to go over the previous implementation unless you're really worried they will get confused and, in such case, only mention which version changed it so they could compare when reading.

lib/Dancer2/ConfigReader/File/Simple.pm
```diff
@@ -0,0 +1,147 @@
...
+
+If you need to add additional functionality to the reading
+mechanism, you can extend this class. An example of this is
+in: B</t/lib/Dancer2/ConfigReader/File/Extended.pm>
```
xsawyerx, line 115: I think it's better to include an example than point to a test file.

xsawyerx (commented):
If I understand correctly, the way you see it is:
* The `App` now uses `Role::Config`
* The role `Role::Config` has `config_readers` and `config_files`
* Each of the Config Reader classes consumes `Role::ConfigReader`
* Each of these classes also has its own `config_files` which they merge together
* The `Role::Config` will then merge these "merged" config files into its own `config_files` and its configuration itself, which then the App is able to use

Did I understand this right?

This makes sense, but it also suffers from the assumption that all configurations are files, so while it's more abstract it's - and I'm sorry to say this - not abstract enough.

The naming for `Role::Config` and `Role::ConfigReader` and `ConfigReader` gave me a bit of confusion trying to grok this.

I'm also missing the role `HasLocation`. I don't see where it went. Is it no longer necessary because the location is sent through the App to `Role::Config`?

I think this warrants further discussion because I don't think we should merge this when it's halfway between being too rigid and not being flexible enough. Once you could correct whatever I got wrong, I'm happy to propose an improved structure.

mikkoi:
> If I understand correctly, the way you see it is:
> 
>     The App now uses Role::Config
>     The role Role::Config has config_readers and config_files
>     Each of the Config Reader classes consumes Role::ConfigReader
>     Each of these classes also has its own config_files which they merge together
>     The Role::Config will then merge these "merged" config files into its own config_files and its configuration itself, which then the App is able to use
> 
> Did I understand this right?

Yes, precisely.
I made a conscious design decision to keep the changes to a minimum and maintain complete compatibility.
That is the reason why `config_files` is kept even though it doesn't make much sense in this new setup.
Instead of `config_files` we should have something like `config_origin`.
The same reason why the new solution isn't "abstract enough".

Again the same applies to HasLocation. Location is deduced once and then passed on to the config readers. Just like environment name.

xsawyerx:
I think calculating `location` and `environment` only once and transferring it onwards is a good call. I like that.

I think the usage of roles here might be too much. How about:

* `Dancer2::ConfigReader` as an object holding configuration information.
* It has `config_readers` who all consume the `Role::ConfigReader` role.
* They can each have whatever attributes they want. The `ConfigReader::File::Simple` (or `Config::Type::File::Simple`) will have `config_files` as an attribute since it reads from [simple] configuration files. A different configuration might have `elastic_params` because it uses ElasticSearch to include configurations, for example.
* The `App` would create an instance of the `ConfigReader` object with the location and environment.
* The `ConfigReader` will determine which configuration reader it should use and instantiate them and use them to collect the configuration.

Two things left in mind:
* If you edit the value of the configuration readers to use using the `set` keyword, will we be able to use this by the time `ConfigReader` tries to instantiate configuration readers?
* We shouldn't (at least for now) support recursive levels of configuration readers, like "read the config file, then the config file contains other configuration readers, use them, etc." until we research how others handle this problem. (Thought: We do support local config files - so theoretically, we could use that second-level configuration part to hook into reading from non-configuration files...)

mikkoi:
If we ditch `config_files` attribute, we create a non-compatible change.

The name is bad but I like the idea that the `Config` object would actually know where the configs come from, at least on a surface level (list of files or sources).


>     If you edit the value of the configuration readers to use using the set keyword, will we be able to use this by the time ConfigReader tries to instantiate configuration readers?

I don't think its possible. For the same reason I couldn't use hooks. It creates a dependency on user code before App is ready - and config should be read and ready before App is accessible for users.

>     We shouldn't (at least for now) support recursive levels of configuration readers, like "read the config file, then the config file contains other configuration readers, use them, etc." until we research how others handle this problem. (Thought: We do support local config files - so theoretically, we could use that second-level configuration part to hook into reading from non-configuration files...)

Agree. Too many moving parts for the moment.

xsawyerx:
> If we ditch config_files attribute, we create a non-compatible change.

Non-compatible with what?

> The name is bad but I like the idea that the Config object would actually know where the configs come from, at least on a surface level (list of files or sources).

Sorry, I don't understand your comment here.

mikkoi:
Attribute `config_files` is part of the public interface, isn't it? If it isn't, then of course, it's not a compatibility issue.

Just thinking that I like the idea that the config readers maintain a list of files or sources where the config is read from.

xsawyerx:
> Attribute `config_files` is part of the public interface, isn't it? If it isn't, then of course, it's not a compatibility issue.

I can't imagine this being used for anything. The Config Reader is a role composed into classes. If someone is using it, that's not in a way that we should worry about officially supporting.

> Just thinking that I like the idea that the config readers maintain a list of files or sources where the config is read from.

It makes sense if the specific sense of configuration as files, but in the abstract sense of configuration as values by keys, files are a secondary - and more important, optional - concept.

cromedome:
> Attribute `config_files` is part of the public interface, isn't it? If it isn't, then of course, it's not a compatibility issue.
The only thing someone could be using this for, I'd think, is logging the contents of what's there (and that is already happening on app startup). Maybe some devs got clever with the contents. We could remove it and see who complains, or for non-file based configs (etcd, Vault, wherever else you might pull config from) we could put a one-line description in about where the config came from. Personally, I'd rather remove it, make a developer release, document the removal, and see who complains.

> Just thinking that I like the idea that the config readers maintain a list of files or sources where the config is read from.
As do I. I am not sure what I'd do with that info yet, but having it is intriguing. If only for some debugging purposes down the road.

lib/Dancer2/Core/Role/Config.pm
cromedome: Seconded.

lib/Dancer2/Core/Role/Config.pm
mikkoi: POSIX removed, replaced with `\s+`.

lib/Dancer2/Core/Role/Config.pm
mikkoi: Reformatted.

lib/Dancer2/Core/Role/Config.pm
mikkoi: Reformatted.

lib/Dancer2/ConfigReader/File/Simple.pm
mikkoi: Corrected.

lib/Dancer2/ConfigReader/File/Simple.pm
mikkoi: Example added.

lib/Dancer2/Core/Role/Config.pm
mikkoi: Rewritten.

lib/Dancer2/Core/Role/Config.pm
mikkoi: Example added.

lib/Dancer2/Core/Role/Config.pm
mikkoi: Rewritten.

mikkoi:
I have changed all the small things and I removed the attribute`Dancer2::Core::Role::Config::config_files`.

mikkoi:
@xsawyerx
Did you really mean we should make so big change?
To abandon App having Config as a role, and instead instantiate ConfigReader?
It means we would need to implement the functionality to access configuration content in App and in ConfigReader.

Or else, we could redesign the whole thing so that Config and ConfigReader have nothing to do with each other, except that configuration is something that ConfigReader produces and then gives/assigns to Config.

It seems like a very big change.

xsawyerx:
I do think we need to make such a big change and I don't think this big change is this "big." Of course, I could be wrong about that. (But it would still be the right thing to do.)

mikkoi:
To be on the safe side, I rebased the branch from master. So I had to force push.

The new commits are:

* [Create Role::HasEnvironment and tests](https://github.com/PerlDancer/Dancer2/pull/1637/commits/839c0be65c5f1792c6edee38e9edebdf729f6a7b)
[839c0be](https://github.com/PerlDancer/Dancer2/pull/1637/commits/839c0be65c5f1792c6edee38e9edebdf729f6a7b)
* [Add more logging to ConfigReader::File::Simple](https://github.com/PerlDancer/Dancer2/pull/1637/commits/b5afaa8b26bf0dff4171595b93b1bf7ee921880e)
[b5afaa8](https://github.com/PerlDancer/Dancer2/pull/1637/commits/b5afaa8b26bf0dff4171595b93b1bf7ee921880e)
* [Add isa type to HasLocation::location](https://github.com/PerlDancer/Dancer2/pull/1637/commits/d1597255003ebe76fc5dcc506a4b7eeea1180362)
[d159725](https://github.com/PerlDancer/Dancer2/pull/1637/commits/d1597255003ebe76fc5dcc506a4b7eeea1180362)
* [Separate normalizers from Config to ConfigUtils](https://github.com/PerlDancer/Dancer2/pull/1637/commits/94f21b7e300c1e1e39210e6f523cd4df63b5276b)
[94f21b7](https://github.com/PerlDancer/Dancer2/pull/1637/commits/94f21b7e300c1e1e39210e6f523cd4df63b5276b)
* [Split Role::Config to ConfigReader and Role::HasConfig](https://github.com/PerlDancer/Dancer2/pull/1637/commits/ffd6c4d9752cdbf95f07f700e4dccca078e6ecf2)
[ffd6c4d](https://github.com/PerlDancer/Dancer2/pull/1637/commits/ffd6c4d9752cdbf95f07f700e4dccca078e6ecf2)
* [Add Mikko Koivunalho to contributors list](https://github.com/PerlDancer/Dancer2/pull/1637/commits/c99b4b4fcd0ea483b9aa0b1f4a4efa8aa46f2077)
[c99b4b4](https://github.com/PerlDancer/Dancer2/pull/1637/commits/c99b4b4fcd0ea483b9aa0b1f4a4efa8aa46f2077)

The main changes are:
There is now Role::HasConfig which has those parts of the old Role::Config which were about reading and changing the config.
There is now Role::HasEnvironment to separate the environment reading parts (from ENV vars, etc.)
ConfigReader is instantiated as an attribute to Dancer::Core::App.

mikkoi:
The test `t/issues/gh-634.t` is failing because it uses `config_location` from `Role::Config`.
The test needs to be thought out again.

mikkoi:
The tests now run.
Do we need more fine tuning before merging?

mikkoi:
Hi.
Can we open this discussion again?
I am all open for changing anything in the code.
I have been using this code in a website for two year now.

Thank you.

cromedome:
@mikkoi I am not opposed to continuing a discussion of this, but I am booked solid until mid-May. Unsure if @xsawyerx or anyone else has time before then to work through this with you.

mikkoi:
@cromedome @xsawyerx 
Great.
I can wait.
Let's continue when you have the time.

cromedome:
@mikkoi I am starting to take another look at this. Give me a little bit to work with it :)

lib/Dancer2/ConfigReader.pm
```diff
@@ -0,0 +1,224 @@
...
+        $config->{$key} = normalize_config_entry( $key, $value );
+    }
+    return $config;
+}
+
+sub _build_config_readers {
+    my ($self) = @_;
+
+    my @config_reader_names = $ENV{'DANCER_CONFIG_READERS'}
+                            ? (split qr{ \s+ }msx, $ENV{'DANCER_CONFIG_READERS'})
```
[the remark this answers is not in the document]
mikkoi: Good thing you noticed. I wonder why in the first I used a space. A comma is the correct delimiter to use.
I changed to comma and added a test to check the failure.

mikkoi:
Renamed `Dancer2::ConfigReader::File::Simple` to `Dancer2::ConfigReader::Config::Any`.

mikkoi:
@cromedome 
I watched the [Dancer2 presentation](https://youtu.be/pCTj-lT2Y40?si=H34giMDjrP0ShaXx) at TPRC.
Great things going on and many improvements. Dancer2 is on a roll. :smile:

cromedome:
Merged! Thanks so much!

PR #1641, opened 2022-02-28
Bugfix/route matches

veryrusty:
Mixed regex and splats in paths such as `get '/(any|some)/thing/*' => sub {...}` result in more values from the match than there are regex comments and the generation of `use of unitialized value` warnings.

Add a simple test and check the token/spat comment is defined before testing equality to Shh those warnings.


cromedome:
👍 Thanks @veryrusty!

xsawyerx (approved):

cromedome:
Merged, thank you!

PR #1643, opened 2022-03-10
Require perl 5.12.5

cromedome:
Some of our dependencies now require Perl 5.12, so bump our minimum version. Again.


veryrusty:
@cromedome We should remove perl 5.10 from travis.yml and the github actions ci tests too.

cromedome:
@veryrusty Thank you for reminding me!

veryrusty:
Looks good to me. :+1:

xsawyerx (approved):

grr:
> Some of our dependencies now require Perl 5.12, so bump our minimum version. Again.

which dependecies? i just checked the MIN_PERL_VERSION on metacpan and the testers page for each and couldn't find any required dependency that require > 5.10. 

the only one i can find in the cpanfile was Dist::Zilla::Plugin::Test::UnusedVars which is only an 'author_requires', so wouldn't affect normal users.

PR #1644, opened 2022-03-12
Add deprecation policy to docs; link git guide

cromedome:
One last change to prep us for the release of 0.400000 - capture the deprecation policy in the docs.

Also we were missing a link to the git guide in the documentation map.


SysPete (approved):
LGTM, thank you @cromedome!

PR #1660, opened 2022-03-31
(Rebased) Pr/behind proxy middleware

xsawyerx:
This is a rebased branch of #590. :)

It's only for review against the original PR. Please do not merge it.

### DIFF
--- a/lib/Dancer2/Middleware/BehindProxy.pm
+++ b/lib/Dancer2/Middleware/BehindProxy.pm
@@ -1,0 +1,58 @@
+package Dancer2::Middleware::BehindProxy;
+# ABSTRACT: Support Dancer2 apps when operating behing a reverse proxy
+
+use warnings;
+use strict;
+
+use parent 'Plack::Middleware';
+use Plack::Middleware::ReverseProxy;
+use Plack::Middleware::ReverseProxyPath;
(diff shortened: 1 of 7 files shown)


cromedome (commented):
I really like how this came together. We removed a test in `t/request.t` and we don't have any tests for `lib/Dancer2/Middleware/BehindProxy.pm`. Asking here so we can capture this for our future selves: do we not need to test this going forward because of the changes, or should we re-implement the test we removed in another fashion? That's about my only concern moving forward with this.

lib/Dancer2/Middleware/BehindProxy.pm
```diff
@@ -0,0 +1,58 @@
+package Dancer2::Middleware::BehindProxy;
+# ABSTRACT: Support Dancer2 apps when operating behing a reverse proxy
```
veryrusty, line 2: Typo: *behind*

veryrusty:
I still think this is a better way to handle the psgi `env` munging needed (compared with including it as part of our Request class). 

It's taken some time, but I concur with the comments on the original PR about the namespace; something like `Plack::Middleware::Dancer2BehindProxy` would be better (and hidden from pause?)

PR #1662, opened 2022-04-01
(Rebased) rename show_errors as show_stacktrace

xsawyerx:
This is a rebase of @nunorc's work in #829.

I added another commit to make sure `show_errors` still works, but documented it's deprecated.


cromedome:
:+1:

Issue #1663, opened 2022-04-19
add_route called in a plugin BUILD block will get a prefix that cannot be overridden


GeekRuthie:
In Dancer2::Core::App::add_route, putting the prefix => option (line 1230) above the attrs that are given by the caller will let the caller override the prefix. 

If you attempt to use $self->app->add_route in a plugin's BUILD block, and your application uses prefixes, it'll *probably* get one, likely the last prefix statement in your load order, and adding prefix => undef to your add_route call doesn't help.

GeekRuthie:
Note:  I've got a patch for this coming, just adding some tests to t/app.pl to test that it works correctly. :)

PR #1665, opened 2022-05-14
Allow overriding of prefix in add_route. #1663

GeekRuthie:
The order of the arguments in add_route was such that, whenever you
called add_route from a plugin, any 'prefix' entry you submitted
was ignored utterly. Rearranging the order here lets plugin
authors override the prefix on their add_route calls, even with
undef.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1226,8 +1226,8 @@ sub add_route {
 
     my $route = Dancer2::Core::Route->new(
         type_library => $self->config->{type_library},
-        %route_attrs,
         prefix => $self->prefix,
+        %route_attrs,
     );
 
     my $method = $route->method;
--- a/t/app.t
+++ b/t/app.t
@@ -129,6 +129,66 @@ for
     is $resp->[2][0], $path, 'got expected route';
 }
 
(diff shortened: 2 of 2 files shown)


xsawyerx (approved):

xsawyerx:
Before merging, the warnings in `t/app.t` need to be investigated.

GeekRuthie:
I've pushed a commit that fills in some missing bits for the test state.  These warnings were occurring when a 404 was reached, and the SERVER_PORT and SERVER_NAME were not set.

cromedome:
Is there a reason a number of tests were commented out in `t/app.t`? The tests that changed to account for your patch make sense to me, but it seems like the ones that have been commented out are not related to the functionality in the branch. 

Entirely possible I have missed something though :)

GeekRuthie:
that was my goof.  I shorted out a bunch of other tests while I was working on the failures @xsawyerx had pointed out, and flat forgot to re-enable them.  Fix pushed to this branch.

cromedome:
Approved :+1:

@xsawyerx I'm considering your approval to still be valid as the core code remains unchanged since when you approved it.

cromedome:
Merged!

SysPete:
closed via merge of 5b56729b56db447d8ac8fc98f196c1982345f6cc

PR #1667, opened 2022-07-10
Remove failing module from GitHub actions config

cromedome:
Minion is having some issues with the SQLite backend in GitHub actions presently. I don't have the time to diagnose a non-core issue right now, and the build failure emails are annoying, so away it goes.


SysPete (approved):

cromedome:
Merged, thanks!

PR #1676, opened 2023-02-20
Stringify VERSION_FROM correctly in Makefile.PL

### LINKED ISSUE #1675: Duplicate entries in Makefile.pl for VERSION_FROM created by 'dancer2 gen -a my::app'
'dancer2 gen -a my::app' puts 2 entries (lib/my/app.pm lib/my/app.pm) in VERSION_FROM field in Makefile.PL 
this causes 'perl Makefile.pl' to fail with 
Could not open 'lib/my/app.pm lib/my/app.pm': No such file or directory at /usr/share/perl/5.34/ExtUtils/MM_Unix.pm line 2967.

removing one of the entries manually cures the error, but doesn't fix it for next project generated.

cromedome:
I'm not sure if the problem is in Path::Tiny, or how a Path::Tiny object is stringified by Dancer2::Template::Simple, but when generating a skeleton app, the combination caused the new app's `Makefile.PL` to be written with two files in `VERSION_FROM`, which produces an error when Perl tries to produce a Makefile. Explicitly stringifying the appfile path before generating the new app's Makefile prevents the problem from happening.

Fixes #1675.

### DIFF
--- a/lib/Dancer2/CLI/Gen.pm
+++ b/lib/Dancer2/CLI/Gen.pm
@@ -154,7 +154,7 @@ sub run {
 
     my $vars = {
         appname          => $app_name,
-        appfile          => $app_file,
+        appfile          => $app_file->stringify,
         apppath          => $app_path,
         appdir           => File::Spec->rel2abs( $app_path ),
         apppath          => $app_path,


xsawyerx (approved):

yanick (approved):
Looks bening enough. :-)

PR #1678, opened 2023-02-24
fix for issue about attempted deserialization of multipart form data …

emil-perhinschi:
… when a serializer is set

a quick solution for issue 1677

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -181,13 +181,10 @@ sub data { $_[0]->{'data'} ||= $_[0]->deserialize() }
 sub deserialize {
     my $self = shift;
 
-    # if a serializer is used (JSON or Mutable) the deserialization of 
-    #   the whole 'multipart' section of the request is attempted
-    #   starting with the boundary
+    # don't attempt to deserialize if the form is 'multipart/form-data'
     if (
         $self->content_type 
         && $self->content_type =~ /^multipart\/form-data/i 
-        && $self->serializer
         ) {
         return;
     }
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -185,7 +185,7 @@ sub deserialize {
     #   the whole 'multipart' section of the request is attempted
     #   starting with the boundary
     if (
-        $self->content_type =~ /^multipart\/form-data/ 
+        $self->content_type =~ /^multipart\/form-data/i 
         && $self->serializer
         ) {
         return;
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -181,6 +181,17 @@ sub data { $_[0]->{'data'} ||= $_[0]->deserialize() }
 sub deserialize {
     my $self = shift;
 
+    # if a serializer is used (JSON or Mutable) the deserialization of 
+    #   the whole 'multipart' section of the request is attempted
+    #   starting with the boundary
+    if (
+        $self->content_type =~ /^multipart\/form-data/ 
+        && $self->serializer
+        ) {
+        return;
+    }
+
+
     my $serializer = $self->serializer
         or return;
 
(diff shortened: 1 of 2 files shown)


cromedome:
Hi Emil,

Thanks for reporting the bug and for your PR!

I see what you did, and I think it's a fair approach to the problem, but in running the Dancer2 test suite against your proposed fix, it produces a *lot* of uninitialized warning messages that weren't present before. Could you please run your fix against our test suite and resolve the warnings? I'll take another look once the warnings are handled.

Thanks so much!

emil-perhinschi:
> in running the Dancer2 test suite against your proposed fix, it produces a _lot_ of uninitialized warning messages that weren't present before.

Hello Chromedome, 

I did run `prove -v` inside the lastest Dancer2 `master` branch and there were no uninitialized warnings. 
sftp://192.168.100.51/home/emilper/personal/Dancer2/prove_verbose.log

I have attached the log.

emil-perhinschi:
Hello Jason, 

you mean the uninitialized errors in https://ci.appveyor.com/project/xsawyerx/dancer2/builds/46335065 ? 

Can't find the file 00-report-prereqs.t in master ... where can I find/get those tests ? 

thank you,

Emil

emil-perhinschi:
found them, ran, made the fix, ran `dzil test` again, pushed 

Just in case, there are a few modules that are missing in `cpanfile` and not installed by Dist::Zilla: 
Test::CPAN::Meta
Dancer2::Session::Cookie
Test::NoTabs
which appear to be needed in tests

emil-perhinschi:
Is there another problem with my fix ? AppVeyor build succeeded .

cromedome (approved):
I would like to see a test to show that the bug is fixed. Nonetheless, this gets a :+1:  from me.

emil-perhinschi:
updated the pull request with a test

cromedome:
@emil-perhinschi Awesome. I ran your test without your fixes applied to verify the bug, and again with them to confirm everything works properly.  This gets a big :+1:  from me. 

@xsawyerx? @yanick? Penny for your thoughts!

lib/Dancer2/Core/Request.pm
```diff
@@ -181,6 +181,18 @@ sub data { $_[0]->{'data'} ||= $_[0]->deserialize() }
 sub deserialize {
     my $self = shift;
 
+    # if a serializer is used (JSON or Mutable) the deserialization of 
+    #   the whole 'multipart' section of the request is attempted
+    #   starting with the boundary
+    if (
+        $self->content_type 
+        && $self->content_type =~ /^multipart\/form-data/i 
+        && $self->serializer
```
yanick, line 190: I think this last condition is not necessary. For one it's taken care of at line 196 right below, and for another I think the logical condition here is "no matter if we have a serializer or not, we don't no nuthing if the content type is `multipart/form-data`. Nuthing, I tell you.

yanick (changes requested):
I have this one comment I left, but beside that, looks good!

lib/Dancer2/Core/Request.pm
```diff
@@ -181,6 +181,18 @@ sub data { $_[0]->{'data'} ||= $_[0]->deserialize() }
 sub deserialize {
     my $self = shift;
 
+    # if a serializer is used (JSON or Mutable) the deserialization of
```
yanick, line 184: I think the comment describe the problem the code fixes. But without context I am prone to think it describes what the block does. 

Maybe `# don't attempt to deserialize if the form is 'multipart/form-data'`

emil-perhinschi:
updated as requested

yanick (approved):

cromedome:
This was merged but never closed. Thanks again! :)

emil-perhinschi:
thank you

will it make into a CPAN release ? else we can continue using github and manual installs, but that's not the optimum solution :)

cromedome:
It's queued up for the next release, tentatively next week.

PR #1680, opened 2023-07-22
Minor docs fix #1669

### LINKED ISSUE #1669: Dancer2::Core::Error document is obsolete
The documentation for `Dancer2::Core::Error` includes instantiating a new object and then using the `render` method (which doesn't exist anymore) and then setting it as the response using `Dancer2::Core::Response`'s `set`, which doesn't exist.

Instead, we just create a new instance and call `throw`.

GeekRuthie:
As @xsawyerx pointed out in [https://github.com/PerlDancer/Dancer2/issues/1669](#1669) , the Synopsis on the POD for this module were incorrect, showing an older style of functionality.  Here's a trivial update to the documentation

### DIFF
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -475,7 +475,6 @@ __END__
 
 =head1 SYNOPSIS
 
-    # taken from send_file:
     use Dancer2::Core::Error;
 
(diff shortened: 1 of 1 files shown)


cromedome:
Approved and merged :+1: . Thank you!

PR #1682, opened 2023-07-31
Bump minimum version of Perl to 5.14

cromedome:
Moo now has a minimum version of Perl 5.14, and without Moo, we have nothing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,7 @@ jobs:
         perl-version:
           - '5.32'
           - "5.30"
+          - "5.14"
 #          - "5.28"
 #          - "5.26"
 #          - "5.24"
@@ -95,8 +96,6 @@ jobs:
 #          - "5.20"
 #          - "5.18"
 #          - "5.16"
-#          - "5.14"
-          - "5.12"  # Not available?
 
     container:
       image: perl:${{ matrix.perl-version }}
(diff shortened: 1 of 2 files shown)


racke (approved):

yanick (approved):

grr:
> Moo now has a minimum version of Perl 5.14,

i see no evidence of this statement. can't even find a branch or PR in the Moo repo to back this up.

GeekRuthie:
You're both right here--to a point.  According to [this link](https://metacpan.org/pod/Moo#COMPATIBILITY-WITH-OLDER-PERL-VERSIONS) in the current Moo docs, if you're going to run it prior to 5.14, it will likely require additional dependencies.

Moo **should** run on versions farther back with those additional dependencies, just as Dancer likely **will** run on older versions of Perl, going back a good ways.  But there's no guarantee it will do so, and we do not support work dependent on those older Perls.

The decision by the Core Team has been made to follow the Toolchain Gang's decision in [the Lyon Amendment](https://github.com/Perl-Toolchain-Gang/toolchain-site/blob/master/lyon-amendment.md); our commitment is to support versions of Perl that are ten years or newer--as of this writing, 5.16, and in 2024, 5.20. So both this ticket and #1643 are pretty much moot, really.

It sounds like you're picking at a nit here, @grr, especially given you also responded to #1643  in a very similar fashion.  Is there a reason for this, something with code you're using or building that requires older Perls?

PR #1686, opened 2023-08-19
Change GitGuide to the more standard Contributing

cromedome:
Many larger projects I see these days (and a lot of smaller ones too) have a file in the repo called Contributing. I think it makes it much clearer what the document is all about.


cromedome:
There’s more to be done with this.

cromedome:
@yanick @GeekRuthie *now* this is ready! :)

GeekRuthie (approved):

PR #1688, opened 2023-08-20
Fix various bugs/issues in Dancer2 Pod

cromedome:
Several minor fixes that seemed silly to put on separate PRs.

I'm not sure if dropping the pod for `runner()` will be controversial or not. I don't think it fits with the rest of this information though.

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -110,7 +110,7 @@ Returns the current runner. It is of type [Dancer2::Core::Runner](https://metacp
 # SECURITY REPORTS
 
 If you need to report a security vulnerability in Dancer2, send all pertinent
-information to [mailto:dancer-security@dancer.pm](mailto:dancer-security@dancer.pm). These matters are taken
+information to [dancer-security@dancer.pm](mailto:dancer-security@dancer.pm). These matters are taken
 extremely seriously, and will be addressed in the earliest timeframe possible.
 
 # SUPPORT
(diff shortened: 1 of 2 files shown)


xsawyerx:
What's the reason for removing the pod for `runner`?

cromedome:
Look at our [README](https://github.com/PerlDancer/Dancer2#readme) - it's full of a bunch of good information about what dancer is, where to find different docs, how to get help, etc. And then you have what seems to be one random function thrown in the middle. It seems out of place.

Now, consider that this is also the pod when you run `perldoc Dancer2` - it makes quite a lot more sense. 

This is a situation of having my cake and wanting to eat it too. I don't really like seeing the runner pod in the README, but I think it belongs in the pod. We could not autogenerate the README and keep a separate one, but now we'd be maintaining the same information in two places - the README and `Dancer2.pm`. That's not great.

I'm not sure we're able to not include the `runner` docs when auto-generating the README.

So that's where this comes from. Perhaps the solution is `worry about actually important things`?

yanick (approved):
:100:

GeekRuthie (approved):

cromedome:
I suppose my other objection to keeping Pod for `runner()` it's the only thing documented in `Dancer2.pm`. If you look at distributions like DBIx::Class and Mojolicious (or frameworks from other languages), they follow this convention too. 

I promise I am not trying to argue this; your comments @xsawyerx forced me to justify this change to myself as much as anything :) 

Given the approvals, I am merging this. That being said, I am happy to reverse or reopen this issue at any time.

Issue #1712, opened 2024-06-27
Unable to use send_as in template


abeverley:
This is a follow-on problem as a result of #1622 and its fix 3331ad7062

In the highly unlikely event that `send_as` is called during the render of a template, the error `Unable to load serializer class for plain: Can't locate Dancer2/Serializer/plain.pm` will be experienced. This is because 3331ad7062 removes the longjump when a template is processed, which means statements [such as this](https://github.com/PerlDancer/Dancer2/blob/f174c092c4568adbfbabd3aa74a479f5e34f1f9a/lib/Dancer2/Core/App.pm#L975) do not return immediately, leading to unexpected behavior when the subsequent statements are executed.

The use-case here is for the `Dancer2::Plugin::LogReport` plugin, which (in the worst case scenario if can't do anything else) will render a plain text error if there is an unhandled exception when rendering a template.

I have a proposed solution which I will submit as a separate PR.

cromedome:
Resolved with merge at 1f913fde. Thanks!

PR #1713, opened 2024-06-27
Fix unable to use send_as during template processing (#1712)

abeverley:
Fix the problems as described in #1712. Namely these are further issues subsequent to #1622 regarding unusual actions that take place during template rendering. This could be, for example, custom exception handling during failed template rendering, which result in plain text being sent back to the client instead of the template.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -896,10 +896,16 @@ sub template {
         $self->set_with_return( sub {
             $local_response ||= shift;
         });
-        my $content = $template->process( @_ );
+        # Catch any exceptions that may happen during template processing
+        my $content = eval { $template->process( @_ ) };
         $self->set_with_return($old_with_return);
+        # If there was a previous response set before the exception (or set as
+        # part of the exception handling), then use that, otherwise throw the
+        # exception as normal
         if ($local_response) {
             $self->with_return->($local_response);
+        } elsif ($@) {
+            die $@;
         }
         return $content;
     }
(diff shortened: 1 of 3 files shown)


lib/Dancer2/Core/App.pm
```diff
@@ -896,10 +896,16 @@ sub template {
...
+        # Catch any exceptions that may happen during template processing
+        my $content = eval { $template->process( @_ ) };
         $self->set_with_return($old_with_return);
+        # If there was a previous response set before the exception (or set as
+        # part of the exception handling), then use that, otherwise throw the
+        # exception as normal
         if ($local_response) {
             $self->with_return->($local_response);
+        } elsif ($@) {
+            die $@;
```
skington, line 908: I think it would be more robust to *immediately* take a copy of $@ after calling eval. For all you know, set_with_return - or any future code added between the eval and the check of $@ in the future - might modify $@, which is after all a global variable.

abeverley: Thanks, implemented with b36481f2

cromedome: Nice suggestion, @skington!

cromedome:
:+1: from me. Will publish another release as soon as we get another +1 on this.

yanick (approved):
Looks good to me. And :100: for the test cases!

GeekRuthie:
Looks good to me, also!

cromedome:
Merged, thanks!

abeverley:
Fantastic, thanks all.

Issue #1716, opened 2024-10-20
Scope:User not installable


waghanza:
Hi,

With the following dockerfile
```dockerfile
FROM perl:5.41-slim

RUN apt-get -qq update && \
  apt-get -qy install --no-install-recommends build-essential

WORKDIR /usr/src/app
COPY cpanfile d2app.psgi ./
RUN cpanm --notest --installdeps .

CMD plackup -s Gazelle --port 3000 --env production --app d2app.psgi
```

I can not install Scope::Upper

```
71.56 --> Working on Scope::Upper
71.56 Fetching https://backpan.perl.org/authors/id/V/VP/VPIT/Scope-Upper-0.34.tar.gz ... OK
72.19 Configuring Scope-Upper-0.34 ... OK
72.32 Building Scope-Upper-0.34 ... ! Installing Scope::Upper failed. See /root/.cpanm/work/1729441177.7/build.log for details. Retry with --force to force install it.
72.67 FAIL
72.72 --> Working on Class::XSAccessor
72.72 Fetching https://backpan.perl.org/authors/id/S/SM/SMUELLER/Class-XSAccessor-1.19.tar.gz ... OK
73.28 Configuring Class-XSAccessor-1.19 ... OK
73.39 Building Class-XSAccessor-1.19 ... OK
74.73 Successfully installed Class-XSAccessor-1.19
74.82 --> Working on Math::Random::ISAAC::XS
74.82 Fetching https://backpan.perl.org/authors/id/J/JA/JAWNSY/Math-Random-ISAAC-XS-1.004.tar.gz ... OK
75.44 Configuring Math-Random-ISAAC-XS-1.004 ... OK
75.63 ==> Found dependencies: Test::NoWarnings
75.69 --> Working on Test::NoWarnings
75.69 Fetching https://backpan.perl.org/authors/id/H/HA/HAARG/Test-NoWarnings-1.06.tar.gz ... OK
76.14 Configuring Test-NoWarnings-1.06 ... OK
76.25 Building Test-NoWarnings-1.06 ... OK
76.46 Successfully installed Test-NoWarnings-1.06
76.50 Building Math-Random-ISAAC-XS-1.004 ... OK
77.03 Successfully installed Math-Random-ISAAC-XS-1.004
77.07 ! Installing the dependencies failed: Module 'Scope::Upper' is not installed
77.07 ! Bailing out the installation for ..
```

any idea ?

Regards,

PS : I'm on ARM, mac m1

[...]

PR #1717, opened 2024-10-22
Remove Scope::Upper recommends

veryrusty:
Ref #1716

Thanks to @waghanza for noting that Scope::Upper is failing to build on devel perl releases. 
Thanks to @SysPete for confirming we should have removed the recommends on Scope::Upper when Return::MultiLevel was removed.


SysPete (approved):
Thanks @veryrusty!

cromedome (approved):
Thanks @veryrusty ! I'll get a new release put together soon.

PR #1718, opened 2024-10-30
Add Module::Pluggable as a requirement, excluding broken versions

veryrusty:
Module::Pluggable added checks for unreadable files in v6.1 and v6.2. Those tests fail when installed as root, which may happen under docker, such as under github actions.

While we have no explicit dependency on Module::Pluggable, our dependencies do, such as Config::Any.

@cromedome this _should_ be sufficient for the github actions to "just work". They run in docker containers as the root user.  This is a bit of a crutch until Module::Plugable is patched. Upstream issue: https://github.com/simonwistow/Module-Pluggable/issues/27

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -22,7 +22,6 @@ requires 'Import::Into';
 requires 'JSON::MaybeXS';
 requires 'List::Util', '1.29';   # 1.29 has the pair* functions
 requires 'MIME::Base64', '3.13'; # 3.13 has the URL safe variants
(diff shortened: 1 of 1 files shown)


cromedome (approved):
The second commit is particularly great, not having to bump the version of another dependency! :+1:

GeekRuthie (approved):

cromedome:
Merged, thanks!

PR #1719, opened 2024-11-25
Force install Module::Pluggable so CI can run

cromedome:
GitHub Actions, regardless of what we tell it, still insists on installing the incorrect version of Module::Pluggable. Instead, we'll hit it with a really big hammer and force it to install.

This can be removed once the author fixes the issue installing Module::Pluggable as root.

I don't like doing this but I also don't like holding up a release because tests aren't passing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
       - name: Install dependencies
         run: |
             perl -v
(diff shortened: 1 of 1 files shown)


veryrusty (approved):
Unfortunately this is the module we are looking for. :+1: for using the force.

Issue #1722, opened 2025-04-07
Configurable Template (Toolkit) class


markov2:
My application adds translations to (Template Toolkit's) Template.pm via the extension mechanism.  However, this is currently not easy to merge with Dancer2's TT integration Dancer2::Template::TemplateToolkit, where the use of the Template class is hard-coded.

As work-around, I had to clone this implementation into Dancer2::Template::LogReport (to be released)

Please offer a way to specify the Template extension class.

abeverley:
Just to mention that a way to achieve this would be to simply add an attribute for the template base class used in `Dancer2::Template::TemplateToolkit`. This would default to `Template` and then could be overridden.

cromedome:
Resolved by #1723. Closing.

PR #1723, opened 2025-04-14
Enable use of a different Template::Toolkit base class (#1722)

abeverley:
Enable use of different Template::Toolkit base class in `Dancer2::Template::TemplateToolkit` (see #1722)


cromedome:
@abeverley This looks good! Can you add a test and a quick bit of docs to it?

yanick (approved):
Looks good to me!

PR #1727, opened 2025-06-11
Don't create CPAN package files when generating new apps

cromedome:
This... is kinda fluff, I admit. Don't add files typically used to package applications for CPAN when generating a new app.

If you use a custom app skeleton, this could be mute. But if you're like a lot of people and use the default skeleton, this is one little bit of less work after the fact.

I considered making this the default but didn't want to change what is accepted default behavior.


GeekRuthie (approved):
This is *very* good fluff.  I'll use that feature a fair bit, I betcha.

PR #1728, opened 2025-06-12
split cookie values on `&` only

### LINKED ISSUE #1701: Cookie parsing bug
Hey!

I found a bunch of warnings for a service at work built on Dancer2. When I did some investigating, I figured out it was because we were setting some values in the cookies that aren't handled properly when the browser hits the site again.

```perl
#!/usr/bin/env perl

use Dancer2;
use Data::Printer;

my $cookie_value = "hmm; this & that";

any '/' => sub {
    my $cookie = cookie("test");
    if ($cookie) {
        my @values = $cookie->value();
        p @values;
    }
    cookie("test", $cookie_value, expires => '+30m');
    send_as plain => "NOM NOM NOM";
};

dance();
```

The cookie header from curl looks like

```
< Set-Cookie: test=hmm%3b%20this%20%26%20that; Expires=Tue, 31-Oct-2023 01:33:10 GMT; HttpOnly; Path=/
```

Awesome. But when I hit it in the browser, it dumps out these values:

```
[
    [0] "hmm; this ",
    [1] " that"
]
```

That's with HTTP::XSCookies installed. If I take that out, I get

```
< Set-Cookie: test=hmm%3B%20this%20%26%20that; Path=/; Expires=Tue, 31-Oct-2023 01:28:47 GMT; HttpOnly
```

and

```
[
    [0] "hmm",
    [1] " this ",
    [2] " that"
]
```

I think the problem is in Dancer2::Core::Request::_build_cookies. It can split the cookie values:

```perl
    # convert to objects
    while (my ($name, $value) = each %{$cookies}) {
        $cookies->{$name} = Dancer2::Core::Cookie->new(
            name  => $name,
            # HTTP::XSCookies v0.17+ will do the split and return an arrayref
            value => (is_arrayref($value) ? $value : [split(/[&;]/, $value)])
        );
    }
```

I have no idea why the split behaves differently between HTTP::XSCookies and Cookie::Baker. As far as I can tell, the parsed cookie value is identical.

yanick:
Seems that ::Request was being overly zealous and was splitting the cookie value on `;` as well. This is not the same behavior as HTTP::XSCookie, which only split on `&`.

Fix #1701

### DIFF
--- a/lib/Dancer2/Core/Request.pm
+++ b/lib/Dancer2/Core/Request.pm
@@ -579,7 +579,7 @@ sub _build_cookies {
         $cookies->{$name} = Dancer2::Core::Cookie->new(
             name  => $name,
             # HTTP::XSCookies v0.17+ will do the split and return an arrayref
-            value => (is_arrayref($value) ? $value : [split(/[&;]/, $value)])
+            value => is_arrayref($value) ? $value : [split '&', $value ]
         );
     }
     return $cookies;
--- a/t/cookie.t
+++ b/t/cookie.t
@@ -4,7 +4,6 @@ use Test::Fatal;
 use Test::More;
 
 BEGIN {
-
     # Freeze time at Tue, 15-Jun-2010 00:00:00 GMT
     *CORE::GLOBAL::time = sub { return 1276560000 }
 }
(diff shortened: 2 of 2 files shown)


cromedome:
Thanks @yanick! :+1:

cromedome (commented):
Still :+1: from me. Thanks for looking at the test failures.

yanick:
It's obviously a :+1: from me too. So I say: merge it!

PR #1729, opened 2025-06-14
make the censor function customizable

yanick:
related to #530


lib/Dancer2/Core/Error.pm
```diff
@@ -48,6 +48,27 @@ has title => (
...
 
+has censor => (
+    is => 'ro',
+    isa => CodeRef,
+    lazy => 1, 
+    default => sub {
+        my $self = shift;
+
+        my $custom = $self->has_app && $self->app->setting('error_censor') 
+            or return \&_censor;
```
yanick, line 59: Related to #530, do we want to change our built-in _censor for `Data::Censor` directly? Is the code reduction worth the new dependency?

cromedome: I can go either way on this. I agree with @bigpresh's assessment in the original ticket, but what we have seems good enough for most things. @bigpresh or anyone else care to weigh in here?

bigpresh: Obviously I'd be more than happy to see Data::Censor being used (and it would reduce a little code, and Data::Censor itself has no  further dependencies to worry about) but I don't think there's a massive benefit either way.

What I think would be valuable would be to be able to configure the given censoring solution, which isn't really possible here as far as I can see.  If the `censor` attribute could be given an object that has a `censor` function, then you could instantiate your own Data::Censor object with your desired settings (e.g. custom lists of sensitive fields, custom replacements, callbacks per field etc) then pass it to the `censor` attribute to be used.

If we did switch to depending on Data::Censor and using it as the default censoring method, then maybe we could provide easy config settings to help configure it - but being able to pass in your own pre-configured object is probably enough?

yanick:
I'll bring in the config change to take an object, and make Data::Censor the default. The new patch should come in soon!

yanick:
There we go. Once more thing punted out of core! \o/

yanick:
I think it's good and ready for the merge?

bigpresh (approved):
Some suggestions left, but other than the sensitive fields pattern the rest are take-it-or-leave-it suggestions, I'm happy enough as it stands :)

Sorry for being a bit slow to respond.

lib/Dancer2/Core/Error.pm
```diff
@@ -48,6 +48,51 @@ has title => (
...
+
+            my $module = $custom =~ s/::[^:]*?$//r;
+
+            require_module($module);
+
+            return eval '\&'.$custom;
+        }
+
+         my $data_censor = use_module('Data::Censor')->new(
+             sensitive_fields => qr/pass|card?num|pan|secret/i,
```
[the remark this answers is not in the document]
yanick: Good eye! I moved the default that was there for the old `_censor` function, and didn't see it had been wrong all along! :-)

As for including other tokens, my reasoning was to keep the config exactly the same than what we had with `_censor` so that the change is non-breaking.

lib/Dancer2/Core/Error.pm
```diff
@@ -523,6 +537,60 @@ This is only an attribute getter, you'll have to set it at C<new>.
 
 The message of the error page.
 
+=attr censor 
+
+The function to use to censor error messages. By default it uses the C<censor> method of L<Data::Censor> C<_censor>"
```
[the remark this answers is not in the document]
yanick: Extra censor censored!

lib/Dancer2/Core/Error.pm
```diff
@@ -523,6 +537,60 @@ This is only an attribute getter, you'll have to set it at C<new>.
...
+
+         # default censor function used by `error_censor` 
+         # is equivalent to
+         sub MyApp::censor {
+             Data::Censor->new(
+                sensitive_fields => qr/pass|card?num|pan|secret/i,
+                replacement      => "Hidden (looks potentially sensitive)",
+            )->censor(@_);
+         }
+         setting error_censor => 'MyApp::censor';
```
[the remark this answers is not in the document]
yanick: I think we should keep the details explicit, so that people actually know what would be censored and what would not. But I'll add the comment, it totally doesn't hurt. :-)

t/lib/CustomCensor.pm
```diff
@@ -0,0 +1,11 @@
+package CustomCensor;
```
cromedome, line 1: It appears as if you created this and didn't use it. Am I missing something?

That's my only question. Otherwise, this is 👍 from me. Really like the changes you made!

cromedome (commented):
Just the one question; this is pretty good to go!

t/lib/CustomCensor.pm
yanick: ooops. Looks like I started thinking I'd use it and code/thoughts went elsewhere. Removed!

cromedome:
Merged, thanks @bigpresh and @yanick!

PR #1730, opened 2025-06-15
log hook entries as they are executed

yanick:
For #1512


bigpresh (approved):
This looks very useful!

PR #1731, opened 2025-06-18
Retire our Template::Tiny fork; use CPAN's

cromedome:
We ported the changes from our fork to the official Template::Tiny distribution, eliminating the need for our fork. This PR makes explicit use of Template::Tiny 1.16 that contains the changes our fork once had, and removes all tests and forked code from our repo. The template adapter now looks for the "official" version.

Thanks to @ether, @dams and @yanick for helping to make this happen!

PR #1732, opened 2025-06-18
Split Dancer2::Template::Simple from core

cromedome:
This is primarily useful for migrating from Dancer (1) anyhow. We were only using it in core to generate new apps from skeletons, and used Template::Tiny for everything else. Now we will have no template implementations in core (only template engines).

I'll release this as its own CPAN dist when we release Dancer2 2.0.0. It's pretty well ready to go, but no need to confuse people until then.

This depends on #1731 and cannot be merged until that is.


cromedome:
Merged, thanks everyone!

PR #1735, opened 2025-06-20
Scaffold tutorial app; scaffold other apps shipped with Dancer2 core

### LINKED ISSUE #850: Scaffold tutorial code as an example
new user poking around dancer2.  the tutorial looks interesting:  http://search.cpan.org/~sukria/Dancer2-0.01/lib/Dancer2/Tutorial.pod .

alas, the idea of a tutorial should not just be to illustrate, but also to work out of the box.  could someone please tar up a compute working tutorial, including all the files in the right location?  I have gotten to print the template file to login, but the template was not processed.  I obviously made at least one mistake, and probably more.

### LINKED ISSUE #1115: Profiles for command line scaffolding

cromedome:
This ships the tutorial app as a new app scaffold, with the ability to specify which app skeleton in a directory you wish to scaffold with. Currently, the functionality is:

`dancer2 gen -a TestApp`

Which produces a new app from the default skeleton. You can pass an additional argument, `--skel-name`, or `-n` to specify a different skeleton, ala:

`dancer2 gen -a TestApp -n tutorial`

We can potentially ship other skeletons our users could find useful, but we should be cautious about this, because keeping these up to date adds some maintenance overhead for us.

In conjunction with the `--skel`/`-s` option, users can specify a directory with their own library of default app skeletons though:

`dancer2 gen -a TestApp -s /path/to/my/app/templates -n tutorial`

Closes #850. Closes #1115.


bigpresh (approved):
Looks good!

My only minor reservation is whether we should bother including jQuery, vs just pointing the `<script>` tag to pull it from [jQuery CDN](https://releases.jquery.com/) -  we already do that for `bootstrap` so it would make sense to do the same for jQuery I think, rather than ship a great big bundle of Javascript in the Dancer2 dist?

cromedome:
I'll push an updated version today/this weekend that uses jQuery from CDN in the default app, and removes it from the tutorial app (it's not used, but a remnant of having generated a new app from the CLI). Thanks for the idea!

veryrusty:
@cromedome The two `cpanfile`'s need another look
* If this is going out with Dancer2 2.0, `default/cpanfile` needs Dancer2::Template::Simple added
* The tutorial uses Dancer2::Template::TemplateToolkit, but that's not listed in the tutorials' cpanfile.

cromedome:
> @cromedome The two `cpanfile`'s need another look
>
Nice catch!
 
>     * If this is going out with Dancer2 2.0, `default/cpanfile` needs Dancer2::Template::Simple added
>
It should be using Tiny now, so I updated it appropriately. I'll update docs accordingly on another branch that I have in progress.
 
>     * The tutorial uses Dancer2::Template::TemplateToolkit, but that's not listed in the tutorials' cpanfile.

Thanks, fixed!

@veryrusty @bigpresh please take another look when you have a few minutes. Appreciated!

cromedome:
Merged, thanks!

Issue #1737, opened 2025-07-06
Adding an error hook for errors during hook processing


abeverley:
I am interested in adding a hook for errors that take place during hook execution (in addition to the other [error hooks](https://metacpan.org/pod/Dancer2::Manual#Error-Hooks)).

My particular interest is for exceptions that happen during the `after` hook. At the moment, there is little that can be done with such an error once it takes place. Given that exceptions should probably never happen in this situation, I would like to be able to control these better, rather than having a standard Dancer2 `error` take place.

If I write a PR for such a hook is there a chance it can be merged?

Thanks,

Andy

cromedome:
@abeverley you betcha! I am a bit behind on the planned release, so I'll be happy to hold a bit longer for this to get included.

abeverley:
Wow, that would be great, thanks. I'll aim to get a PR with you tomorrow.

cromedome:
Resolved by #1738

PR #1738, opened 2025-07-08
Add new hook for exceptions in hooks (#1737)

abeverley:
This adds a new type of hook that is called in the event of an exception in a hook. Because of its nature it's a bit more complicated than a normal hook, in particular because it could result in recursive calling.

Because this hook can be called in quite a variety of ways during response handling, and because the current hook exception behavior is quite blunt, I also wanted to make it quite flexible in terms of controlling what happens. In particular, I wanted to be able to change the default response in the event of an exception in a hook. The comments in the code and the documentation should all be self-explanatory in this regard.

For #1737


yanick (approved):
Looks really good. Has a +1 from me!

cromedome:
Awesome! 👍

abeverley:
Brilliant, thanks both!

PR #1740, opened 2025-07-20
Remove Dancer2::Template::Simple from tests

cromedome:
So I still have Dancer2::Template;:Simple installed locally, so some tests passed when they shouldn't. Thankfully we have Github Actions. This PR fixes the tests I missed in my earlier PR/


bigpresh (approved):

veryrusty (approved):
:+1: Looks good. Thanks @cromedome 

There's a couple of other references to Template::Simple still about that need attention; I'll note those elsewhere.

cromedome:
Merged, thank you!

PR #1741, opened 2025-08-19
Cover items missed in earlier documentation branch

cromedome:
This adds some config docs for things missed as part of the grant, and fixes up some other inconsistencies and omissions in the manual, migration guide, etc.


bigpresh (approved):
Looks cromulent!  My initial reaction was I wasn't sure about the cutesyness, but it adds some character and fun to the docs, so I'm not against it :)

PR #1742, opened 2025-08-29
Rename CLI options for future expansion

cromedome:
An upcoming version of Dancer2 will allow you to scaffold plugins and config readers in addition to applications. As they currently stand, the existing CLI options are only meaningful to applications. This slight change makes the same options more meaningful for all types of projects.

### DIFF
--- a/lib/Dancer2/CLI/Gen.pm
+++ b/lib/Dancer2/CLI/Gen.pm
@@ -17,12 +17,12 @@ use IPC::Open3 qw();
 use Try::Tiny;
 use File::Which;
 
-option application => (
+option name => (
     is            => 'ro',
-    short         => 'a',
-    doc           => 'application name',
+    short         => 'n',
+    doc           => 'project name',
     format        => 's',
-    format_doc    => 'appname',
+    format_doc    => 'projectname',
     required      => 1,
     spacer_before => 1,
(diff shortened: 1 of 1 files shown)


bigpresh (commented):
Potentially a "breaking change" for anyone who's got used to using `dancer2 gen -a MyName` but I can see the logic behind it, and if we're going to make a change like that, a major milestone version is the time :)

I do wonder if we should continue to support the old form at least for now as an alternative, though?  (Not sure how easily/possible that is via CLI::Osprey, not very familiar with it)

It'll want to be made clear in the changelog that `dancer2 gen -a MyApp` now needs to be `dancer2 gen -n MyApp`

If we're making a confusing/breaking change here, should we add an arg for what we're generating e.g. `-t` / `-type` with the default being `app`, so you could document `dancer2 gen -t app -n MyApp` (leaving the clear path open for future generation of other _things_ e.g. `dancer2 gen -t plugin -n MyPlugin`)?

Or would `dancer2 gen ...` be for generating a new app, and have new "commands" for other stuff, like .e.g `dancer2 gen-plugin -n MyPlugin` ?

Possibly wants a little more thought here?

yanick (approved):
Looks simple enough!

Issue #763, opened 2014-11-13
Strict configuration mode


xsawyerx:
Dancer 1 has a strict configuration mode that makes sure you don't hit configuration keys that don't have values (because you might have misspelled the key).

@Ovid requested we add it to Dancer2 as well.

xsawyerx:
What the hell...?

xsawyerx:
I'm just now realizing we still have stuff in the Config document that don't exist in D2.

cromedome:
Having been bit by my own typos twice in the last two days, I am suddenly feeling the urge to work on this... keep you all posted.

cromedome:
@DavsX - someone can correct me if I am wrong here, but I think objects in Moo are just blessed references to hashes. I don't think we prevent someone from accessing the internals of the config object if we try.

All the same, I think if strict mode is enabled, the encouraged behavior needs to be config->setting.

Spending some time this morning abusing Moo and seeing how to accomplish this.

mikkoi:
Hi,
Any progress on this issue?

PR #1754, opened 2026-01-12
Strict config: Warn on unknown config keys with opt-out

xsawyerx:
Closes GH #763:.

This adds a default-on config option `strict_config` that checks whether all of the keys in the config are spelled correclty and exist and warns accordingly.

It checks for top-level keys and known built-in engines. It does *not* check for any plugin or other cases that do not come with Dancer2 because they are able to access arbitrary keys.

(Of course, it does not know how to check for misspelled `strict_config`. :)

Includes allowlists for core keys and built-in engines (logger/session/ serializer/template) so valid engine options don’t produce noise.

Add a focused regression test that exercises warnings and the opt-out.

We might want to add this to the default scaffolding.

### DIFF
--- a/lib/Dancer2/Manual/Config.pod
+++ b/lib/Dancer2/Manual/Config.pod
@@ -402,6 +402,17 @@ the options are known. Set this to false if you intentionally store
 application-specific settings at the top level or use custom engines with
 additional configuration keys.
 
+=head3 strict_config_allow (array)
+
+List of top-level configuration keys that should be ignored by strict config
+warnings. This allows you to keep C<strict_config> enabled while permitting
+application-specific keys.
+
+    strict_config: 1
+    strict_config_allow:
+      - my_custom_key
+      - feature_flags
+
 =head3 no_server_tokens (boolean)
 
 If set to true, Dancer2 will B<not> add an "X-Powered-By" header and also append
--- a/lib/Dancer2/ConfigReader.pm
+++ b/lib/Dancer2/ConfigReader.pm
@@ -54,6 +54,7 @@ my %KNOWN_CORE_KEYS = map +( $_ => 1 ), qw(
     type_library
     views
     strict_config
+    strict_config_allow
 );
 
 my %KNOWN_ENGINE_CONFIG = (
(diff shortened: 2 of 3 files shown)


lib/Dancer2/Manual/Config.pod
```diff
@@ -394,6 +394,14 @@ Conforms to the environment variable C<DANCER_STARTUP_INFO>.
 If set to true, Dancer2 will display full stack traces when a warning or a
 die occurs. (Internally sets Carp::Verbose). Default to false.
 
+=head3 strict_config (boolean)
+
+If set to true (the default), Dancer2 will warn about unknown configuration
+keys at the top level and about unknown keys in built-in engine sections where
+the options are known. Set this to false if you intentionally store
+application-specific settings at the top level or use custom engines with
```
cromedome: `if you intentionally store application-specific settings at the top level` - what is *your* suggested best practice here? I think we should explicitly specify what we think is best practice, or perhaps specify a way to ignore certain top-level keys. I know there's always the option to not enable `strict_config`, but there are also apps I've worked on where it would be a nice feature to enable, but not warn for things I know are valid.

Maybe instead we provide some convenience to specify what we consider top-level keys? I'm not trying to bloat this PR; I like the feature, just remarking on things I have come across. I can be talked down here :)

xsawyerx: I would say that we can make one more addition: top-level keys to accept (or rather, ignore). That allows people to put as many top-level keys as they want (including only one and then put everything under it) without getting warnings.

    strict_config: true
    accept_keys: [ myname, mydbconn ]
    myname: Sawyer
    mydbconn:
      filename: foo.db

cromedome: Awesome. I'm on board with this.

xsawyerx: Done. Pushed.

cromedome: I'm good with merging this.

xsawyerx:
Now just needs @yanick to take a peek. :)

yanick:
Looks good to me. If we wanted to be *really* cute, there are two things we could on top of that:

Have the engines be responsible for the schema of their parameters. Like, if an engine has a `$CLASS->param_schema`, we use that for the check, if not, well, we assume it's free-form.

If a key is not matches, use String::Approx or similar to have a 'you used 'foo', did you mean 'fool' type of warning'. 

But yeah, those are two heaped spoons of sugar. The PR as is is already Very Good Indeed :TM:  It gets my vote! :+1:

xsawyerx:
So I'm actually thinking of something different now, @yanick @cromedome: A general `strict` mode!

I'm looking at fixing the UTF-8, charset, and Unicode issues. The trick would be to translate from bytes to characters for internals and translate to bytes for output. The problem is that this might break stuff for existing code.

We can go with a lenient approach (if decoding UTF-8 fails, we just don't decode and then warn), or a strict approach in which if it fails, we throw an error. What I'm thinking is that we're lenient by default (to not break code, if we can) but we can introduce a `strict` config which will make this strict *and* will make configuration keys strict.

Alternatively, we take the lenient approach to UTF-8 and throw warnings, until a version in which we just make it strict and that's it.

yanick:
> I like it, but I think it adds complexity, maintainability, and we don't know how many will use the strict config, so maybe we should start without?

Oh yeah. Absolutely start without. As I said, this is pure, unadulterated added sugar. :-)

PR #1758, opened 2026-01-18
Move MIME ownership to App

xsawyerx:
Seems like Dancer2::Core::Runner is primarily handles the old-style dispatching (`dance`, `psgi_app`), but it's still hold onto the Dancer2::Core::MIME that everyone is using. So, time to cut it out.

With this commit, I'm moving the MIME ownership (Dancer2::Core::MIME) to the app itslf (Dancer2::Core::App). Now the users (basically Dancer2::Core::Response and Dancer2::Core::Error) have a MIME instance upon creation from App.

However, to handle cases in which there is no current app (this happens in tests), we instantiate a new one by default. We also have two cases in which we create a new one - `new_from_plack()` and `new_from_arrayref` - which are like advanced cases where all bets are off.

I deleted the `mime_type` temporarily from Dancer2::Core::Runner to see which tests and which code fails and nothing else (other than the old dispatching mechanism) is using Dancer2::Core::Runner anymore.

The goal is to remove hidden global dependencies on the Runner (which is tied to old‑style dispatching) and make MIME handling per‑app and test‑friendly.

Benefits:

* Per‑app MIME state: `default_mime_type` and custom types are now app‑scoped, not global. This is pretty cool. Each app is allowed to have its own mime handling.
* Error/Response no longer depend on Runner’s global state, which simplifies testing and supports future refactors (e.g., Throwable‑based errors).
* Fewer implicit dependencies on the old Runner dispatching path.

Key changes:

* Add `mime_type` attribute to Dancer2::Core::App and wire it into:
  * `default_mime_type` config trigger (app‑local)
  * Response builder (`_build_response`)
  * `send_file` MIME lookup
* Add `mime_type` attribute + default in Dancer2::Core::Response and Dancer2::Core::Error.
* Update `Response::content_type` to use `$self->mime_type` instead of `Dancer2->runner->mime_type`.
* Update MIME API: introduce `reset_to_default` and deprecate `reset_default` because this name doesn't confuse me. (We can fully remove it after a release or two - it's Core anyway...)
* Update tests to avoid relying on Runner MIME.
* Update a test that uses `$@` immediately which gave me grief.

### DIFF
--- a/lib/Dancer2/Core/MIME.pm
+++ b/lib/Dancer2/Core/MIME.pm
@@ -46,7 +46,7 @@ sub reset_to_default {
 
 sub reset_default {
     my ($self) = @_;
-    Carp::croak 'DEPRECATED: reset_default internal method, replaced by reset_to_default';
+    Carp::carp 'DEPRECATED: reset_default internal method, replaced by reset_to_default';
     goto &reset_to_default;
 }
 


lib/Dancer2/Core/MIME.pm
```diff
@@ -33,12 +36,18 @@ has custom_types => (
...
+sub reset_to_default {
+    my ($self) = @_;
+    $self->default( DEFAULT_MIME_TYPE() );
+}
+
 sub reset_default {
     my ($self) = @_;
-    $self->default("application/data");
+    Carp::croak 'DEPRECATED: reset_default internal method, replaced by reset_to_default';
+    goto &reset_to_default;
```
cromedome: Isn't this line unreachable?

t/error.t
```diff
@@ -180,8 +180,10 @@ subtest 'App dies with serialized error' => sub {
 subtest 'Error with exception object' => sub {
     local $@;
     eval { MyTestException->throw('a test exception object') };
+    my $exception = $@;
     my $err = Dancer2::Core::Error->new(
-        exception       => $@,
+        app             => Dancer2::Core::App->new(),
```
cromedome: 👏 👏 👏

cromedome (commented):
This looks good, question withstanding.

Issue #1594, opened 2021-02-26
Use Unicode::UTF8 if available


cromedome:
If [Unicode::UTF8](https://metacpan.org/pod/Unicode::UTF8) is available to us, use it. See @xsawyerx's suggestion in #1143.

cromedome:
Resolved by #1759.

PR #1759, opened 2026-01-21
Finally, correct UTF-8 and charset handling:

### LINKED ISSUE #686: to_json is double encoding UTF8?
try:

```
package foo;
use Dancer2;
use utf8;

get '/' => sub {
    return to_json({foo => "Simões"});
};
1;
```

vs

```
package foo;
use Dancer2;
use utf8;

get '/' => sub {
    return to_json({foo => "Simões"}, { utf8 => 0} );
};
1;
```

The second works :-|

When fixing this, take care not to break the automatic serializer.
Also, I think this is a regression.

xsawyerx:
There are a bunch of tickets on this, and this finally resolves them. Closes #686, #1124, #1143, #1594.

This helps us have decode boundaries, enforce JSON characters, and honor charset config.

Basically, Dancer2 mixed bytes and characters across requests, serializers, and responses. UTF‑8 in paths/params was left as bytes, JSON could double‑encode, and charset config was ignored in some response paths.

So my approach is:

* Enforce a clear rule: bytes at the edges, characters inside.
* Decode request inputs (`path`/`params`/`uploads`) to characters when valid UTF‑8. (Plack itself doesn't do this.)
* Keep JSON serializers emitting UTF‑8 bytes, but guarantee their input is character data. Add strict mode to fail on invalid UTF‑8.
* Honor configured charset consistently for text responses and file/template IO.

The goal is to:

* Eliminates double‑encoding and mojibake by ensuring JSON sees characters.
* Makes UTF‑8 routing and params consistent with body decoding.
* Honors charset config uniformly across responses and templates.
* Provides a strict mode for apps that want hard UTF‑8 enforcement.

Changes:

* Request boundary decoding with lenient/strict UTF‑8:
    * Decode `PATH_INFO` and params when valid UTF‑8, warn and keep bytes by default. `strict_utf8` turns those warnings into errors.
    * Added `strict_utf8` config (and env `DANCER_STRICT_UTF8`), passed into request and serializer configs.
* JSON serialization:
    * Walk data structures to ensure scalars are characters. Avoid decoding ASCII bytes to preserve numeric scalars.
    * Always emit UTF‑8 bytes. `strict_utf8` errors are re‑thrown instead of logged.
* Response encoding/charset:
    * `Response->charset` is added. Empty means no encoding.
    * Prefer charset from `Content‑Type` header. Otherwise use response charset.
    * Response `serializer` is `rw` so `set serializer => ...` updates the current response and avoids stale serializer refs (fixes `log_levels` test).
* File/template IO:
    * File utilities accept an optional charset. `Template::Tiny` honors app charset.
    * Error pages read with configured charset.

Also updated documentation on:

* Corrected JSON keyword docs: both `to_json` and `encode_json` return UTF‑8 bytes. Clarified expected UTF‑8 input for `decode`/`from_json`.
* Added `strict_utf8` config and clarified charset behavior.

Tests:

* Updated expectations around default charset (no charset unless configured).
* Added `utf8_strict.t` for lenient/strict UTF‑8 behavior.
* Adjusted tests that assumed byte output when charset is unset.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -744,7 +744,7 @@ sub _build_default_config {
     my $public = $ENV{DANCER_PUBLIC} || path( $self->location, 'public' );
     return {
         content_type   => ( $ENV{DANCER_CONTENT_TYPE} || 'text/html' ),
-        charset        => ( $ENV{DANCER_CHARSET}      || '' ),
+        charset        => ( $ENV{DANCER_CHARSET}      || 'UTF-8' ),
         strict_utf8    => ( $ENV{DANCER_STRICT_UTF8}  || 0 ),
         logger         => ( $ENV{DANCER_LOGGER}       || 'console' ),
         views          => ( $ENV{DANCER_VIEWS}
--- a/lib/Dancer2/Core/Error.pm
+++ b/lib/Dancer2/Core/Error.pm
@@ -362,6 +362,7 @@ sub throw {
 
     $self->response->status( $self->status );
     $self->response->content_type( $self->content_type );
+    $self->response->charset( $self->charset ) if defined $self->charset;
     $self->response->content($message);
 
     $self->has_app &&
--- a/lib/Dancer2/Manual/Config.pod
+++ b/lib/Dancer2/Manual/Config.pod
@@ -242,8 +242,8 @@ L<Dancer2::Plugin::Database/"AUTOMATIC UTF-8 SUPPORT">
 
 =back
 
-Default value is empty which means don't do anything. HTTP responses without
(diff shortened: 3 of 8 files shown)


lib/Dancer2/Core/Role/Serializer.pm
```diff
@@ -49,6 +49,9 @@ around serialize => sub {
         1;
     } or do {
         my $error = $@ || 'Zombie Error';
+        if ( blessed($self) && $self->config->{strict_utf8} ) {
```
cromedome: I'm not sure I am following this one. Please explain the failure condition?

veryrusty (approved):
UTF-8 by default, yet changable is you want something else.  👍 

Nice work @xsawyerx  !!

PR #1761, opened 2026-01-25
Add Logging engine Log::Any and add support for context logging

mikkoi:
Log::Any has a new feature: [Logging structed data](https://metacpan.org/pod/Log::Any::Proxy#Logging-Structured-Data). Essentially it means that you can add a hash after or in place of the logging message.
This can be really convenient if you simultaneously write the information into logs as JSON and pump the JSON into a logviewer where you can search or categorise by any JSON element.

This PR adds a logger engine for Log::Any and changes Dancer2::Core::Role::Logger to support the action.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -77,6 +77,7 @@ recommends 'Type::Tiny::XS';
 recommends 'URL::Encode::XS';
 recommends 'YAML::XS';
 recommends 'Unicode::UTF8';
+recommends 'Log::Any';
 
 suggests 'Fcntl';
 suggests 'MIME::Types';
@@ -92,6 +93,7 @@ test_requires 'Test::Fatal';
 test_requires 'Test::More';
 test_requires 'Test::More', '0.92';
 test_requires 'Test::Exception';
+test_requires 'Log::Any';
 
 author_requires 'Test::NoTabs';
 author_requires 'Test::Pod';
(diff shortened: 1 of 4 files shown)


mikkoi:
> > > You can accomplish this already with Log::Any.
> > 
> > 
> > This doesn't work for me. It does not support structured logging.
> 
> It does; if you build a new skeleton app, make sure to disable the console logger in `environments/development.yml`. I just generated a new app from scratch and tried it. Can you try again and see if this meets your needs?

It does close but not quite. I get this:

    {"category":"Dancer2::Logger::LogAny","level":"debug","message":"{'baz' => 'bah','foo' => 'bar'}","timestamp":"2026-01-31T22:37:48.531242Z"}

When I want:

    {"category":"Dancer2::Logger::LogAny","level":"debug","message":"","baz":"bah","foo":"bar","timestamp":"2026-01-31T22:37:48.531242Z"}

The problem is the *{'baz' => 'bah','foo' => 'bar'}* which would need to be processed separately.

It also doesn't handle a message **and** structure:

    `debug "Something happens here", { foo => 'bar', baz => 'bah' };`

> 
> That being said, I find `logging_context()` intriguing.

We can add all sorts of additional info there which a logviewer can index.

cromedome:
> This is quite radical proposal, but how about if we changed Dancer2's own logging system to Log::Any? With the caveat that we can maintain backwards compatibility, of course.

I don't have an immediate answer, and need to dwell on that for a bit. There's a lot to think over there, and I am not sure the gains are worth the effort. I think Log::Any is overkill for a lot of our users. Other core team members might have another view they'd like to share.

cromedome:
@mikkoi in the interim, I'd love a new/updated PR with the non-Log::Any part of this.

Issue #1630, opened 2021-07-30
Conflict between apps when using DSL object in plugins


nfg:
Hello!

At $WORK we have a fairly complicated Dancer2 app that's grown over the years. It's composed of three different Dancer2 apps, one handling web requests, another handling API requests, and another for a weird edge case that isn't worth getting into. On development, our psgi file looks something like

```perl
builder {
    mount '/' => App::Web->to_app;
    mount '/api' => App::API->to_app;
    mount '/other' => App::Other->to_app;
};
```

Cool. We have some Dancer2::Plugin cases to share code between them. It's been working pretty well so far.

For a new ticket, we're adding a new endpoint in App::API that uses a plugin, and this plugin -- which worked fine before -- spits out errors like this:

```
Can't call method \"var\" on an undefined value at lib/App/Plugin/StaticPage.pm line 273.\n
```

And the line is

```
$plugin->dsl->app->request->var('layout' => $page->layout);
```

OK. I dug into it. Eventually I figured out that for that bit of code, `$plugin->app` and `$plugin->dsl->app` referred to the different apps. The web app adds some template hooks, and from there everything blows up. It looks like whichever app loads the plugin first stakes its claim on the `dsl()` function installed by Dancer2::Plugin at https://github.com/PerlDancer/Dancer2/blob/92ed2f64a9b814fc4f0f2c06c9123d4cf49cbaf4/lib/Dancer2/Plugin.pm#L477

```perl
            *{"${caller}::dsl"} = sub {$dsl};
```

I'll see if I can put together a merge request reproducing this behaviour.

xsawyerx:
I think this is the same issue as discussed in #1449.

PR #1764, opened 2026-01-29
Make Plugin dsl keyword App-specific

xsawyerx:
Plugins were caching a single DSL instance at import time. In a multi‑app process, that meant `plugin->dsl` and bare dsl calls resolved to the DSL of the first app that loaded the plugin, breaking per‑app behavior.

This change makes the plugin `dsl` accessor resolve dynamically on each call by walking the call stack to find the consuming app's DSL, and explicitly skips plugin packages to avoid recursion.  This preserves compatibility with bare `dsl` usage while restoring correct per‑app resolution.

This approach keeps the existing API (including bareword `dsl`) as is.

### DIFF
--- a/lib/Dancer2/Plugin.pm
+++ b/lib/Dancer2/Plugin.pm
@@ -461,17 +461,14 @@ END
 
     die $@ if $@;
 
-    my $app_dsl_cb = _find_consumer();
-
-    if ( $app_dsl_cb ) {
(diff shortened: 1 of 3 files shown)


xsawyerx:
Also fixes #1630.

xsawyerx:
Considering @yanick is the expert on the gnarly plugin layer, I'll take it as gospel. :)

PR #1765, opened 2026-01-29
`send_as` should use the full serializer, including hooks

### LINKED ISSUE #1709: send_as JSON doesn't seem to be calling serializer hooks
I'm working on a new plugin to intercept send_as serialized requests, and (despite the fact that the call to execute_hook is in the Role's "around"), I'm not seeing the before_serializer or after_serializer hooks firing.  The code in send_as *seems* to be calling up Dancer2::Core::Serializer::JSON::serialize here, but hooks just aren't firing!  I've tried adding a before_serializer hook in the application, as well as this, in the plugin I'm working with:

`    $plugin->app->add_hook( Dancer2::Core::Hook->new(
        name => 'engine.serializer.before',
        code => sub { warn 'Oh, yeah!' },
    ));
`

I think I'll put in some tracers in D2 code to make sure what *apparently* should be happening here, actually *is*.  (It's possible I'm Doing It Wrong, too, of course.)

xsawyerx:
Closes #1709.

### DIFF
--- a/lib/Dancer2/Core/App.pm
+++ b/lib/Dancer2/Core/App.pm
@@ -1030,7 +1030,14 @@ sub send_as {
     # load any serializer engine config
     my $engine_options =
         $self->_get_config_for_engine( serializer => $type, $self->config ) || {};
-    my $serializer = $serializer_class->new( config => $engine_options );
+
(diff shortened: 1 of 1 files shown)


GeekRuthie:
👍

PR #1768, opened 2026-02-13
Default strict_config to false

cromedome:
So as not to blow up the logs for existing applications, restore the default to false. To encourage use of `strict_config` however, change the default for newly-generated apps to true.


xsawyerx:
Maybe we should set it to true in the scaffolded app?

PR #1782, opened 2026-03-16
Restore pre-Path::Tiny HasLocation upward traversal semantics

xsawyerx:
A regression in app root detection was introduced by the `Path::Tiny` migration in `Dancer2::Core::Role::HasLocation`.

When walking up parent directories to find an app root (`lib`+`bin` or `.dancer`), switching to `$subdir->parent` changed behavior for relative paths (notably `.`), which could stop effective upward traversal and lead to incorrect root selection on some systems.

Fix by restoring pre-migration traversal behavior:

* advance with `Path::Tiny::path($subdir, '..')` each iteration
* stop at filesystem root via absolute-path comparison

This keeps existing caller semantics intact while fixing template/public/config resolution failures caused by misdetected app location.

### DIFF
--- a/lib/Dancer2/Core/Role/HasLocation.pm
+++ b/lib/Dancer2/Core/Role/HasLocation.pm
@@ -76,9 +76,11 @@ sub _build_location {
             last;
         }
 
-        $subdir = $subdir->parent;
+        # Preserve pre-Path::Tiny traversal behavior: keep walking "up"
+        # even when current subdir is ".".
+        $subdir = Path::Tiny::path( $subdir, '..' );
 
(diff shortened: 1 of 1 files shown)


Grinnz:
On my machine this fails a lot of tests still, seems to be the same failures as reported in the CI

xsawyerx:
Hmm... Okay. I need to figure out how to reproduce it.

xsawyerx:
@toddr-bot your output is incredibly verbose that even I can't compete with it. It makes it incredibly difficult to read through. Please stop or find a way to be succinct and clear.

cromedome:
Closing in favor of #1802 .

PR #1790, opened 2026-03-24
1784 json serializer decode hashref

### LINKED ISSUE #1784: UTF-8 changes to Serializer::JSON break on readonly values
We have a data structure in YAML (swagger api definition) that gets Loaded (YAML::XS) in a Dancer2 app, where that structure can also be emmitted as JSON (for swagger-ui and similar tooling).

This "just worked" before the v2.1.0 release, afterwards we were getting `undef` returned instead of the expected serialized json. This was  tracked down to:
```
Modification of a read-only value attempted at Dancer2/Serializer/JSON.pm line 97.
```
with the new character detection tripping over the read-only boolean `true` value in the data structure. Simple repro (perl 5.32, Dancer2 v2.1.0, YAML::XS 0.904):
```
#!perl

use warnings;
use strict;
use Dancer2::Serializer::JSON;

use YAML::XS;
#$YAML::XS::Boolean = 'JSON::PP';

my $yaml =<<'YAML';
parameters:
  - name: body
    in: body
    required: true
YAML

# On perl < 5.36, the boolean (true) will be loaded as
# special perl variables PL_sv_yes and PL_sv_no, which are readonly
my $defn = YAML::XS::Load($yaml);

my $serializer = Dancer2::Serializer::JSON->new();
my $serialized = $serializer->serialize($defn);

print 'Serialization ' . ( ( defined $serialized ) ? 'succeeded' : 'failed' ) . "\n";
```

### LINKED ISSUE #1789: Dancer 2.1.0 JSON serializer changes cause infinite recursion with DBIx result objects with relationships
After updating to Dancer 2.1.0, my team's application started experiencing some breakages related to infinite recursion:

```log
Deep recursion on subroutine "Dancer2::Serializer::JSON::_ensure_characters" at /usr/local/lib/perl5/site_perl/5.38.5/Dancer2/Serializer/JSON.pm line 90.                                                       
Deep recursion on subroutine "Dancer2::Serializer::JSON::_ensure_characters" at /usr/local/lib/perl5/site_perl/5.38.5/Dancer2/Serializer/JSON.pm line 90.                                                       
Deep recursion on subroutine "Dancer2::Serializer::JSON::_ensure_characters" at /usr/local/lib/perl5/site_perl/5.38.5/Dancer2/Serializer/JSON.pm line 90.                                                       
Deep recursion on subroutine "Dancer2::Serializer::JSON::_ensure_characters" at /usr/local/lib/perl5/site_perl/5.38.5/Dancer2/Serializer/JSON.pm line 90.                                                       
Deep recursion on subroutine "Dancer2::Serializer::JSON::_ensure_characters" at /usr/local/lib/perl5/site_perl/5.38.5/Dancer2/Serializer/JSON.pm line 90.                                                       
Deep recursion on subroutine "Dancer2::Serializer::JSON::_ensure_characters" at /usr/local/lib/perl5/site_perl/5.38.5/Dancer2/Serializer/JSON.pm line 90.                                                       
```

After troubleshooting, it appears that the new behavior does not take into account `DBIx::Class` result objects that specify relationships, where the result of following the relationships would be an infinite recursion scenario.

We were able to work around this initially by reverting the Dancer2 version and then by adjusting our controller code to map the inflated column data to an array before attempting to return it for the serializer to handle, but it may be the case that others are bitten by this change in a similar way and expect the payload to be handled by `JSON::MaybeXS::encode_json` as it was before.

Caused by #1759.

veryrusty:
Fixes for JSON serializer `_ensure_characters`:
* don't recurse into blessed objects. Resolves #1789
* create a new hash in case keys were marked read-only. Resolves #1784

The test suite will likely fail due to hash randomization with this applied, so require sorted keys (canonical representation) in encoded json (string) comparisons.

### DIFF
--- a/lib/Dancer2/Serializer/JSON.pm
+++ b/lib/Dancer2/Serializer/JSON.pm
@@ -84,6 +84,7 @@ sub _ensure_characters {
     }
 
     if ( is_plain_hashref($entity) ) {
+        my %ret;
         for my $key ( keys %{$entity} ) {
             my $value = $entity->{$key};
             my $decoded_key = _ensure_scalar( $key, $strict_utf8, $self );
@@ -92,12 +93,12 @@ sub _ensure_characters {
 
             if ( $decoded_key ne $key ) {
                 delete $entity->{$key};
-                $entity->{$decoded_key} = $decoded_value;
+                $ret{$decoded_key} = $decoded_value;
             } else {
-                $entity->{$key} = $decoded_value;
+                $ret{$key} = $decoded_value;
             }
         }
-        return $entity;
+        return \%ret;
     }
 
     return $entity;
--- a/lib/Dancer2/Serializer/JSON.pm
+++ b/lib/Dancer2/Serializer/JSON.pm
@@ -2,7 +2,7 @@ package Dancer2::Serializer::JSON;
 # ABSTRACT: Serializer for handling JSON data
 
 use Moo;
-use Ref::Util qw< is_arrayref is_hashref >;
+use Ref::Util qw< is_plain_arrayref is_plain_hashref >;
 use JSON::MaybeXS ();
 use Encode qw(decode FB_CROAK);
 use Scalar::Util 'blessed';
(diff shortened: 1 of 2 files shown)


cromedome (approved):
Solid fixes. 👍 from me. @xsawyerx???

glance-:
Fixes the same issue for me.

xsawyerx (approved):

xsawyerx:
@veryrusty++ !

cromedome:
Merged, thank you!

PR #1791, opened 2026-03-30
Correct error codes for file checks in send_file

glance-:
Back in 8caaf43f9cbdb , the return codes for send_file on a missing file was altered.

If a file doesn't exist, the obvious return code should be 404, and if you're trying to do send_file on something that's not a file , 403 is a sane return code.


xsawyerx (approved):

cromedome (approved):
Thanks @glance- ! 👍

PR #1801, opened 2026-08-09
Fix failing CI

cromedome:
Our CI setup has been failing for a while. Mostly this was due to build failures in `HTTP:: XSCookies`, but further investigation indicated that the problems ran deeper. Some of the Perl images we had been using were deprecated or simply non-existent, and some of our tools like `dzil` were not available on older versions of Perl. 

This provides the most up-to-date CI file that I know how to provide. Until there's a newer release of `HTTP::XSCookies`, we simply skip it when setting up to run CI. Once a new version has been published, we can remove the extra code in the YAML file that skips its installation. Both pull requests that I submitted have been merged, so hopefully this will not be a long wait. See these for reference: https://github.com/gonzus/http-xscookies/pull/11 and https://github.com/gonzus/http-xscookies/pull/12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,12 @@ on:
     branches: '**'
   schedule:
     - cron: '42 5 * * *'
+  workflow_dispatch:
+
+# A new push to a branch supersedes any run still going on that branch.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 # Build and test Dancer on several versions of Perl using an image that already has a lot of modules installed.
 # This will provide a fast feedback if a commit broke anything in the unit-tests.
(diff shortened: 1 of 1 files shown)


GeekRuthie (approved):
Looks good to me.  I'm no CI guru, but "up-to-date", "working", and "quirks commented" are all looking good to me.

.github/workflows/ci.yml
```diff
@@ -31,22 +37,36 @@ jobs:
...
           - '5.36-bookworm'
-          - '5.20-buster'
 
     container:
       image: perldocker/perl-tester:${{ matrix.perl-version }}     # https://hub.docker.com/r/perldocker/perl-tester
 
     name: Build on Linux with Perl ${{ matrix.perl-version }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v5
```
veryrusty: My preference is to bump action versions to their latest (unless there are breaking changes). Then you can forget about them for longer ;)   So:
* actions/checkout@v7
* actions/upload-artifact@v7
* actions/upload-artifact@v8

cromedome: Done! Worked much better once I realized that second `upload-artifact` was `download-artifact`! 

Please re-review when able, @veryrusty. Thanks!

veryrusty (approved):
Thanks for bumping the action versions @cromedome  👍 merge away!

PR #1802, opened 2026-08-09
Fix directory detection heuristic

### LINKED ISSUE #1781: Test failures with 2.1.0
Release 2.1.0 is failing tests on my system (and there are some other failing reports on cpantesters, but not consistently), the errors seem related mainly to finding template files.

https://www.cpantesters.org/cpan/report/c12e9f60-2052-11f1-aa81-9acf868642b9

cromedome:
Fixes #1781. Supersedes #1782.

  ## Problem

  `_build_location` walks up from the caller's directory looking for the application
  root. Its test for "is this an app root?" was *"does it hold both `lib/` and `bin/`?"*
  — which also describes every Perl distribution root, most home directories, `/usr`,
  and `/usr/local`.

  This was harmless while the upward walk was effectively stalling, since it rarely
  left the script's own directory. 2.1.0 made the walk work, and the weak test
  immediately started selecting directories far above the application. Apps then
  resolved `config.yml`, `views/`, `public/` and `logs/` against the wrong root, which
  is the template and public-file failures in #1781.

  ## Reproducing

  The trigger is any directory holding **both `lib/` and `bin/`** above the app. A
  Dancer2 checkout already has `lib/`:

  ```bash
  mkdir bin
  prove -l t/auto_page.t     # fails
  rmdir bin
  prove -l t/auto_page.t     # passes
  ```

  That's why a plain `cpanm Dancer2` fails for some people and not others: `~/bin` +
  `~/lib` is enough, and the walk climbs out of the extracted tarball into the home
  directory. Reconstructing that:

  ```bash
  mkdir -p /tmp/fakehome/bin /tmp/fakehome/lib
  git archive HEAD --prefix=Dancer2-2.1.0/ | tar -x -C /tmp/fakehome
  cd /tmp/fakehome/Dancer2-2.1.0 && prove -Ilib -r t/
  ```

  11 failing test files on `main`, 0 on this branch.

  ## Fix

  Recognise an app root by what it actually holds, closest-first as the walk climbs:

  1. a `.dancer` file — the explicit marker
  2. `environments/`, `views/` or `public/`
  3. `config.<ext>` for any extension `Config::Any` handles
  4. `lib/` + `bin/` — kept for apps with no Dancer2 artifacts, but checked last so it
     only wins when found closer than real evidence. Still skips `blib/`.

  Because `t/` holds `config.yml`, `views/` and `public/`, it now matches immediately
  and the walk never approaches the distribution root or `~`.

  A second commit stops the walk at a checkout or distribution boundary (`.git`,
  `.hg`, `.svn`, `Makefile.PL`, `Build.PL`, `dist.ini`, `cpanfile`) rather than
  escaping above it. That only applies to directories already rejected as app roots,
  so a `dancer2 gen` app — which ships a `cpanfile` and a `Makefile.PL` alongside
  `.dancer` and `config.yml` — is accepted several tiers earlier.

  The walk is also canonicalised with `realpath` up front, and now stops when
  `parent()` reaches its own fixed point instead of comparing against `rootdir`. A
  relative `$subdir` became a chain of `..` segments, which made the `blib/` regex
  meaningless and the root comparison non-portable. That's the real issue behind
  #1782, whose stated premise — `path('.')->parent` being a fixed point — doesn't hold
  on current Path::Tiny.

  ## Note on the test change

  `t/issues/memleak/die_in_hooks.t` asserts on stderr and was relying on detection
  *failing*: it previously found no app root, fell back to its own directory, and so
  never loaded `t/issues/config.yml` with its `logger: "Note"`. It now has a config of
  its own stating that requirement.

### DIFF
--- a/lib/Dancer2/Core/Role/HasLocation.pm
+++ b/lib/Dancer2/Core/Role/HasLocation.pm
@@ -51,6 +51,22 @@ my @_config_files = map "config.$_",
 # directory worth calling the app root.
 my @_app_dirs = qw< environments views public >;
 
+# Files and directories marking the root of a checkout or a distribution.
+my @_project_markers =
+    qw< .git .hg .svn Makefile.PL Build.PL dist.ini cpanfile >;
+
+# Is this directory a ceiling the walk must not cross? Whatever the app root
+# is, it sits at or below its own project root, never in the home directory
+# or the build directory above it. Only consulted for directories _is_app_root
+# has already rejected, so a project root that is also an app root - which is
+# the usual case, 'dancer2 gen' writes a cpanfile and a Makefile.PL - has been
+# accepted long before this is reached.
+sub _is_project_boundary {
+    my $dir = shift;
+    $dir->child($_)->exists and return 1 for @_project_markers;
+    return 0;
+}
+
 # Does this directory look like the root of a Dancer2 application?
 #
 # Historically the only test was "holds both lib/ and bin/". That is a poor
@@ -102,6 +118,10 @@ sub _build_location {
             $subdir_found = 1;
             last;
         }
+
+        # a checkout or distribution root that is not itself an app root:
+        # stop rather than escape into whatever happens to lie above it
+        last if _is_project_boundary($subdir);
 
         # stop at the volume root, where parent() is its own fixed point
         my $parent = $subdir->parent;
--- a/lib/Dancer2/Core/Role/HasLocation.pm
+++ b/lib/Dancer2/Core/Role/HasLocation.pm
@@ -41,6 +41,44 @@ sub _build_location_path {
     return Path::Tiny::path( $self->location );
 }
 
+# Config file basenames a Dancer2 app root may hold. Keep this in sync with
+# Config::Any->extensions, which is what Dancer2::ConfigReader actually reads.
+my @_config_files = map "config.$_",
+    qw< cnf conf ini json jsn pl perl xml yml yaml >;
+
+# Directories a Dancer2 app root may hold. These are precisely the things
+# 'location' is used to resolve, so a directory holding one of them is a
+# directory worth calling the app root.
+my @_app_dirs = qw< environments views public >;
+
+# Does this directory look like the root of a Dancer2 application?
+#
+# Historically the only test was "holds both lib/ and bin/". That is a poor
+# proxy: it matches every Perl distribution root, most home directories, /usr
+# and /usr/local. It went unnoticed for years because the upward walk in
+# _build_location was effectively stalling, so it rarely got the chance to
+# leave the script's own directory. Once the walk started working the weak
+# test began selecting directories far above the real app. See GH #1781.
+sub _is_app_root {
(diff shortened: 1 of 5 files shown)


lib/Dancer2/Core/Role/HasLocation.pm
```diff
@@ -41,6 +41,60 @@ sub _build_location_path {
...
+    qw< cnf conf ini json jsn pl perl xml yml yaml >;
+
+# Directories a Dancer2 app root may hold. These are precisely the things
+# 'location' is used to resolve, so a directory holding one of them is a
+# directory worth calling the app root.
+my @_app_dirs = qw< environments views public >;
+
+# Files and directories marking the root of a checkout or a distribution.
+my @_project_markers =
+    qw< .git .hg .svn Makefile.PL Build.PL dist.ini cpanfile >;
```
xsawyerx, line 56: My worry about this is that it's an inaccurate proxy. "If you're in a repo, that's a Dancer directory," but that's not necessarily true.

Imagine you have a repo of a personal program. Inside it, there's a directory `app/` and *that* is a Dancer app. If you were to walk up, you would consider the repo to be *the Dancer2 app* but it isn't.

I think we should figure out "these are very likely Dancer apps" and the rest are just a big no.

lib/Dancer2/Core/Role/HasLocation.pm
```diff
@@ -41,6 +41,60 @@ sub _build_location_path {
     return Path::Tiny::path( $self->location );
 }
 
+# Config file basenames a Dancer2 app root may hold. Keep this in sync with
+# Config::Any->extensions, which is what Dancer2::ConfigReader actually reads.
+my @_config_files = map "config.$_",
+    qw< cnf conf ini json jsn pl perl xml yml yaml >;
```
xsawyerx, line 47: This seems right but fragile. Not *that* fragile, because we don't change this often, but a hard-coding. I'm okay with keeping it though.

xsawyerx (approved):
Good, except the one concern I had there.

lib/Dancer2/Core/Role/HasLocation.pm
cromedome: I wasn't the most thrilled with doing in this way, but we need *something* that works. Which is better than what we have now, which doesn't.

lib/Dancer2/Core/Role/HasLocation.pm
cromedome: I suspected this might be the most controversial part of the PR. I was trying to have some sort of backstop so we don't end up going all the way up to the root of the filesystem, but like you said, thar be dragons.

Are `@_app_dirs` enough? Do I need to be more robust and look to see if there are any env vars set that change the location of these? My thought is maybe and yes.

bigpreshkt (approved):
This still feels potentially a bit fragile, but it's clearly better than what we had.

cromedome:
Merged what we have as it fixes #1781. Will open a separate ticket for improvement of this.

PR #1804, opened 2026-09-10
Remove insecure fallback session-ID generator; use Crypt::URandom always

bigpresh:
## Summary

The default session engine (`Dancer2::Core::Role::SessionFactory`) previously chose between two session-ID generation paths at runtime:

- If both `Math::Random::ISAAC::XS` and `Crypt::URandom` were installed, IDs came from an ISAAC CSPRNG seeded with OS randomness.
- Otherwise it fell back to `sha1()` over a hand-rolled seed made of `rand()`, `__FILE__`, a process-local counter, `$$`, the object string and a `List::Util::shuffle` of the base64url alphabet.

That fallback was **not cryptographically strong**. `rand()` and `shuffle` both draw from the same drand48 stream, so the shuffle contributes no independent entropy; the real seed state is ~48 bits plus values an attacker can often narrow (`time`, `$$`, a compile-time constant path). Recovering the stream makes subsequently generated session IDs predictable — session prediction, not just fixation. Because both modules were **optional**, this weak path was in fact the default for any install lacking them.

## Changes

- `generate_id()` now always reads 20 bytes from the OS CSPRNG via `Crypt::URandom` (`lib/Dancer2/Core/Role/SessionFactory.pm`). Output keeps the **exact same shape** — 4-byte big-endian timestamp + 20 bytes of entropy = 24 bytes = 32 base64url characters — so existing sessions, `validate_id()` and anything matching on ID length work unchanged. The timestamp prefix keeps IDs roughly monotonic (used by cleanup scripts).
- `Crypt::URandom` is now a **hard runtime requirement** (cpanfile + generated app/tutorial cpanfiles) instead of a recommendation. `Math::Random::ISAAC::XS` is no longer needed and was dropped from all cpanfiles. The lazy-seeding/fork-safety concerns behind the ISAAC closure are moot: every call fetches fresh entropy straight from the OS.
- Hardened `validate_id()` in the same file:
  - Uses `\A ... \z` instead of `^ ... $`, so IDs with a trailing newline no longer validate (previously `"abc\n"` passed).
  - Rejects `undef` and IDs longer than 128 characters, so a malicious multi-megabyte cookie value is rejected cheaply instead of being `stat()`'d by the file session backend.
- POD updated to match; `t/session_object.t` asserts the new `validate_id` behaviour (trailing/embedded newlines, over-long IDs, `undef`).

## Testing

Full test suite passes: 84 files, 977 tests.

Related security review: [Claude-Bug-Security-Sweep.md] section 5 "Session-ID fallback PRNG is not cryptographically strong".


cromedome (approved):
I think it took longer for me to wrap my head around the first paragraph than it did the fix 😂 

👍 from me. Bonus points for cleaning up the app skel and removing another dependency!

yanick (approved):
Looks good and straight-forward enough!

cromedome:
Merged, thanks!

Issue #1835, opened 2026-09-16
Add a characterization test suite


bigpresh:
Adds a characterization suite under `t/unit`, `t/integration` and `t/e2e`, written against
this distribution by Curtis "Ovid" Poe, covering the behaviour fixed above and the
surrounding areas: config reading, merging and strict mode; session lifecycle; error
rendering and its fallback to static pages; dispatch flow; engine hooks; template
rendering; cookies; routes; requests; and logger levels.

These pin behaviour as it stands today rather than asserting anything new, so they document
what the framework currently promises and turn a future change to it into a visible
decision. The comments say why each thing is asserted, including where current behaviour is
pinned without being endorsed.

Fixed in cf187716 and 7d5f1a9b.

Issue #1834, opened 2026-09-16
share/.gitignore hid the skeleton's environment configs from the dist


bigpresh:
`share/.gitignore` is shipped data, copied into the user's new application by `_check_git`,
and its `sessions/`, `logs/` and `environments/` entries describe a running Dancer2
application. But living at that path made git honour it against *this distribution's own*
`share/` tree as a side effect — where `environments/` covers the skeleton's per-environment
configs, so they were never committed and never shipped.

**Fix.** Renamed to `share/gitignore`, a name git does not honour as an ignore file. The
generated application still receives it named `.gitignore`, which the copy now states
explicitly rather than leaving to the destination directory.

Fixed in ab0c22b0.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

Issue #1833, opened 2026-09-16
%D was documented as a log_format character but never implemented


bigpresh:
The `log_format` documentation listed `%D`, described as "timer", but `map_chars_to_subs`
returns no `D` key — so a logger configured with it (`log_format: '[%D] %m'`) warned
`%D not supported.` through Carp on **every single message logged**, and rendered the field
as `-`. Anyone who followed the documentation got a warning per log line. The formatting
lives in the shared role, so this applied to every logger engine.

**Fix.** `%D` is removed from the list rather than implemented: there is no timing state in
the logger role to implement it from, and adding some is a feature rather than a fix for
the warning. A cross-check of the whole documented list against `map_chars_to_subs` found
no other mismatch in either direction. Also documents what an unrecognised format character
does, since that is what a user who mistypes one will hit.

Fixed in c3d65b94.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

Issue #1832, opened 2026-09-16
Serializer::Mutable documented one header precedence order for both directions


bigpresh:
The DESCRIPTION gave a single precedence order covering both directions. `serialize()`
calls `_get_content_type('accept')` and `deserialize()` calls it with `'content_type'`, so
the orders genuinely differ — and the code is right to differ: `Accept` is what a client
uses to say what it wants *back*, which need not be what it sent.

**Fix.** The documentation describes both directions instead of contradicting one of them.

Fixed in df48697e.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

PR #1836, opened 2026-09-16
Security and bug fixes from the generated-test-suite review

### LINKED ISSUE #1822: CR and LF were not stripped from response header names
`headers_to_array` applied two substitutions to every header *value* — one folding linear
whitespace, one removing CR and LF, commented "remove CR and LF since the char is invalid
here" — and applied nothing at all to the header *name* pushed onto the same array two
lines later. A CRLF injected into a name therefore reached the PSGI header array intact:

```perl
$response->header("X-Bad\r\nInjected: yes" => 'v')
```

which is the same response-splitting shape the value is already guarded against. Whether
it reaches the wire depends on the PSGI server; some validate header names, many do not.

**Fix.** The name now gets the same treatment as the value, so no element of the array
can carry CR or LF.

Fixed in ff89ac12.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1823: AutoPage layout guard was case-sensitive, serving layouts on macOS/Windows
`Dancer2::Handler::AutoPage` refuses to serve a template out of the layout directory as a
page, but decided that from the spelling of the request path
(`m{^/\Qlayout_dir\E/}`). On a case-insensitive filesystem — the default on macOS (APFS)
and Windows (NTFS) — a differently-cased request misses the guard while resolving to the
very same file. So `/layouts/main` was correctly refused while `/Layouts/main` returned
200 and rendered the layout.

**Fix.** The path-prefix test is kept as a cheap fast path, but the decision is now made
from what is actually on disk rather than from how the request was spelled.

The first attempt at this resolved both directories with `realpath` and compared them with
`Path::Tiny::subsumes`, the containment approach `send_file` uses. That was not enough,
and is worth recording because it looks sufficient: `realpath` resolves symlinks but does
not canonicalise case, and `subsumes` is a prefix match on the path string. So
`views/Layouts` and `views/layouts` still compared as two different directories while
naming one — the guard did not fire, and the bug survived the fix on exactly the platforms
where it was reachable. It passed CI only because the Linux jobs cannot reach the bug
either way; it was caught when the macOS job was added to the matrix.

The guard now compares filesystem identity — device and inode — walking the page
directory's ancestors up to `views` and stopping there, since nothing above `views` is
Dancer2's to reason about. No spelling can disguise identity, so this settles case
folding, symlinks and hardlinks together. Inode numbers are compared with `eq` rather than
`==`: where they are too large for perl to hold as integers `stat` returns them as decimal
strings, and comparing numerically would round them, which here would mean serving a page
that should have been refused. Some Windows configurations report an inode of 0 for every
file, which would make every directory compare equal to every other; where identity is
unusable the comparison falls back to matching paths case-insensitively, which assumes the
folding rules but fails closed.

The page directory comes from `$template->views` (which `Dancer2::Core::Role::Template`
guarantees is absolute) rather than `view_pathname`: that method is engine-specific, and
`Dancer2::Template::TemplateToolkit` overrides it to return a bare template name for TT2's
own `INCLUDE_PATH` to resolve, so it does not name a location on disk.

Fixed in 293fce08, corrected in 753b3853 and 7aa24256. Verified on a loopback vfat mount,
which reproduced the macOS failure before the correction and passes after.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1824: Response content assigned a second time is never encoded
`is_encoded` latched on the first assignment and `encode_content` returns early whenever
it is set, so with charset UTF-8:

```perl
$response->content('first');        # encodes, latches is_encoded
$response->content("hi\x{263A}");   # returned untouched
```

left a wide character in the PSGI body, a `Content-Length` of 3 (the character count)
rather than the 5 bytes UTF-8 needs, and a `Content-Type` still announcing
`charset=UTF-8`. Reachable from ordinary application code: an `after` hook that rewrites
`response->content`, or `halt()` called once content had already been set.

**Fix.** `is_encoded` now describes the content currently assigned rather than the
response for the rest of its life — the `around content` modifier clears it before
encoding, so each assignment is judged on its own merits. The callers that mark
already-read bytes by hand are unaffected: `Handler/File.pm` sets `is_encoded(1)` *after*
calling `content()`, and `send_file` assigns through the raw hash element, bypassing the
modifier entirely.

Fixed in ff89ac12.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1825: Serializer::Mutable fails on a Content-Type with a charset parameter
The format lookup was an exact hash-key match on the raw header value, so an ordinary
charset parameter made it miss:

```
Content-Type: text/x-yaml                   200, deserialized as YAML
Content-Type: text/x-yaml; charset=utf-8    400
```

The miss selected the JSON fallback, which was then handed a YAML body and failed the
request. The same happened on `Accept`, where `text/x-yaml; charset=utf-8` silently
returned JSON instead of YAML. JSON escaped notice only by accident: the fallback when the
lookup misses *is* JSON, so `application/json; charset=utf-8` still worked and hid the
defect for the commonest case.

**Fix.** The header is normalised before the lookup — parameters stripped at the first
`;`, trimmed, lowercased — so the mapping is keyed by bare content types, as its own
documentation describes them, and an uppercase spelling matches too. This is what the rest
of the distribution already does; `Response.pm` calls `content_type_charset` precisely to
split the charset off the type. A comma-separated multi-type `Accept` header is unchanged:
still matched whole, not split, so it falls through to the default.

Fixed in df48697e.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1826: uri_for_route dies on a route parameter whose value is 0
For a route declared `get 'item' => '/item/:id'`, `uri_for_route('item', { id => 7 })` and
`{ id => 'abc' }` both worked, but `{ id => 0 }` died with "Route item uses the parameter
'id', which was not provided" — when it had been provided. The value was tested for truth:

```perl
my $value = $route_params->{$param} or die ...
```

so any defined-but-false value was rejected, though 0 is an ordinary database ID, list
index or page number.

**Fix.** The value is tested for definedness, so 0 is accepted and round-trips. The empty
string is a separate case and is still refused, with its own message: `:param` compiles to
`([^/]+)` and so matches at least one character, meaning an empty value produces `/item/`
— a URI that 404s against the very route it was generated from. Handing back a URL that
cannot work is worse than refusing to build one.

Further questions about encoding parameter values (slashes, `%`) are tracked separately in
#1799.

Fixed in c6ce8cba and c15b8acc.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1827: to_app recompiles hook wrappers on every call
`compile_hooks` wraps each hook and writes the wrappers back with `replace_hook`, so
calling `to_app` again wrapped the already-wrapped hooks. On the failure path each layer
treated the inner layer's croak as a fresh hook failure and fired
`core.app.hook_exception` itself — one dying `before` hook reporting 1, then 2, then 3
times as `to_app` was called again.

**Fix.** `compile_hooks` records each wrapper it produces and passes its own earlier work
through untouched. That is deliberately per-hook rather than a per-app "already compiled"
flag: `finish()` adds postponed plugin hooks immediately *after* calling `compile_hooks`,
so a flag would leave those permanently unwrapped on a later `to_app` call — and an
unwrapped hook does not merely go unreported, it dies through the dispatcher without ever
reaching `core.app.hook_exception`.

Fixed in 8fd31a32.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1828: A NUL byte in a static file request warns once per request
The default static handler's file-existence condition handed the raw `PATH_INFO` to
`Path::Tiny`, which warned `Invalid \0 character in pathname` on every such request —
putting an attacker-supplied path into the log, once per request, at no cost to the
attacker. `Dancer2::Handler::File` rejects the same input outright.

**Fix.** Such paths are refused before `Path::Tiny` sees them. The 404 the client receives
is unchanged.

Fixed in 2446a09f.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1829: dancer2 gen -g dies after writing the application
`dancer2 gen -a G::App --path DIR -d gapp -g` wrote the whole application and then died
with `Can't locate object method "absolute" via package "DIR/gapp"`, exiting 255. The same
happened for `-r <uri>`, which implies `-g`.

`_check_git` took its path from `$vars->{apppath}`, which `run()` had stored as a plain
string, and called `->absolute` on it; the very next line of `run()` keeps the
`Path::Tiny` object for exactly this purpose, as `$vars->{appdir}`. Since `_check_git`
runs after the files are copied, the user was left with a generated application, no
repository — `git init`, `git add`, the initial commit and any `git remote add` all come
after the `chdir` that died — and a stack-shaped error where the "Your new application is
ready" banner should have been.

**Fix.** Uses the absolute path `run()` had already computed.

Fixed in ab0c22b0.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1830: dancer2 gen names the application directory after the raw module name
`dancer2 gen -a Other::App` with no `-d` created a directory named `Other::App`, colons and
all, rather than `Other-App`. `_get_app_path` exists to produce the dashed spelling and was
being called, but the directory option defaulted to the raw application name and overrode
it — so a single generated application disagreed with itself: the directory was
`Other::App` while the `Makefile.PL` beside it cleaned `Other-App-*`.

**Fix.** The option now has no default and falls back to the dashed name. An explicit `-d`
is still honoured verbatim.

Fixed in ab0c22b0.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

### LINKED ISSUE #1831: dancer2 gen writes a MANIFEST.SKIP pattern that can never match
The line appended to a generated `MANIFEST.SKIP` was built from the full filesystem path
the application was generated into — `^/tmp/xYz/myapp-` — so it could never match:
`ExtUtils::Manifest` matches these patterns against paths relative to the distribution
root, and every other pattern in the skeleton's `MANIFEST.SKIP` is relative.

**Fix.** Appends the dashed distribution name already computed for `Makefile.PL`'s
cleanfiles.

Fixed in ab0c22b0.

Found by the generated test suite Curtis "Ovid" Poe wrote against this distribution.

bigpresh:
Security and bug fixes arising from a review of Dancer2 against a generated test suite written by Curtis "Ovid" Poe. The work was done and reviewed in a private repository so the security issues could be fixed before disclosure; this is that branch, approved there by @cromedome and @veryrusty.

## Security

Two of these are covered by published advisories:

- **GHSA-6xw8-v24c-m783** — `Dancer2::Handler::File` served files from outside `public_dir`. The request path was joined onto `public_dir` and then tested only for being a readable file, never for still being inside `public_dir`. `Path::Tiny` does not collapse `..`, so `/../secret` escaped at unlimited depth. Reachable only by an application that enables the `File` route handler *and* sets `static_handler: 0` — the default static handler refuses such paths first.
- **GHSA-v527-r4px-7vx7** — a `before` hook that refused a request could be overridden. An `on_hook_exception` handler that set a response and halted had that response destroyed by the cleanup call that followed, so dispatch read a fresh, unhalted response and ran the very route the hook had refused, side effects and all.

Two more are lower severity and have ordinary tickets:

Fixes #1822
Fixes #1823

## Bug fixes

Fixes #1824
Fixes #1825
Fixes #1826
Fixes #1827
Fixes #1828
Fixes #1829
Fixes #1830
Fixes #1831

## Documentation

Fixes #1832
Fixes #1833

## Misc

Fixes #1834
Fixes #1835

## Notes for review

This branch has been reviewed in full already on a private fork (because it included security issues), so the useful thing to look at is what came after the approvals:

- **The AutoPage fix was wrong on its first attempt and was corrected.** `293fce08` resolved both directories with `realpath` and compared them with `Path::Tiny::subsumes`. Neither canonicalises case, so on a case-insensitive filesystem `views/Layouts` and `views/layouts` compared as two directories while naming one — the guard never fired, and the bug survived the fix on exactly the platforms where it was reachable. It passed CI only because the Linux jobs cannot reach the bug either way. `753b3853` compares device and inode instead, and `7aa24256` compares inode numbers with `eq` rather than `==` (per `perldoc -f stat`, a large inode comes back as a decimal string, and comparing numerically rounds it). Verified on a loopback vfat mount and on the macOS CI job; @cromedome separately confirmed the Windows path.
- **`Changes` is deliberately terse.** Each entry points at its ticket rather than describing the bug, since the tickets carry the detail.
- **Merge, please, rather than squash or rebase.** The commit messages carry the reasoning for each fix, and the SHAs above are cited from the advisories and the tickets.

Please do not squash.


cromedome (approved):
Reiterating that @veryrusty and I already reviewed and approved in a private repo.

PR #1839, opened 2026-09-17
Also refuse code eval when $YAML::UseCode is set

bigpresh:
Follow-up to https://github.com/PerlDancer/Dancer2/pull/1838, spotted while backporting that fix to the original Dancer module (PerlDancer/Dancer#1250).

## The gap

`#1838` made `deserialize` localise `$YAML::LoadBlessed` and `$YAML::LoadCode` to `0` before handing a request body to `YAML::Load`. But `YAML::Loader::Base` decides on code loading with a plain OR:

```perl
$self->load_code($YAML::LoadCode || $YAML::UseCode)
  if defined $YAML::LoadCode or defined $YAML::UseCode;
```

so an ambient `$YAML::UseCode = 1` defeats `LoadCode = 0` on its own — and the `!!perl/code` string eval is **not** gated on `LoadBlessed` at all, so neither `LoadBlessed = 0` nor `LoadCode = 0` closes it by itself. That is the exact "hostile surrounding process" scenario the original PR set out to defend against (e.g. a session backend enabling code round-tripping process-wide).

## The fix

`deserialize` now also localises `$YAML::UseCode` to `0`. `$YAML::UseCode` has existed since well before every version named in the cpanfile note, so the `1.30` floor is unchanged.

## Tests

- Each subtest now floods the ambient environment (loads all three toggles to `1`) so the guard — not the safe default — is what keeps the serializer safe, and the no-leak test covers all three.
- New subtest with a side-effecting `!!perl/code` payload: the control shows `YAML::Load` evaluates it when flooded; `deserialize` does not. As with the bless assertions, it tests only the security property, not the shape of what comes back.

Full suite passes (208 files, 1852 tests).


cromedome (approved):
Looks good. 👍 from me

yanick (approved):
Yup, straightforward, and not a bad idea at all. 😅

